PR #10, opened 2021-12-01
logs support

### LINKED ISSUE #9: Support for Logs

scottlepp:
When using the logs panel, values have to be converted to string.  the toString function returns a Nullable(String) type
take the following query for example:

```
SELECT log_time as time, machine_group, toString(avg(disk_free)) as avg_disk_free
FROM mgbench.logs1
where log_time > toDateTime('2017-01-07 01:13:00', 'EST')
and log_time < toDateTime('2017-01-07 01:18:00', 'EST')
group by machine_group, log_time 
order by log_time
```

image

* add a converter to handle this type
* set logs to true in plugin.json to indicate this datasource supports logs

Fixes #9


hanjm:
Awesome and very useful feature, could it disable log mode at the `explore` UI?

PR #27, opened 2022-01-11
Adhoc filters!!!

bossinc:
adhoc filters with docs
removed enterprise stuff from docs
fixed template vars


src/data/adHocManager.ts
```diff
@@ -0,0 +1,21 @@
+export default class AdHocManager {
```
scottlepp, line 1: looks like we could just export a function here.  no class needed

src/data/CHDatasource.ts
```diff
@@ -1,29 +1,36 @@
...
   annotations = {};
   settings: DataSourceInstanceSettings<CHConfig>;
+  templateSrv: TemplateSrv;
+  adHocManager: AdHocManager;
 
   constructor(instanceSettings: DataSourceInstanceSettings<CHConfig>) {
     super(instanceSettings);
     this.settings = instanceSettings;
+    this.templateSrv = getTemplateSrv();
+    this.adHocManager = new AdHocManager();
```
scottlepp, line 26: not sure why we need this.  can we just import the function and use it where its needed?

scottlepp (changes requested):
looks like the AdHocManager class isn't necessary.  can just export a function

src/data/CHDatasource.ts
```diff
@@ -1,29 +1,36 @@
...
   }
 
-  async metricFindQuery(query: CHQuery) {
-    if (!query.rawSql) {
+  async metricFindQuery(rawQuery: string) {
```
scottlepp, line 29: seems like there are scenarios where this could be a string or CHQuery.  not sure if something has changed in core grafana

src/data/adHocFilter.ts
```diff
@@ -0,0 +1,20 @@
+
+export function applyAdHocFilter(rawSql: string, adHocFilters: AdHocVariableFilter[]): string {
+  if (rawSql != '' && adHocFilters && Object.keys(adHocFilters).length > 0) {
+    let whereClause = 'WHERE';
+    for (let k of adHocFilters) {
```
scottlepp, line 5: const instead of let

src/data/adHocFilter.ts
```diff
@@ -0,0 +1,20 @@
...
+      let v = isNaN(Number(k.value)) ? `'${k.value}'` : Number(k.value)
```
scottlepp, line 6: const instead of let

src/data/adHocFilter.ts
```diff
@@ -0,0 +1,20 @@
...
+      whereClause += ` ${k.key} ${k.operator} ${v} ${k.condition ? k.condition : 'AND'}`;
```
scottlepp, line 7: if there is no condition will this apply an unnecessary AND?  seems like the last filter should not apply a condition

if the query already has a where clause this probably doesn't work?

scottlepp (changes requested):
what happens if the query already has a where clause?

src/data/adHocFilter.ts
```diff
@@ -0,0 +1,20 @@
+
+export function applyAdHocFilter(rawSql: string, adHocFilters: AdHocVariableFilter[]): string {
+  if (rawSql != '' && adHocFilters && Object.keys(adHocFilters).length > 0) {
```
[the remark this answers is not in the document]
bossinc: oh nice! you are right. I was using records before

src/data/adHocFilter.ts
bossinc: Oh wait! What if we apply adhoc filters to inline queries. I am not sure if there would ever be a case where you are wanting an adhoc filter to apply to your outer query if there is an inline one. @scottlepp what do you think?

scottlepp: I was thinking we could just apply it to the outer query.  I'm guessing ad hoc filters won't be used with complex nested queries.  If the outer query ( or any simple query ) already has a where clause, we would just append to that.

scottlepp (approved):

PR #41, opened 2022-01-20
Better macro docs wording. Fix unix milliseconds bug. Remove table and column macro

bossinc:
Added examples and better descriptions for macro docs
Fixed but that was converting from time range to milliseconds instead of seconds
Removed `$__table` and `$__column`. We currently are not setting them in the backend. These macros will be added back when there is a visual query editor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,13 +96,11 @@ FROM test_data
 WHERE $__timeFilter(date_time)
 ```
 
-| Macro example | Description |
-| -- | --|
-| *$__timeFilter(dataRow)* | Will be replaced by a time range filter using the specified name. |
-| *$__fromTime* | Replaced by the start of time range in ms wrapped by toDateTime function. Example: toDateTime(intDiv(1415792726371,1000)) |
-| *$__toTime* | Replaced by the end of time range in ms wrapped by toDateTime function. Example: toDateTime(intDiv(1415792726371,1000)) |
-| *$__table* | Will be replaced by the table in use. |
-| *$__column* | Will be replaced by the column in use. |
+| Macro                       | Description                                                                                                      | Output example                                    |
+| --------------------------- | ---------------------------------------------------------------------------------------------------------------- | ------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


pkg/plugin/driver.go
```diff
@@ -116,10 +116,9 @@ func (h *Clickhouse) Converters() []sqlutil.Converter {
...
 func (h *Clickhouse) Macros() sqlds.Macros {
 	return map[string]sqlds.MacroFunc{
-		"fromTime":   macros.FromTimeFilter,
-		"toTime":     macros.ToTimeFilter,
-		"timeFilter": macros.TimeFilter,
-		"table":      macros.Table,
-		"column":     macros.Column,
+		"fromTime":        macros.FromTimeFilter,
+		"toTime":          macros.ToTimeFilter,
+		"timeFilterMilli": macros.TimeFilterMilli,
```
scottlepp, line 121: suggest "timeFilter_ms" to align with global var convention:  https://grafana.com/docs/grafana/latest/variables/variable-types/global-variables/#__interval_ms

README.md
```diff
@@ -96,13 +96,12 @@ FROM test_data
...
-| *$__toTime* | Replaced by the end of time range in ms wrapped by toDateTime function. Example: toDateTime(intDiv(1415792726371,1000)) |
-| *$__table* | Will be replaced by the table in use. |
-| *$__column* | Will be replaced by the column in use. |
+| Macro                            | Description                                                                                                                      | Output example                                          |
+| -------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------- |
+| *$__timeFilter(columnName)*      | Replaced by a conditional that filters the data (using the provided column) based on the time range of the panel in seconds      | `time >= '1480001790' AND time <= '1482576232' )`       |
+| *$__timeFilterMilli(columnName)* | Replaced by a conditional that filters the data (using the provided column) based on the time range of the panel in milliseconds | `time >= '1480001790671' AND time <= '1482576232479' )` |
```
scottlepp, line 102: $__timeFilter_ms like https://grafana.com/docs/grafana/latest/variables/variable-types/global-variables/#__interval_ms

scottlepp (commented):
would just use ms rather than milli similar to grafana global vars

scottlepp (approved):

PR #48, opened 2022-01-27
Visual Query Builder

scottlepp:
A visual query builder.  3 modes:  Table, Aggregate, Timeseries

**Table**: Simple queries
Screen Shot 2022-01-27 at 2 49 34 PM

**Aggregate**:  A bit more advanced, with aggregate functions and grouping
Screen Shot 2022-01-27 at 4 06 53 PM

**Timeseries**:  Builder focused on Timeseries queries and returning timeseries format
Screen Shot 2022-01-27 at 4 08 43 PM


bossinc:
I have a few notes after playing with it:

1. When you switch from the sql editor back to the query builder the confirm option reads `yes` instead of `continue` or something like that.
2. In the `show as` options, `timeseries` should be `time series`
3. I am thinking `show as` might be best as the first option above database. What's the reasoning putting it next to table?
3. When you switch between `table`, `aggregate`, and `time series` the data is deleted that is relevant to the one you switch off. (not sure if this is a big deal)
4. Can selecting different `show as` options change the visualization, so switching to time series will show time series visualization?
5. I think a `select all` button for fields might be good or if no option is selected maybe do wild card.
6. If we could input anything into each field we could make inline queries. I'm not sure if we want to avoid complex queries with the visual editor.

bossinc:
When using Time Series a comma is added to select.
image

bossinc:
Also a comma is added after the select field when using `Aggregate`

scottlepp:
@bossinc with time series you have to choose a time field

PR #49, opened 2022-01-28
AST inner queries

bossinc:
This should handle cases where inner queries are used anywhere, not just FROM clauses like before.
Also creating an ad hoc filter option query will not be as restrictive. Ad hoc filter queries now use the AST.

### DIFF
--- a/src/data/adHocFilter.ts
+++ b/src/data/adHocFilter.ts
@@ -41,15 +41,21 @@ export class AdHocFilter {
   }
 
   private applyFiltersToAST(ast: AST, whereClause: string): AST {
-    if (!ast || !ast.get('FROM')) return ast;
+    if (!ast || !ast.get('FROM')) {
+      return ast;
+    }
 
     const tableName = this._targetTable;
-    if (!tableName) return ast;
+    if (!tableName) {
+      return ast;
+    }
 
     for (let clause of ast.get('FROM')!) {
       if (typeof clause === 'string') {
         const tableRE = RegExp(`\\b${tableName}\\b`, 'g');
-        if (!clause.match(tableRE)) continue;
+        if (!clause.match(tableRE)) {
+          continue;
+        }
         const where = ast.get('WHERE');
         // If there is no defined WHERE clause create one
         // Else add an ad hoc filter to the existing WHERE clause
@@ -58,7 +64,11 @@ export class AdHocFilter {
           // example: "(SELECT * FROM table) as r" will have a FROM clause of "table) as r". We need ") as r" to be after the new WHERE clause
 
           // first we get the remaining part of the FROM phrase. ") as r"
-          const fromPhraseAfterTableName = ast.get('FROM')![ast.get('FROM')!.length - 1]!.toString().trim().substring(tableName.length);
+          const fromPhrase = ast.get('FROM');
+          const fromPhraseAfterTableName = fromPhrase!
+            [fromPhrase!.length - 1]!.toString()
+            .trim()
+            .substring(tableName.length);
           // apply the remaining part of the FROM phrase to the end of the new WHERE clause
           ast.set('WHERE', [`${whereClause} ${fromPhraseAfterTableName}`]);
           // set the FROM clause to only have the table name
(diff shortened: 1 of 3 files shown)


src/data/adHocFilter.ts
```diff
@@ -26,37 +36,60 @@ export class AdHocFilter {
...
-  // FROM phrases might contain more than just the table name
-  private getTableName(fromPhrase: string): string {
-    return fromPhrase
-      .trim()
-      .split(' ')[0]
-      .replace(/(;|\(|\))/g, '');
+    // Each node in the AST needs to be checked to see if ad hoc filters should be applied
+    ast.forEach((clauses: Clause[]) => {
+      for (let c of clauses) {
+        if (c !== null && typeof c !== 'string') {
```
scottlepp, line 86: isString lodash

src/data/adHocFilter.ts
```diff
@@ -26,37 +36,60 @@ export class AdHocFilter {
...
+      return ast;
+    }
+
+    const tableName = this._targetTable;
+    if (!tableName) {
+      return ast;
+    }
+
+    for (let clause of ast.get('FROM')!) {
+      if (typeof clause === 'string') {
```
scottlepp, line 54: isString lodash

src/data/adHocFilter.ts
```diff
@@ -1,13 +1,23 @@
...
+    const ast = sqlToAST(query);
+    this.setTargetTableFroAST(ast);
+  }
+
+  private setTargetTableFroAST(ast: AST) {
+    if (!ast.get('FROM')) {
+      return;
+    }
+    const from = ast.get('FROM')![0];
+    if (typeof from === 'string') {
```
scottlepp, line 16: isString lodash

scottlepp (changes requested):
I think there is one case to change from let to const.

might want to use isString.  handles primitives and String objects

src/data/ast.ts
```diff
@@ -9,69 +10,134 @@ export default function sqlToAST(sql: string): Map<string, Clause> {
...
-  clauses.set('OFFSET', null);
-  clauses.set('UNION', null);
-  clauses.set('INTERSECT', null);
-  clauses.set('EXCEPT', null);
-  clauses.set('INTO OUTFILE', null);
-  clauses.set('FORMAT', null);
+function getASTBranches(sql: string): Clause[] {
+  const clauses: Clause[] = [];
+  const re = /\b(AND|OR|,)\b/gi;
+  let bracketCount = 0;
```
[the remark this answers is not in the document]
bossinc: I like this! I think it would be best if it was `const bracket = {count, lastCount, phrase}` because regExp, index and lastPhraseIndex do not have a strong relation with bracket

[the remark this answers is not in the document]
scottlepp: maybe the object is called ParserState or something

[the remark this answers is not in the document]
bossinc: I think it looks a lot better with the new `bracket` var. Do you think we still need a parserState object? I'm not strongly opinionated either way

src/data/adHocFilter.ts
```diff
@@ -26,37 +37,60 @@ export class AdHocFilter {
...
-      if (tableName !== this._targetTable) {
-        return ast;
-      }
-      // If there is no defined WHERE clause create one
-      // Else add an ad hoc filter to the existing WHERE clause
-      if (ast.get('WHERE') === null) {
-        ast.set('FROM', ` ${tableName} `);
-        // set where clause to ad hoc filter and add the remaining part of the from phrase to the new WHERE phrase
-        return ast.set('WHERE', `${whereClause} ${fromPhrase.substring(tableName.length)}`);
+  private applyFiltersToAST(ast: AST, whereClause: string): AST {
```
scottlepp, line 44: this doesn't look specific to adHocFilters.  could put it in ast.ts?

scottlepp (approved):

src/data/adHocFilter.ts
bossinc: That sounds reasonable to me 👍

src/data/ast.ts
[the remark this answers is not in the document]
scottlepp: its not needed,  maybe a bit cleaner to have the state in one object

PR #62, opened 2022-02-03
Remove conditional alls

bossinc:
When processing template variables, this will removes template variable conditionals from WHERE clauses when template variables have `All` selected.

Updated the AST to have each phrase separator, like `,` `AND` `OR`, to be placed in its own array element.
Here is an example:
Input: `SELECT time, data FROM table WHERE data = stuff AND time < today`
Old AST structure: 
```
SELECT: ['time, data'],
FROM: ['table'],
WHERE: ['data = stuff AND time < today']
```

New AST structure:
```
SELECT: ['time', ',', 'data'],
FROM: ['table'],
WHERE: ['data = stuff', 'AND', 'time < today']
```

### DIFF
--- a/src/components/editor.ts
+++ b/src/components/editor.ts
@@ -1,11 +1,11 @@
-import SqlToAST from 'data/ast';
+import sqlToAST from 'data/ast';
 import { Format } from 'types';
 import { isString } from 'lodash';
 
 export const getFormat = (sql: string): Format => {
   // convention to format as time series
   // first field as "time" alias and requires at least 2 fields (time and metric)
-  const ast = SqlToAST(sql);
+  const ast = sqlToAST(sql);
   const selectList = ast.get('SELECT') || [];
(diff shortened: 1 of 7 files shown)


src/components/editor.ts
```diff
@@ -1,19 +1,15 @@
-import sqlToAST from 'data/ast';
+import SqlToAST from 'data/ast';
 import { Format } from 'types';
 import { isString } from 'lodash';
 
 export const getFormat = (sql: string): Format => {
   // convention to format as time series
   // first field as "time" alias and requires at least 2 fields (time and metric)
-  const ast = sqlToAST(sql);
+  const ast = SqlToAST(sql);
```
scottlepp, line 8: why are we capitalizing functions?

scottlepp (changes requested):
Not sure why we are using pascal case for functions.   I believe Grafana follows camel case.

scottlepp (approved):

Issue #65, opened 2022-02-05
[Feature request] Improving Ad-hoc filter flexibility


heanlan:
In current ad-hoc filter feature, we need to include a valid ClickHouse query. The query results will be used to populate the ad-hoc filter's selectable filters. 

In certain use cases where users would like to have selectable filters on all the available fields, say ~40 fields. It's not very flexible to write a query with all the 40 fields name. Is that possible for us to have something like, we don't need to write a query, and by default, all the fields in the table are selectable?

PR #67, opened 2022-02-10
ad hoc from database or table

### LINKED ISSUE #52: Ad-hoc queries - improve flexibility
The current approach to ad-hoc queries is restrictive. It requires the user to specify a query for populating of ad-hoc values - thus in turn limiting the filter to a single column. Whilst we can create multiple ad-hoc filters (one for each column we wish to filter on), this doesn't provide much over normal query variables - other than automatic injection of clauses.

Whilst this approach makes sense for normal OLTP databases, where DISTINCT is an expensive operation, this doesn't make a lot of sense for ClickHouse where DISTINCT  is a reasonable operation.

We therefore propose that the ad-hoc filter allows selection of any column - identifying the list via

 `SELECT name, type FROM system.columns WHERE database  IN ('db_name'`)

`db_name` could potentially be the default database associated with the datasource - or maybe overriden in the ad-hoc query configuration with a comma separated list.

This will populate a drop down of columns. On selection of a column, a DISTINCT query can identify the top values to populate a drop down. The operator selection can inturn be based on the column type.

Ideally the value drop down of top values will not be infinite - this will cause browser performance issues. Initially, id propose a configurable top N and allow filtering i.e. the user can type text and this be used in `LIKE` for the `DISTINCT`.

scottlepp:
This allows populating ad-hoc variables from:

1. default - all databases/tables if no default database has been set ( not ideal )
2. all tables from the default database
3. helper variable override - from databases (comma delimited) set on the helper variable
4. helper variable override - from a database.table set on the helper variable

Fixes #52 #65

### DIFF
--- a/src/data/CHDatasource.ts
+++ b/src/data/CHDatasource.ts
@@ -19,7 +19,7 @@ export class Datasource extends DataSourceWithBackend<CHQuery, CHConfig> {
   annotations = {};
   settings: DataSourceInstanceSettings<CHConfig>;
   adHocFilter: AdHocFilter;
-  skipAdHoc = false; // don't apply adhoc filters to the query
+  skipAdHocFilter = false; // don't apply adhoc filters to the query
 
   constructor(instanceSettings: DataSourceInstanceSettings<CHConfig>) {
     super(instanceSettings);
@@ -53,11 +53,11 @@ export class Datasource extends DataSourceWithBackend<CHQuery, CHConfig> {
     let rawQuery = query.rawSql || '';
     // we want to skip applying ad hoc filters when we are getting values for ad hoc filters
     const templateSrv = getTemplateSrv();
-    if (!this.skipAdHoc) {
+    if (!this.skipAdHocFilter) {
       const adHocFilters = (templateSrv as any)?.getAdhocFilters(this.name);
       rawQuery = this.adHocFilter.apply(rawQuery, adHocFilters);
     }
-    this.skipAdHoc = false;
+    this.skipAdHocFilter = false;
     rawQuery = removeConditionalAlls(rawQuery, templateSrv.getVariables());
     return {
       ...query,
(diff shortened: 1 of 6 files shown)


src/data/CHDatasource.ts
```diff
@@ -159,15 +191,59 @@ export class Datasource extends DataSourceWithBackend<CHQuery, CHConfig> {
...
-    const rawSql = this.templateSrv.replace('$clickhouse_adhoc_query');
-    if (rawSql === '$clickhouse_adhoc_query') {
-      return new ArrayDataFrame([]);
+  private async fetchTags(): Promise<Tags> {
+    const tagSource = this.getTagSource();
+    this.skipAdHocFilter = true;
+
+    if (tagSource.source === undefined) {
+      this.adHocFilter.setTargetTable('default');
+      const rawSql = 'SELECT name, type, table FROM system.columns';
```
bossinc, line 200: Do we want to limit these results too?

src/data/CHDatasource.ts
```diff
@@ -159,15 +191,59 @@ export class Datasource extends DataSourceWithBackend<CHQuery, CHConfig> {
...
+    source = source === ADHOC_VAR ? defaultDatabase! : source;
+    if (source.toLowerCase().startsWith('select')) {
+      return { type: TagType.query, source };
     }
+    if (!source.includes('.')) {
+      const sql = `SELECT name, type, table FROM system.columns WHERE database IN ('${source}')`;
+      return { type: TagType.schema, source: sql, from: source };
+    }
+    const [db, table] = source.split('.');
+    const sql = `SELECT name, type, table FROM system.columns WHERE database IN ('${db}') AND table = '${table}'`;
```
bossinc, line 236: Do we want to limit this query as well as the one a few lines above?

bossinc (approved):

src/data/CHDatasource.ts
scottlepp: We may have to add some capability to filter/limit.  They can always override this for now with the helper variable.

PR #77, opened 2022-02-15
Sql editor to query builder

bossinc:
When switching from SQL editor to the query builder it will try to convert the sql into a visual query.
This PR also fixes a bug with the AST not processing `,` correctly.

### DIFF
--- a/src/components/queryBuilder/GroupBy.tsx
+++ b/src/components/queryBuilder/GroupBy.tsx
@@ -19,6 +19,12 @@ export const GroupByEditor = (props: GroupByEditorProps) => {
     setIsOpen(false);
     setGroupBy(e.map((item) => item.value!));
   };
+  // Add selected value to the list if it does not exist.
+  groupBy.forEach(x => {
+    if (!columns.some(y => y.value === x)) {
+      columns.push({ value: x, label: x});
+    }
+  });
   return (
     <div className="gf-form">
       <InlineFormLabel width={8} className="query-keyword" tooltip={tooltip}>
@@ -34,6 +40,7 @@ export const GroupByEditor = (props: GroupByEditorProps) => {
           onChange={onChange}
           onBlur={() => props.onGroupByChange(groupBy)}
           value={groupBy}
+          allowCustomValue={true}
         />
       </div>
     </div>
--- a/src/components/queryBuilder/GroupBy.tsx
+++ b/src/components/queryBuilder/GroupBy.tsx
@@ -20,11 +20,7 @@ export const GroupByEditor = (props: GroupByEditorProps) => {
     setGroupBy(e.map((item) => item.value!));
   };
   // Add selected value to the list if it does not exist.
-  groupBy.forEach((x) => {
-    if (!columns.some((y) => y.value === x)) {
-      columns.push({ value: x, label: x });
-    }
-  });
+  groupBy.filter((x) => !columns.some((y) => y.value === x)).forEach((x) => columns.push({ value: x, label: x }));
   return (
     <div className="gf-form">
       <InlineFormLabel width={8} className="query-keyword" tooltip={tooltip}>
--- a/src/components/queryBuilder/OrderBy.tsx
+++ b/src/components/queryBuilder/OrderBy.tsx
@@ -8,6 +8,7 @@ import {
   FullField,
   BuilderMode,
   BuilderMetricField,
+  SqlBuilderOptionsAggregate,
 } from './../../types';
 import { selectors } from './../../selectors';
 import { styles } from '../../styles';
@@ -45,6 +46,7 @@ const OrderByItem = (props: {
         width={20}
         options={columns}
         onChange={(e) => onOrderBySortFieldUpdate(e.value!)}
+        allowCustomValue={true}
       ></Select>
       <Select<OrderByDirection>
         value={orderByItem.dir}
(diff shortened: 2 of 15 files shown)


src/components/queryBuilder/utils.ts
```diff
@@ -75,10 +79,11 @@ const getAggregationQuery = (
     })
     .join(', ');
   if (groupBy && groupBy.length > 0) {
-    metricsQuery = groupBy.map((g) => `${g}`).join(', ') + ', ' + metricsQuery;
-  }
-  if (metricsQuery !== '' && selected !== '') {
-    selected = `${selected},`;
+    metricsQuery =
+      groupBy
+        .map((g) => `${g}`)
```
scottlepp, line 84: this map doesn't seem necessary

src/components/queryBuilder/utils.ts
```diff
@@ -259,6 +264,195 @@ export const getSQLFromQueryOptions = (options: SqlBuilderOptions): string => {
...
 };
 
+export function getQueryOptionsFromSql(sql: string): SqlBuilderOptions {
+  const ast = sqlToAST(sql);
+  const fromPhrase = ast.get('FROM');
+  if (!fromPhrase || fromPhrase.length > 1 || !isString(fromPhrase[0])) {
+    return {} as SqlBuilderOptions;
+  }
+
+  const databaseAndTable = fromPhrase[0].trim().split('.');
```
scottlepp, line 274: if they set a default database, then the sql query doesn't require database.table.  they can just use table

```
select foo from bar
```

src/components/queryBuilder/utils.ts
```diff
@@ -259,6 +264,195 @@ export const getSQLFromQueryOptions = (options: SqlBuilderOptions): string => {
...
+  if (groupBy && groupBy.length > 0) {
+    (builder as SqlBuilderOptionsAggregate).groupBy = groupBy;
+  }
+  return builder;
+}
+
+function getFiltersFromAst(whereClauses: Clause[]): Filter[] {
+  // first condition is always AND but is not used
+  const filters: Filter[] = [{ condition: 'AND' } as Filter];
+  for (let c of whereClauses) {
```
scottlepp, line 335: const

src/components/queryBuilder/utils.ts
```diff
@@ -259,6 +264,195 @@ export const getSQLFromQueryOptions = (options: SqlBuilderOptions): string => {
...
+      hasTo = true;
+    }
+  }
+  return hasFrom && hasTo;
+}
+
+function getMetricsFromAst(selectClauses: Clause[]): { metrics: BuilderMetricField[]; fields: string[] } {
+  const metrics: BuilderMetricField[] = [];
+  const fields: string[] = [];
+  for (let c of selectClauses) {
```
scottlepp, line 424: const

scottlepp (changes requested):
Looking good!  The only thing that needs to be changed is the expectation of database.table.  If a default database is set they don't have to include database. in the query.

src/components/queryBuilder/GroupBy.tsx
```diff
@@ -19,6 +19,12 @@ export const GroupByEditor = (props: GroupByEditorProps) => {
     setIsOpen(false);
     setGroupBy(e.map((item) => item.value!));
   };
+  // Add selected value to the list if it does not exist.
+  groupBy.forEach((x) => {
```
scottlepp, line 23: nit.  .filter would probably be a bit cleaner

src/components/queryBuilder/OrderBy.tsx
```diff
@@ -171,10 +174,18 @@ export const getOrderByFields = (
...
+      break;
     case BuilderMode.List:
     default:
-      return fieldsList.map((m) => {
+      values = fieldsList.map((m) => {
         return { value: m.name, label: m.label };
       });
   }
+  // Add selected value to the list if it does not exist.
+  (builder as SqlBuilderOptionsAggregate).orderBy?.forEach((x) => {
```
scottlepp, line 185: again.  .filter may be a bit cleaner.  any time you have an if condition within a for loop think of using .filter

src/components/queryBuilder/utils.ts
```diff
@@ -259,6 +263,195 @@ export const getSQLFromQueryOptions = (options: SqlBuilderOptions): string => {
...
+    return 'datetime';
+  }
+  return '';
+}
+
+function isWithInTimeRangeFilter(phrases: string[]): boolean {
+  if (!phrases || phrases.length === 0) {
+    return false;
+  }
+  let hasFrom = false;
```
scottlepp, line 408: const has = {from: false, to: false} might be a bit cleaner

src/components/queryBuilder/utils.ts
```diff
@@ -259,6 +263,195 @@ export const getSQLFromQueryOptions = (options: SqlBuilderOptions): string => {
...
+
+function getMetricsFromAst(selectClauses: Clause[]): { metrics: BuilderMetricField[]; fields: string[] } {
+  const metrics: BuilderMetricField[] = [];
+  const fields: string[] = [];
+  for (const c of selectClauses) {
+    if (!isString(c) || !c.trim() || c.trim() === ',') {
+      continue;
+    }
+    let isMetric = false;
+    for (const agg of Object.values(BuilderMetricFieldAggregation)) {
```
scottlepp, line 428: probably another case where .filter and .map would make this a bit more readable

scottlepp (approved):
LGTM - seems we could use .filter and .map instead of for loops in a few cases.  up to you

PR #88, opened 2022-02-23
Fix conditional all

### LINKED ISSUE #86: Repeating panel queries don't handle the "All" selection correctly
**Describe the bug**

When using a multi-valued dashboard variable and a repeating panel, the wrong value for the variable is substituted in each instance of the panel when the "All" value is selected. Or, more precisely, no value is substituted and the condition using the repeating variable is just dropped.

I suspect this isn't a bug with this particular datasource, maybe instead a bug with the sqlds library or somewhere else in the stack. I unfortunately don't know enough about Grafana to really track it down to the right place.

**Versions**

Grafana v8.4.1
Plugin built from revision fd5cdcf5c054fe8eef40915dae1145cfdbc7a0a7 aka. current latest.

**Query**

Dashboard showing this issue: https://mon.dolphin-emu.org/d/rvvg02qGk/analytics-quirks-report

Base query set for the repeating panel:
```sql
SELECT
  gameid AS `Game ID`,
  COUNT(*) AS `Count`
FROM default.event
WHERE
  type = 'quirk' AND
  quirk = '$quirk' AND
  gameid IS NOT NULL AND
  $__timeFilter(ts)
GROUP BY gameid
ORDER BY `Count` DESC
```

When selecting one or multiple value, but not "All", a sample substituted query (correct, note that it has the `AND quirk = ...`):
```sql
SELECT gameid AS `Game ID` , COUNT(*) AS `Count` FROM default.event WHERE type = 'quirk' AND quirk = 'uses-DVDLowOffset' AND gameid IS NOT NULL AND ts >= '1645024345' AND ts <= '1645629145' GROUP BY gameid ORDER BY `Count` DESC
```

When selecting the "All" value, every single panel has the same substituted query (broken, it doesn't have any filter on `quirk`, should be same as the previous case):
```sql
SELECT gameid AS `Game ID` , COUNT(*) AS `Count` FROM default.event WHERE gameid IS NOT NULL AND ts >= '1645024431' AND ts <= '1645629231' GROUP BY gameid ORDER BY `Count` DESC
```

**To Reproduce**

1. Add a multi-valued variable to a dashboard.
2. Create a repeating panel using the variable as a query filter.

**Expected behavior**

Every panel properly substitutes the right value for the repetition variable in the "All" selection case, instead of stripping out the filter.

**Screenshots**

Dashboard link (same as linked above): https://mon.dolphin-emu.org/d/rvvg02qGk/analytics-quirks-report

**Additional context**

n/a

scottlepp:
when removing conditional all, only remove lines from the where clause containing a variable
Fixes #86

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.12.2
+
+Fix - when removing conditional all, only remove lines with variables
+
 ## 0.12.1
 
 Fix - handle large decimals properly
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clickhouse-datasource",
-  "version": "0.12.1",
+  "version": "0.12.2",
   "description": "Clickhouse Datasource",
   "scripts": {
     "build": "grafana-toolkit plugin:build",
--- a/src/data/ast.ts
+++ b/src/data/ast.ts
@@ -118,7 +118,7 @@ export function removeConditionalAllsFromAST(ast: AST, queryVarNames: string[]):
   if (where) {
     for (let i = 0; i < where.length; i++) {
       const c = where[i];
-      if (isString(c) && queryVarNames.some((v) => c.includes(v))) {
+      if (isString(c) && queryVarNames.some((v) => c.includes(v)) && c.includes('$')) {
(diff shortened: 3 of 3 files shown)


codeincarnate (approved):
LGTM

PR #135, opened 2022-06-13
able to use timeFilter in the adhoc query

### LINKED ISSUE #120: Improve Adhoc filter
**Is your feature request related to a problem? Please describe.**
When you have a very large dataset spanning over multiple months/years, the adhoc query filter generates a single query:
`SELECT DISTINCT $column FROM $table LIMIT 1000`
This query can be very expensive as it can span over millions/billions of rows.
Also it'd be great to order the response by count or something else.

**Describe the solution you'd like**
We should add the timerange filter of the current dashboard into the adhoc filter.
So let's say your dashboard is doing now - 15 min the query should be something like:
`SELECT DISTINCT $column FROM $table WHERE $__timeFilter($timecolumn) LIMIT 1000`
Here the $__timeFilter would replace the query with the proper timerange.

Obviously you can't necessary know the name of the timefiltering column, so we should create a new constant variable like `clickhouse_adhoc_query_timecolumn` where user would specify the timecolumn to be used.

In general customising the query filter generated would be very interesting.
Another option would be to create a variable like called `$adhoc_filtered_column` which would contain the value of the column selected by the user.
And then let the user manage how the query filter is created with a constant variable `clickhouse_adhoc_query_filter` where user would use the `$adhoc_filtered_column` and craft their own filters like:
`SELECT DISTINCT $adhoc_filtered_column FROM table WHERE $__timeFilter(timecolumn) ORDER BY COUNT() LIMIT 10`

I prefer option 2 as the user can optimise and customise easily how they want to use adhocfilter

bossinc:
fixes #120 
I did not implement `$column` as a variable. I am not sure what column should return. @gingerwizard 

image

### DIFF
--- a/src/data/CHDatasource.ts
+++ b/src/data/CHDatasource.ts
@@ -126,7 +126,7 @@ export class Datasource extends DataSourceWithBackend<CHQuery, CHConfig> {
     return new Promise((resolve) => {
       const req = {
         targets: [{ ...request, refId: String(Math.random()) }],
-        range: options?.range,
+        range: options ? options.range : (getTemplateSrv() as any).timeRange,
       } as DataQueryRequest<CHQuery>;
       this.query(req).subscribe((res: DataQueryResponse) => {
         resolve(res.data[0] || { fields: [] });


dsztykman:
@bossinc the `$column` should return the column name selected by the adhoc filtered.
Here in your example it would mean that the user selected the column name `machine_name`.
If we don't implement the column as a variable I'm not sure how this works, I mean your example here means that the clickhouse_adhoc_query will always return the distinct machine name regardless of what the user select.
(Unless I didn't understand)

Issue #138, opened 2022-06-16
Alerts: Error related to "non-positive" interval


limick:
**Describe the bug**
When trying to set up an alert using the ClickHouse datasource and clicking on "Test Rule", I get
```
code: 69, message: Value for second argument of function toStartOfInterval must be positive.
```
When I get the same message in the "Query" tab without any limitations on the interval, I sometimes get the same error, but setting a minimum interval of `1s` in the query options fixes it.
When I'm in the "Alerts" tab, it seems to ignore what I set in the query options.

**Versions**
- Grafana 8.0.7
- ClickHouse data source 1.1.2

**Query**
```sql
SELECT t AS time, cluster, usage_active AS " " FROM (
  SELECT $__timeInterval(time) AS t, CONCAT(cluster_name, ' ', instance_id) AS cluster, AVG(usage_active) as usage_active
  FROM telegraf
  WHERE name = 'cpu'
  AND environment_type = 'prod'
  AND $__timeFilter(time)
  GROUP BY t, instance_id, cluster_name
)
ORDER BY t
```

**To Reproduce**
1. Define a query using the ClickHouse datasource and set a positive interval in "Query Options". Check that the query works.
2. Go to the Alerts tab, define any alert and set up an alert as seen in screenshot below
3. Click on "Test Rule" and observe error message mentioned above

**Expected behavior**
Alert should use query with interval as defined in "Query Options".

**Screenshots**
<img width="865" alt="image" src="https://user-images.githubusercontent.com/6831616/174047541-e7251c99-ab96-4cec-add6-eee0b1d40f0d.png">

limick:
I opened a PR that prevents empty intervals from being used (which never works).
I think that's a good enough fix for now. In the long term, the alerts should probably use the same intervals as the queries.

PR #139, opened 2022-06-16
Macros: Don't use 0 interval in $__timeInterval macro

limick:
Since we currently only support an integer number of seconds, the right solution is to always pick at least 1 second. The value 0 will never work.

Fixes https://github.com/grafana/clickhouse-datasource/issues/138

### DIFF
--- a/pkg/macros/macros.go
+++ b/pkg/macros/macros.go
@@ -3,6 +3,7 @@ package macros
 import (
 	"errors"
 	"fmt"
+	"math"
 	"strings"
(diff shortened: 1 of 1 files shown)


limick:
The previous version of this PR contained an erroneous commit that caused the build to fail. That's now fixed.

bossinc (approved):
Makes sense to me! thank you

uebenes:
Question: is it possible to use intervals smaller then 1s in the macro?

PR #147, opened 2022-06-23
add Column macro and update version

bossinc:
I am not sure if there is a better way to make a frontend macro when we need the results of the adhoc filter.

### DIFF
--- a/src/data/CHDatasource.ts
+++ b/src/data/CHDatasource.ts
@@ -8,7 +8,7 @@ import {
   ScopedVars,
   vectorator,
 } from '@grafana/data';
-import { DataSourceWithBackend, getTemplateSrv } from '@grafana/runtime';
+import { DataSourceWithBackend, getTemplateSrv, locationService } from '@grafana/runtime';
 import { CHConfig, CHQuery, FullField, QueryType } from '../types';
 import { AdHocFilter } from './adHocFilter';
 import { isString, isEmpty } from 'lodash';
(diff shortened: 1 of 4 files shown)


src/data/CHDatasource.ts
```diff
@@ -66,6 +67,13 @@ export class Datasource extends DataSourceWithBackend<CHQuery, CHConfig> {
     };
   }
 
+  private addColumnTemplateVariable(scoped: ScopedVars, adHocFilters: any): ScopedVars {
+    if (!adHocFilters || !scoped) return scoped;
+    const col = adHocFilters.length > 0 ? adHocFilters[0].key : '';
+    scoped['__column'] = { text: col, value: col };
```
scottlepp, line 73: probably shouldn't mutate scoped here rather:
```
return {...scoped, '__column': { text: col, value: col }}
```

src/data/CHDatasource.ts
```diff
@@ -66,6 +67,13 @@ export class Datasource extends DataSourceWithBackend<CHQuery, CHConfig> {
     };
   }
 
+  private addColumnTemplateVariable(scoped: ScopedVars, adHocFilters: any): ScopedVars {
+    if (!adHocFilters || !scoped) return scoped;
+    const col = adHocFilters.length > 0 ? adHocFilters[0].key : '';
```
scottlepp, line 72: hmm not sure if having multiple adhoc filters is a valid scenario, but if so this could be a problem

scottlepp (changes requested):

scottlepp (commented):
Perhaps a bit more context/description as to why we need this....or the issue this relates to.

src/data/CHDatasource.ts
bossinc: Oh you are totally right. I didnt think about having an adhoc filter like this `SELECT DISTINCT machine_name, machine_group FROM mgbench.logs1`

bossinc:
> Perhaps a bit more context/description as to why we need this....or the issue this relates to.

https://github.com/grafana/clickhouse-datasource/pull/135#issuecomment-1161703207

Issue #141, opened 2022-06-21
Nullable(T): column: unhandled type Bool


AntonFriberg:
**Describe the bug**
Does not support querying Nullable(Bool) columns.

**Versions**
Grafana: v8.5.3
Plugin: v1.1.2

**Query**
A text example of the query ( if possible ).

```sql
SELECT toNullable(True)
```

**To Reproduce**
Execute query above in the explorer view towards a configured ClickHouse database.

**Expected behavior**
To be able to query Nullbable(Bool) with the same excellent behavior as other types. 

**Screenshots**
[
![Screenshot 2022-06-21 at 11-54-11 Explore - ClickHouse - Grafana](https://user-images.githubusercontent.com/14290517/174772304-09dd2e6e-4d9b-43ac-bd4e-4d3a551b7dbf.png)
](url)

bossinc:
@gingerwizard

gingerwizard:
This will be addressed when we update the driver to 2.0. Next release.

gingerwizard:
@bossinc you can close this - fixed in v2.

PR #151, opened 2022-07-01
Fix converters for driver v2 + JSON support

gingerwizard:
This fixes the converters for version 2.x of the driver - now in use on main. v2 requires strict use of types. 

We also add support for complex types - Tuple, Map and Nested. These are serialized to a JSON string. This thus supports the JSON type (results for this are returned as a Tuple). This allows uses to view JSON logs in the logs panel and explore.

Finally, to test all of this i've added significant tests so we don't miss issues with driver upgrades and types in the future.

This includes some IT tests - these use a Clickhouse instance (specified via env variables CLICKHOUSE_HOST and CLICKHOUSE_PORT - currently not secure). If these aren't specified we use a docker container. The former is appropriate for CI since it uses Drone (which uses docker itself for the tests and we don't support docker in docker yet @scottlepp @bossinc).  Can add support for a secure remote clickhouse instance if needed -this would actually allow us to test secure connections which is currently skipped.


gingerwizard:
Closes https://github.com/grafana/clickhouse-datasource/issues/141

gingerwizard:
The IT tests should skip if the env variables are not set and docker isn't available @scottlepp @bossinc - i can't see the drone failures - can you please share here?

bossinc:
@gingerwizard 
``` + mage -v lint
Running target: Lint
exec: golangci-lint run ./...
pkg/plugin/driver_test.go:130:11: Error return value of `conn.Exec` is not checked (errcheck)
	conn.Exec("DROP TABLE simple_table")
	         ^
pkg/converters/converters.go:232:2: `name` is unused (structcheck)
	name string
	^
pkg/converters/converters.go:233:2: `kind` is unused (structcheck)
	kind interface{}
	^
Error: running "golangci-lint run ./..." failed with exit code 1```

gingerwizard:
@bossinc we now support following env.

CLICKHOUSE_USERNAME
CLICKHOUSE_PASSWORD
CLICKHOUSE_SSL - set to `true` to use SSL
CLICKHOUSE_USE_DOCKER - set to `false` to use an external instance for IT tests

PR #159, opened 2022-08-01
Moving away from custom AST

bossinc:
- Changed how ad hoc filters work with a `settings` option provided in CH 22.7 (We will need to version gate ad hoc filters)
- Conditional alls are now handled with a conditional all function. The function checks if the second parameter is a template var set to `all`, if it then replaces the function with `1=1`, and if not set the function to the first parameter. Example: `$__conditionalAll(location in ($locations), $location)` $location=all: `1=1` $location='Dallas', 'Seattle': `location in ('Dallas', 'Seattle')`

The old custom AST is still used for the visual query builder. I have a branch with the visual query builder not using the old AST that will be merged shortly after this one. 😄


README.md
```diff
@@ -149,6 +150,8 @@ We distribute the following dashboards with the plugin. These are aimed at assis
 
 ### Ad Hoc Filters
 
+Ad hoc filters are only supported with version 22.7+ of ChickHouse.
```
gingerwizard, line 153: Typo. Not everyones type of house.

README.md
```diff
@@ -149,6 +150,8 @@ We distribute the following dashboards with the plugin. These are aimed at assis
...
+
```
gingerwizard, line 154: typo. Not everyones type of house.

scottlepp (changes requested):

gingerwizard:
@bossinc we need to test this when multiple tables are used in the dashboard and the filter only applies to one table. How does the query from the other table react?

PR #173, opened 2022-08-22
Explore in logs

bossinc:


scottlepp:
@bossinc is there an issue this fixes that we can add to the description?  We can already do something like this:  https://github.com/grafana/clickhouse-datasource#visualizing-logs-with-the-logs-panel

PR #186, opened 2022-08-31
ipv4 and ipv6 support + query timeout

gingerwizard:


pkg/converters/converters.go
```diff
@@ -352,3 +368,33 @@ func bigIntNullableConvert(in interface{}) (interface{}, error) {
 	f, _ := new(big.Float).SetInt(**v).Float64()
 	return &f, nil
 }
+
+func ipConverter(in interface{}) (interface{}, error) {
+	if in == nil {
+		return (*string)(nil), nil
```
bossinc, line 374: is this right? Why do we not return just`nil`?

bossinc (approved):

gingerwizard:
@bossinc addressed comment.

Also based on user feedback, I separated query timeout from dial timeout. The default for the former needs to be a lot higher and we should decouple otherwise `Save and Test` behaviour will be poor.

image

Issue #209, opened 2022-10-05
Query builder adds unnecessary AND prefix


gingerwizard:
**Describe the bug**
Query builder for time-series adds an AND prefix when adding a filter.

**Versions**
9.x Grafana
2.x plugin

Grafana cloud
**Expected behavior**

**Screenshots**
<img width="2048" alt="image" src="https://user-images.githubusercontent.com/12695796/194059070-7e0b16f6-063d-41f5-9a24-177b1c973355.png">

PR #213, opened 2022-10-09
add support for date32, datetime64

mshustov:
closes https://github.com/grafana/clickhouse-datasource/issues/209
add support for https://clickhouse.com/docs/en/sql-reference/data-types/date32 and https://clickhouse.com/docs/en/sql-reference/data-types/datetime64


src/components/queryBuilder/utils.ts
```diff
@@ -21,14 +21,15 @@ import {
...
 export const isNumberType = (type: string): boolean => {
-  const numericTypes = ['int', 'float', 'decimal'];
-  const match = numericTypes.find((t) => type.toLowerCase().includes(t));
-  return match !== undefined;
+  return ['int', 'float', 'decimal'].some((t) => type.toLowerCase().includes(t));
 };
+
 export const isDateType = (type: string): boolean => {
-  return ['date', 'datetime'].includes(type.toLowerCase());
+  return type.toLowerCase().includes('date');
```
mshustov, line 30: Maybe we should use `startsWith` instead to exclude matches against `Array(Date)`, tuples `Tuple(UInt8, Date)`?

src/components/queryBuilder/utils.ts
```diff
@@ -229,11 +230,15 @@ export const getSQLFromQueryOptions = (options: SqlBuilderOptions): string => {
         options.timeField,
         options.timeFieldType
       );
+      const trendFilters = getFilters(options.filters || []);
+
       if (isDateType(options.timeFieldType)) {
```
mshustov, line 235: Shouldn't we throw? Is there a valid case to end up in this branch without `options.timeFieldType` to be set?

src/components/queryBuilder/utils.spec.ts
```diff
@@ -1,5 +1,85 @@
...
-import { getQueryOptionsFromSql, getSQLFromQueryOptions } from './utils';
+import { getQueryOptionsFromSql, getSQLFromQueryOptions, isDateType, isNumberType } from './utils';
+
+describe('isDateType', () => {
+  it('returns true for Date type', () => {
+    expect(isDateType('Date')).toBe(true)
+    expect(isDateType('date')).toBe(true)
+  })
+  it('returns true for Nullable(Date) type', () => {
+    expect(isDateType('Nullable(Date)')).toBe(true)
```
mshustov, line 10: let me know if `Nullable(Date)` cannot be used as a timefield

src/components/queryBuilder/utils.ts
bossinc: Changing it to `startsWith` instead of `includes` sounds like a good idea.

src/components/queryBuilder/utils.ts
bossinc: You are right. We should throw if `options.timeFieldType` isn't valid

src/components/queryBuilder/utils.spec.ts
mshustov: @bossinc What about this one?

bossinc (approved):

PR #223, opened 2022-10-19
Update sqlds to 2.3.13

### LINKED ISSUE #195: Unexpected query error
**Describe the bug**

query with `WITH Clause` contains `$__timeFilter(time)` must have a black space before right parenthesis.
**Versions**
Grafana and Plugin versions.

**Query**
this query can work
```
with (select max(time) from unicost.bill where $__timeFilter(time) ) as mtime
SELECT cloud, sum(realPrice) cost FROM unicost.bill WHERE time=mtime  GROUP BY cloud ORDER BY cost DESC;
```
while this one can not
```
with (select max(time) from unicost.bill where $__timeFilter(time)) as mtime
SELECT cloud, sum(realPrice) cost FROM unicost.bill WHERE time=mtime  GROUP BY cloud ORDER BY cost DESC;
```
**To Reproduce**

**Expected behavior**
they are the same query.

**Screenshots**

**Additional context**

sasklacz:
Latest version includes some fixes to macro parsing. This fixes #195

### DIFF
--- a/pkg/macros/macros_test.go
+++ b/pkg/macros/macros_test.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/grafana/clickhouse-datasource/pkg/macros"
+	"github.com/grafana/clickhouse-datasource/pkg/plugin"
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
 	"github.com/grafana/sqlds/v2"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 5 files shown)


pkg/macros/macros_test.go
```diff
@@ -1,15 +1,36 @@
...
 
+type ClickhouseDriver struct {
+	sqlds.Driver
+}
+
+type MockDB struct {
+	ClickhouseDriver
+}
+
+func (h *ClickhouseDriver) Macros() sqlds.Macros {
```
sasklacz, line 23: @yesoreyeram is there a way to import those local Macros here ?

pkg/macros/macros_test.go
```diff
@@ -93,3 +114,31 @@ func TestMacroIntervalSeconds(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, "20", got)
 }
+
+// test sqlds query interpolation with clickhouse filters used
+func TestInterpolate(t *testing.T) {
```
sasklacz, line 119: I copied the test we have in sqlds but here it's running with clickhouse specific macros

pkg/macros/macros_test.go
sasklacz: Scott already suggested the latter

Issue #217, opened 2022-10-13
Display full text for metric errors


e-mars:
**Is your feature request related to a problem? Please describe.**
If there is an error in fetching data, the plugin only displays a generic error as in screenshotsbut not the full error received to help troubleshoot the issue.

**Describe the solution you'd like**
Display full error from ClickHouse

**Describe alternatives you've considered**
View logs or try to reproduce using CLI or curl for particular failing query.

rlazimi-dev:
Thank you @e-mars!

This issue also occurs for ClickHouse version:
```
22.2.2.1
```

and Grafana Version:
```
Version 8.3.2 (commit: afb9e8e5f3, branch: HEAD)
```

e-mars:
on roadmap: https://github.com/grafana/clickhouse-datasource/issues/206

gingerwizard:
@rlazimi-dev do you have any examples here? timeouts show as a context exceeded which defn needs to be improved. Other errors e.g. `SELECT x` are quite descriptive, however. Any other cases? @e-mars

gingerwizard:
timeouts here https://github.com/grafana/clickhouse-datasource/issues/187

gingerwizard:
Issue https://github.com/grafana/clickhouse-datasource/issues/206 not reproducible.

e-mars:
@gingerwizard 
example below:
<img width="1920" alt="error" src="https://user-images.githubusercontent.com/18219420/196676984-00ae19bd-3a76-4beb-b0a7-bc9c4847ba7a.png">

Issue #206, opened 2022-09-30
Provide error details in SQL editor


mshustov:
**Describe the bug**
No error details are provided when sending a non-valid query.

**Versions**
Grafana v9.1.6
plugin 2.0.1

**Query**
 for example, `SELECTffff`

**To Reproduce**
1. Go to Explore
2. Switch to SQL editor
3. Send `SELECTffff` query
4. See `Query error: 500 Internal Server Error`

**Expected behavior**
I'd expect an error message about invalid syntax, `unknown identifier`, `unknown table` to be provided. Debugging a typo in a query without a meaningful error message is not the best UX.

**Screenshots**
![2022-09-30_18-41-00](https://user-images.githubusercontent.com/3198181/193317470-93be997a-8aaa-416d-9b10-3496ff87f81b.png)

gingerwizard:
![image](https://user-images.githubusercontent.com/12695796/196518961-19efa882-9202-493b-91a9-0aee75380b20.png)

this is quite descriptive.

gingerwizard:
Same error on explore as the one above.

gingerwizard:
Versions
Grafana v9.0.1
plugin 2.0.1

Maybe we've dropped this info in 9.1.x. Or are you connecting over HTTP?

gingerwizard:
HTTP isn't great but still better than 500

![image](https://user-images.githubusercontent.com/12695796/196520096-9aaef913-d8b3-4257-a6f6-0c3a68547c06.png)

gingerwizard:
Hmm using 9.2.1 grafana it seems like this has regressed.

PR #226, opened 2022-10-20
use gzip for http

gingerwizard:
closes https://github.com/grafana/clickhouse-datasource/issues/217
closes https://github.com/grafana/clickhouse-datasource/issues/206

Currently, we use lz4 compression irrespective of the protocol. There is an issue in the go client (which ill PR) which means errors over HTTP aren't decompressed if they are compressed with LZ4 - preventing them from being shown to the user. Whilst we'll fix this in the client, we should really use gzip over HTTP - its a standard and is more likely to be proxy friendly.


gingerwizard:
Tested errors are reported on 9.1.x and 8.5.14.

gingerwizard:
@bossinc please review and release if +1. Minor only.

bossinc (approved):

Issue #241, opened 2022-11-21
`Port` should allow string and parse it as `int64` in Go


marcustut:
**Describe the bug**
I am using environment variables in my `datasources.yaml` file like the following:

```yaml
apiVersion: 1
datasources:
  - name: ClickHouse
    type: grafana-clickhouse-datasource
    jsonData:
      defaultDatabase: ${CLICKHOUSE_DATABASE}
      access: proxy
      port: ${CLICKHOUSE_PORT}
      server: ${CLICKHOUSE_HOST}
      username: ${CLICKHOUSE_USER}
      tlsSkipVerify: false
    secureJsonData:
      password: ${CLICKHOUSE_PASSWORD}
    isDefault: true
```

and the problem I encountered here `2022-11-22 02:32:35 logger=provisioning t=2022-11-21T18:32:35.913525135Z level=error msg="Failed to provision data sources" error="Datasource provisioning error: yaml: cannot decode !!str 
'${CLICKHOUSE_PORT}' as a !!int"`

The reason being when environment variables are injected, they're by default string. However, the Go code expects it to be an integer. 

https://github.com/grafana/clickhouse-datasource/blob/615978e7dce78e16a5a25da422be484950c1e1c4/pkg/plugin/settings.go#L14

A quick fix here would be changing 

```go
Port               int64  `json:"port,omitempty"`
``` 
to 
```go
Port               int64  `json:"port,string,omitempty"`
```

**Versions**
Grafana - v8.3.0-beta2
Plugin - v2.0.3

**To Reproduce**
1. Configure `datasources.yaml` with a string number in port.
2. Start grafana
3. See error

**Expected behavior**
Expected it to run as usual which it automatically converts to integer under the hood.

marcustut:
Also, for `bool` fields such as `Secure`, they should be able to parse from string "true" or "false" as well

aangelisc:
Hi @marcustut, I've opened a PR to tackle this issue! Unfortunately the boolean parsing problem is a known issue, see grafana/azure-data-explorer-datasource#442. This is actually a core Grafana issue which I've reopened (grafana/grafana#32061).

PR #243, opened 2022-11-23
allow port to be a string and parse it as int64 in Go #241

v-electrolux:
Allow port to be a string and parse it as int64 in Go

Considered all three solution from this [thread](https://stackoverflow.com/questions/21151765/cannot-unmarshal-string-into-go-value-of-type-int64) and chose the one with custom unmarshaler, because:
- solution with ",string" option work with "port": "443" but no works with "port": 443
- solution with json.Number seems to me much more complicated (in terms of how much line of codes i must change)


sasklacz:
Looks good to me but I'd like somebody with more go and this plugin experience to take a look as well.

scottlepp (commented):
Hmm.  I've never seen this issue.  Does this work?
```
port: !!int ${CLICKHOUSE_PORT}
```

If not I suppose we could do below.   ( and change the client side to send a string )
```
Port               int64  `json:"port,string,omitempty"`
```

EDIT: looks like an issue where the env variable isn't getting set

v-electrolux:
@scottlepp 
No, explicit type conversion does not work
```
Failed to start grafana. error: Datasource provisioning error: yaml: cannot decode !!str `${CLICKHOUSE_PORT}` as a !!int
```

I have following lines in my datasource all.yml file
```yaml
  - name: ClickHouse_Replica_1
    editable: true
    jsonData:
      port: !!int ${CLICKHOUSE_PORT}
      server: $CLICKHOUSE_HOST_01
      username: $CLICKHOUSE_USER
      tlsSkipVerify: true
    secureJsonData:
      password: $CLICKHOUSE_PASSWORD
```

And my doocker-compose for custom grafana image
```yaml
  grafana:
    image: grafana:9.2.6-custom.6
    environment:
      - GF_DATABASE_TYPE=postgres
      - GF_DATABASE_HOST=192.168.1.14:4009
      - GF_DATABASE_NAME=grafana
      - GF_DATABASE_USER=grafana
      - GF_DATABASE_PASSWORD=grafana
      - INFLUX_URL=http://192.168.1.14:8086
      - INFLUX_DB_NAME=test_db
      - INFLUX_ADMIN_USER=test
      - INFLUX_ADMIN_PASSWORD=test
      - CLICKHOUSE_PORT=9000
      - CLICKHOUSE_HOST_01=192.168.1.14
      - CLICKHOUSE_HOST_02=192.168.1.15
      - CLICKHOUSE_HOST_03=192.168.1.16
      - CLICKHOUSE_USER=gf_user
      - CLICKHOUSE_PASSWORD=gf_pass
    deploy:
      mode: replicated
      replicas: 1
```

bossinc:
Hello thank you for your PR! We decided to go in a different direction with #241 
Here is the PR that resolved this problem https://github.com/grafana/clickhouse-datasource/pull/293

PR #248, opened 2022-11-30
fix: Allow default protocol toggle value when saving in settings

### LINKED ISSUE #245: When creating data source, native protocol is selected in the UI, but is not applied unless explicitly clicked
**Describe the bug**
When creating data source, native protocol is selected in the UI, but when saving and testing the error "protocol is invalid, use native or http" appears. When selecting "HTTP" and then back "Native" it begins to work. Seems like the issue is with the default value.

**Versions**
Grafana: 9.2.6
Plugin: 2.0.3

**To Reproduce**
1. Go to data sources and create new clickhouse data source
2. When filling in settings, leave native protocol untouched
3. Click "Save & test"
5. Select "HTTP" protocol and then "Native" again
6. Click "Save & test"
7. No error appears

**Expected behavior**
No error appears when saving with untouched Native protocol.

**Screenshots**
image

Video: https://user-images.githubusercontent.com/1436174/203844631-d8a73156-b093-41c9-90c9-b159f60f0a72.mov

cletter7:
Fixes #245.

When creating data source, the default protocol toggle value is "Native". Allow user to keep it untouched (previously was needed to set to "HTTP" and then back to "Native" in order to make it work).

image

### DIFF
--- a/pkg/plugin/settings.go
+++ b/pkg/plugin/settings.go
@@ -51,6 +51,9 @@ func LoadSettings(config backend.DataSourceInstanceSettings) (settings Settings,
 	if strings.TrimSpace(settings.QueryTimeout) == "" {
 		settings.QueryTimeout = "60"
 	}
+	if settings.Protocol == "" {
+		settings.Protocol = "native"
+	}
 	val, ok := config.DecryptedSecureJSONData["password"]
 	if !ok {
 		return settings, settings.isValid()
@@ -67,9 +70,6 @@ func LoadSettings(config backend.DataSourceInstanceSettings) (settings Settings,
 	tlsClientKey, ok := config.DecryptedSecureJSONData["tlsClientKey"]
 	if ok {
 		settings.TlsClientKey = tlsClientKey
-	}
-	if settings.Protocol == "" {
-		settings.Protocol = "native"
 	}
 	return settings, settings.isValid()
 }
--- a/pkg/plugin/settings.go
+++ b/pkg/plugin/settings.go
@@ -34,9 +34,6 @@ func (settings *Settings) isValid() (err error) {
 	if settings.Port == 0 {
 		return ErrorMessageInvalidPort
 	}
-	if settings.Protocol != "http" && settings.Protocol != "native" {
-		return ErrorMessageInvalidProtocol
-	}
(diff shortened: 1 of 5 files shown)


pkg/plugin/settings.go
```diff
@@ -51,6 +51,9 @@ func LoadSettings(config backend.DataSourceInstanceSettings) (settings Settings,
 	if strings.TrimSpace(settings.QueryTimeout) == "" {
 		settings.QueryTimeout = "60"
 	}
+	if settings.Protocol == "" {
+		settings.Protocol = "native"
+	}
```
cletter7, line 56: @yesoreyeram, @scottlepp, I would appreaciate your advice on this change.
Even though it solves the issue, it might not be the best solution, the value from here will never be saved to the `jsonData`.
From what I understand, the current flow is the following:

```mermaid
  graph TD;
      A[client sends request for creating data source to the server]-->B;
      B[server creates DS and responds with empty jsonData]-->C;
      C[client uses empty jsonData on the config page]-->D;
      D[whenever user interacts with the config page, we update jsonData];
```

But if user never interacts with "protocol" radio buttons, protocol will never be put to the `jsonData`, though everything will still work because of the check on the backend. Maybe it would be better to have some optional method for datasources, that returns default config (`jsonData`). So that we can populate the default config on widget creation. In this case we can set protocol to `native` and it will exist in the `jsonData` even if user doesn't touch this setting. I find it weird that in the UI user sees "Native" protocol being set, but actually it's not set at all.
What do you think? Or maybe there is already a way to set default `jsonData` somehow on datasource creation?

scottlepp: hmm looks like it was already defaulting to native:  https://github.com/grafana/clickhouse-datasource/blob/main/pkg/plugin/driver.go#L104

cletter7: But `LoadSettings` was called before that: https://github.com/grafana/clickhouse-datasource/blob/main/pkg/plugin/driver.go#L81 and it was failing.

cletter7: Changed the code according to what we've discussed.

scottlepp (approved):
LGTM

PR #293, opened 2023-02-06
Improved variable parsing

### LINKED ISSUE #241: `Port` should allow string and parse it as `int64` in Go

aangelisc:
Will correctly parse variables typed as a number or boolean even when supplied as a string (as is the case with environment variables).

Also some docs updates as the `yandex` docker images are out of date.

Fixes #241

### DIFF
--- a/pkg/plugin/settings.go
+++ b/pkg/plugin/settings.go
@@ -11,7 +11,7 @@ import (
 // Settings - data loaded from grafana settings database
 type Settings struct {
 	Server             string `json:"server,omitempty"`
-	Port               int64  `json:"port,omitempty"`
+	Port               int64  `json:"port,string,omitempty"`
 	Username           string `json:"username,omitempty"`
 	DefaultDatabase    string `json:"defaultDatabase,omitempty"`
 	InsecureSkipVerify bool   `json:"tlsSkipVerify,omitempty"`
--- a/pkg/plugin/settings.go
+++ b/pkg/plugin/settings.go
@@ -49,10 +49,13 @@ func LoadSettings(config backend.DataSourceInstanceSettings) (settings Settings,
 		settings.Server = jsonData["server"].(string)
 	}
 	if jsonData["port"] != nil {
-		if port, err := strconv.ParseInt(jsonData["port"].(string), 0, 64); err == nil {
-			settings.Port = port
+		if port, ok := jsonData["port"].(string); ok {
+			settings.Port, err = strconv.ParseInt(port, 0, 64)
+			if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/plugin/settings.go
```diff
@@ -39,9 +40,79 @@ func (settings *Settings) isValid() (err error) {
...
+	var jsonData map[string]interface{}
+	if err := json.Unmarshal(config.JSONData, &jsonData); err != nil {
 		return settings, fmt.Errorf("%s: %w", err.Error(), ErrorMessageInvalidJSON)
 	}
+
+	if jsonData["server"] != nil {
+		settings.Server = jsonData["server"].(string)
+	}
+	if jsonData["port"] != nil {
+		if port, err := strconv.ParseInt(jsonData["port"].(string), 0, 64); err == nil {
```
asimpson, line 52: Is it always a string? Do we want to account for strings _and_ numbers?

aangelisc: Literally my latest commit 😄

asimpson (approved):
:clap: Nicely done!

Issue #232, opened 2022-11-01
DESC TABLE expression not work when table name contains dot


chenlujjj:
**Describe the bug**
When table name contains dot symbol, the DESC TABLE expression does not work

**Versions**
Grafana and Plugin versions.
grafana: 9.1.7
plugin: 2.0.2

**Query**
DESC TABLE db.a.b

table name is a.b

chenlujjj:
I have fixed this locally

gingerwizard:
try in backticks around the table name? e.g. 
```
DESC TABLE `db.a.b`
```

If not, can you raise a PR or suggest changes? @chenlujjj

chenlujjj:
My solution is:
```
DESC TABLE db.`a.b`
```
See this commit: https://github.com/chenlujjj/clickhouse-datasource/commit/c82391af7526b6adb039e7b5b267340790e6f7f8

But it contains some inrelevant changes. Maybe I can checkout another branch and make a clean PR.

PR #298, opened 2023-02-10
fix: quote table names with dots

slvrtrn:
* Resolves https://github.com/grafana/clickhouse-datasource/issues/232 - table names with dots work correctly now, as they are quoted with `""` now;
* Adds `"jsx": "react"` to `compilerOptions` in `tsconfig.json`, fixing WebStorm syntax highlighting issues.

I used @chenlujjj's [commit](https://github.com/chenlujjj/clickhouse-datasource/commit/c82391af7526b6adb039e7b5b267340790e6f7f8) as an example, though I opted for double-quote instead of a backtick to be compatible with `pgsql-ast-parser` library that is used in `ast.ts`, and added a couple more unit tests.


bossinc (approved):
Looks good!

PR #304, opened 2023-02-22
Add a predefined TimeRange filter if there is at least one DateTime* column

slvrtrn:
Demo: 

https://user-images.githubusercontent.com/3175289/220785466-ddedb45e-b80a-4f26-88af-dfc1f2236c83.mov

Full description:

There was a request from our users to restrict the query to the dashboard range automatically.

After some consideration, we decided to add a (removable) pre-defined filter that appears when we switch the tables if and only if we have at least one DateTime column. If we find some, the first DateTime column is picked, and we add the `WithInGrafanaTimeRange` filter automatically. 

What is interesting about this filter is that the column selection there is restricted to DateTime (could be also *64 or Nullable) types only, so it is convenient to switch to another column if we failed to pick the right one, and the type of the filter is `WithInGrafanaTimeRange` by default.

If a user does not need this type of filter, they can delete it and just add another one.

I also added a convenience ClickHouse container to the `docker-compose.yml`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,3 +14,14 @@ services:
       - TERM=linux
       - GF_DEFAULT_APP_MODE=development
       - GF_ENTERPRISE_LICENSE_TEXT=$GF_ENTERPRISE_LICENSE_TEXT
+
+  clickhouse:
+    image: 'clickhouse/clickhouse-server:${CLICKHOUSE_VERSION-23.1-alpine}'
+    container_name: 'grafana-clickhouse-server'
+    ports:
+      - '8123:8123'
+      - '9000:9000'
+    ulimits:
+      nofile:
+        soft: 262144
+        hard: 262144
--- a/src/components/queryBuilder/QueryBuilder.tsx
+++ b/src/components/queryBuilder/QueryBuilder.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import defaultsDeep from 'lodash/defaultsDeep';
 import { Datasource } from '../../data/CHDatasource';
 import { TableSelect } from './TableSelect';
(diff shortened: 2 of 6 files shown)


mshustov:
IMO the timerange filter should be applied by default without the necessity to introduce another UI element.

It's the default behavior of other datasources. For example, `elasticsearch`:
UI


query:
```
{"size":0,"query":{"bool":{"filter":[{"range":{"timestamp":{"gte":1677128427354,"lte":1677150027354,"format":"epoch_millis"}}}]}}
```

 WDYT @bossinc ? If you agree, could you point out a proper place for `timerange` integration?

slvrtrn:
@mshustov, how do you select the DateTime column in that case?

mshustov:
> how do you select the DateTime column in that case?

I suggest providing a column selection and setting the value to `WITHIN DASHBOARD TIME RANGE`. The current implementation in the PR doesn't allow a user to change value from `WITHIN DASHBOARD TIME RANGE`.

slvrtrn:
@mshustov 

Now it is like this:

https://user-images.githubusercontent.com/3175289/220968312-c7455632-a924-4024-987a-04479c6b3a9d.mov

WDYT?

mshustov:
@slvrtrn. Should the user have the ability to remove the filter? The current implementation doesn't support deletion.

slvrtrn:
I simplified the logic now; the deletion bug should be gone.

mshustov (dismissed):

src/components/queryBuilder/QueryBuilder.tsx
```diff
@@ -40,8 +41,21 @@ export const QueryBuilder = (props: QueryBuilderProps) => {
           fields.push({ name: '*', label: 'ALL', type: 'string', picklistValues: [] });
           setBaseFieldsList(fields);
 
+          const dateTimeFields = fields.filter((f) => isDateTimeType(f.type));
+          if (dateTimeFields.length > 0) {
```
asimpson, line 45: This works right now however if you remove all the filters you can't get our new filter back.

Instead of adding to the filter list here, what if we toggle some state that says we have a dateTimeField? Then inside `onFiltersChange` we can reintroduce the filter whenever filters are removed completely?

docker-compose.yml
```diff
@@ -14,3 +14,14 @@ services:
       - TERM=linux
       - GF_DEFAULT_APP_MODE=development
       - GF_ENTERPRISE_LICENSE_TEXT=$GF_ENTERPRISE_LICENSE_TEXT
+
+  clickhouse:
+    image: 'clickhouse/clickhouse-server:${CLICKHOUSE_VERSION-23.1-alpine}'
+    container_name: 'grafana-clickhouse-server'
+    ports:
+      - '8123:8123'
+      - '9000:9000'
```
asimpson, line 23: Thank you for this! :trophy:

asimpson (changes requested):
This is good!

I think I found one area where we have a user experience hole: removing all the filters. We should account for a user who removes the filters but then changes their mind.

src/components/queryBuilder/QueryBuilder.tsx
slvrtrn: > if you remove all the filters you can't get our new filter back.

That was the exact idea - a user is proposed with a predefined filter once, but they are free to delete it entirely, and they won't be prompted with it again.

If we constantly push the predefined filter when filters are empty, it will result in the situation when the user deletes the filter, it just appears again immediately (i.e. it is not possible to delete it), so I wanted to explicitly avoid this behaviour.

Please correct me if I am getting this wrong.

asimpson: Sorry for the delay on this. I was misunderstanding how you wanted this to work. I'm good with this how it is and I'll approve right away.

asimpson (approved):
Nicely done! :trophy:

Issue #295, opened 2023-02-07
toStartOfMonth() Returning Incorrect Values


dbernardi7:
**What happened**: When using `toStartOfMonth()` in a query in Grafana, the result of the query mimicked what `toEndOfMonth()` would produce with all the values indicating the last day of the month. However, when using the ClickHouse CLI the query returned expected beginning of the month values.

**What you expected to happen**: I would've expected to have the first date of the month returned with a `DateTime` input. For example if entering this: `2022-12-07 02:15:10`, I would expect to be returned this: `2022-12-01`.

**How to reproduce it (as minimally and precisely as possible)**: While connecting Grafana to a ClickHouse database have a field that is a `DateTime` data type. Query the field using the function `toStartOfMonth()`. The result will show the last day of the month of the date that you queried.

An example query:

SELECT toStartOfMonth( "DateTime Field" )
FROM datatable

**Screenshots**

![image](https://user-images.githubusercontent.com/86997293/217375782-9c273667-5b3e-49a5-9542-4d1c85f49476.png)

**Anything else we need to know?**: When using the clickhouse client to perform the same exact query, the correct result is shown. Screenshot below:

![image](https://user-images.githubusercontent.com/86997293/217389396-ba048b06-a737-4f67-929c-0f99bf1a4f4c.png)

**Environment**:

- Grafana version: v8.4.4 (fcfb01faeb)
- Plugin version:
- OS Grafana is installed on: NA
- User OS & Browser: Chrome on Mac OS X 10.15
- Others:

jkaflik:
Hi @dbernardi7,

The issue's root cause is simple, but I am unsure how we can mitigate it.
`toStartOfMonth` in ClickHouse return `Date` type. This type is decoded as a Go `time.Time` type in `clickhouse-go` library.

The problem is `Date` type is cast to `time.Time` with a local server timezone. In our case, it's Grafana server timezone.

If you go and pick in Grafana UI time picker the same timezone as you have in Grafana backend, results should be fine.

Most likely, this issue has to be fixed in `clickhouse-go`, so `Date` decode loads a user-defined timezone.

dbernardi7:
Hey @jkaflik,

Thank you so much for your prompt response, I really appreciate it!

I went ahead and changed the organizations timezone to UTC and ran a test query to ensure it worked, and sure enough it did!

[...]

PR #313, opened 2023-03-03
add timezone to a query

mshustov:
Attaches `iana` name of timezone  to each query.
Part of https://github.com/grafana/clickhouse-datasource/issues/295. The backend side of the plugin will use the timezone information to adjust the offset for `time.Time` value.
the timeonze is received from `request: DataQueryRequest`, if `request` doesn't provide one, fall back to the local timezone

### DIFF
--- a/src/data/CHDatasource.ts
+++ b/src/data/CHDatasource.ts
@@ -8,6 +8,8 @@ import {
   ScopedVars,
   VariableModel,
   vectorator,
+  getTimeZone,
+  getTimeZoneInfo,
 } from '@grafana/data';
 import { DataSourceWithBackend, getTemplateSrv } from '@grafana/runtime';
 import { Observable } from 'rxjs';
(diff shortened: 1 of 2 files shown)


src/data/CHDatasource.ts
```diff
@@ -184,11 +186,32 @@ export class Datasource extends DataSourceWithBackend<CHQuery, CHConfig> {
...
+  private getTimezone(request: DataQueryRequest<CHQuery>): string | undefined {
+    // timezone specified in the time picker
+    if(request.timezone) return request.timezone;
+    // fall back to the local timezone
+    const localTimezoneInfo = getTimeZoneInfo(getTimeZone(), Date.now());
+    return localTimezoneInfo?.ianaName;
+  }
+
   query(request: DataQueryRequest<CHQuery>): Observable<DataQueryResponse> {
+    const targets = request.targets
```
mshustov, line 198: I will add a test later if we move forward with the current implementation

mshustov:
moved to https://github.com/grafana/clickhouse-datasource/pull/314

PR #328, opened 2023-03-21
Update project to 9.4.3 using @grafana/create-plugin

slvrtrn:
* Address build deprecation warning and migrate using [@grafana/create-plugin](https://www.npmjs.com/package/@grafana/create-plugin#migrate-your-existing-plugin) tools
* Bump all main `@grafana` libraries from 8.5.11 to 9.4.3
* Add missing `tslib` dependency
* Address breaking changes in the tests - mainly handle async user events and add `waitFor` when necessary to prevent React's "overlapping act() calls" warnings
* Address changes in Jest's behavior in `validate.test.ts`
* Use Grafana 9.4.3 and ClickHouse 23.2 in docker-compose
* Fix a few typos


gingerwizard:
Spoke with @slvrtrn and completely understand why he wants to bump these but this will break support for 8.x users. 
We have a few options:

1. Create a 2.x branch and move main to 3.x. Moving forward we maintain both. Only 2.x supports 8.x. We backport where possible. We now incur a maintenance overhead until 10.x grafana is released.
2. We push this change until 10.x is released. @bossinc any timeline on this? this will block some improvements.
3. We accept that 8.x users will not be supported and release 3.x. 2.x is not maintained.

package.json
```diff
@@ -2,44 +2,76 @@
   "name": "clickhouse-datasource",
   "version": "2.1.1",
   "description": "Clickhouse Datasource",
+  "engines": {
+    "node": ">=16"
+  },
   "scripts": {
-    "spellcheck": "cspell -c cspell.config.json \"**/*.{ts,tsx,js,go,md,mdx,yml,yaml,json,scss,css}\"",
```
bossinc, line 6: We should keep the spellcheck script.

slvrtrn: And `watch` for a better developer experience as it seems to be missing from the new template. I will add them back :+1:

slvrtrn: @bossinc I've added `cspell` dev dependency and the `spellcheck` script.

bossinc:
@gingerwizard Grafana 10 will release in June.

We can create a v2 branch as you suggested, but it would only be worth doing if there are plans to use features from create-plugin before June. If not, we can park this PR until 10 is released.

[Here is a form post explaining the release support lifecycle](https://community.grafana.com/t/clarifications-on-grafana-release-support-lifecycle/77054/3)

gingerwizard:
@bossinc there are. We want to support Explore histograms. Please create a v2 branch.

bossinc (approved):

Issue #300, opened 2023-02-13
Let user to specify "format as" in query builder


jkaflik:
**What happened**:

Currently, the user is forced to follow the predefined algorithm that figures out the expected data format based on fields available in the query and its name. (please see Screenshots section)

This requires the user to remember the rules to follow described in README.md.

**What you expected to happen**:

The user should be able to intentionally select what is the data format expected by him. Please see the screenshot attached from a view in Altinity's CH data source plugin.

**Screenshots**

Grafana CH plugin:

```typescript
export const getFormat = (sql: string): Format => {
  // convention to format as time series
  // first field as "time" alias and requires at least 2 fields (time and metric)
  const selectList = getFields(sql);
  // if there are more than 2 fields, index 1 will be a ','
  if (selectList.length > 2 && isString(selectList[0])) {
    const firstProjection = selectList[0].trim().toLowerCase();
    if (firstProjection.endsWith('as time')) {
      return Format.TIMESERIES;
    }
    if (firstProjection.endsWith('as log_time')) {
      return Format.LOGS;
    }
  }
  return Format.TABLE;
};
```

Altinity's data source plugin:
![Screenshot 2023-02-13 at 16 48 21](https://user-images.githubusercontent.com/101614/218505157-4acb511f-776e-4360-a5a3-4dbe11868432.png)

asimpson:
This is a UX problem in my opinion. If the user selects "Time series" but the returned data doesn't contain the right fields for that format do we throw an error? Do we switch back to Table view? Both? The current reliance on user's having stored the README rules in their brains isn't ideal either. 

We could do something where we have the three formats as a dropdown in the header of the query builder and when you switch between them show some help text for each one? Ideally we work with UX on how this works.

jkaflik:
I suggest providing a dropdown in the header so that user can pick format in advance. Based on selection, there might be either a info box displayed with details on query requirements, or we can be more proactive validating whether query does have everything we need.  (not sure if doable as user can provide his query)
When data frame received from backend, fields has to be validated. 

Please let me know if you want to follow up on it with UX on your side.

asimpson:
@jkaflik I'll post back here once I have a chat with UX. The next step after that will be figuring out when this can get worked on.

asimpson:
Hopefully I'll be able to reconnect with UX next week. Sorry for the delay.

mshustov:
Other things we can do based on the query  selector:
* extend query builder depending on the query type
[...]

PR #329, opened 2023-03-21
Support format dropdown and support for rendering traces

gingerwizard:
This PR adds support for rendering traces in Explore and the traces panel. This is achieved through several changes:

1. Maps, Tuples, and Arrays are now converted to `json.RawMessage`. This allows tags to rendered correctly. I think this is an improvement and shouldn't impact previous approach - which returned them as a string.
2.  For traces to be rendered we need the preferred visualization type to be set - see [here](https://grafana.com/docs/grafana/latest/explore/trace-integration/#data-api) and [here](https://github.com/grafana/sqlds/blob/main/query.go#L153). We have no good way of detecting the user is querying traces so we introduce a drop down where the user can specify the Format of the response. The default is `Auto` which means the current behaviour - so no backward compatibility break. If the user specifies the format it overrides the value sent to the server.

Note we maintain full backwards compatibility.

image

image

Thus also closes https://github.com/grafana/clickhouse-datasource/issues/300


gingerwizard:
@bossinc my tests pass locally. Can you put the CI failures here please.

bossinc:
@gingerwizard 
``` + mage -v lint
Running target: Lint
exec: golangci-lint "run" "./..."
pkg/plugin/driver_test.go:219:16: Error return value of `json.Unmarshal` is not checked (errcheck)
	json.Unmarshal(bytes, &rawJSON)
	              ^
Error: running "golangci-lint run ./..." failed with exit code 1```

**Review by bossinc — APPROVED** [review]
looks great!

**Comment by mustafaakin** [comment]
Great development, I previously opened a discussion on Grafana but did not think opening an issue here. https://github.com/grafana/grafana/discussions/62924

I followed the instructions here to test it locally https://grafana.com/tutorials/build-a-data-source-plugin/ and but from the on the UI I see Plugin unavailable and same in logs and the additional this line: (I use M1 mac)

```
logger=plugin.loader t=2023-03-22T11:19:39.45338+03:00 level=error msg="Could not start plugin" pluginId=grafana-clickhouse-datasource err="fork/exec /Users/mustafa/projects
       │ /tmp/grafana-plugins/clickhouse-datasource/dist/gpx_clickhouse_darwin_arm64: no such file or directory"
```

I did not see any info on DEV_GUIDE.md about it. I might be missing a thing. 

My plugin/dist folder is as follows:

```
CHANGELOG.md          README.md             img                   module.js.LICENSE.txt plugin.json
LICENSE               dashboards            module.js             module.js.map
```

mshustov:
@mustafaakin maybe Grafana team will help you to run the plugin with Docker. In the meantime, you can try to run Grafana from sources following the guide https://github.com/grafana/grafana/blob/main/contribute/developer-guide.md
Don't forget to add to `defaults.ini` a path to the plugin folder https://github.com/grafana/grafana/blob/main/conf/defaults.ini#L27 and set `allow_loading_unsigned_plugins=clickhouse-datasource`

PR #343, opened 2023-03-31
Logs volume histogram support

### LINKED ISSUE #182: Support summary histogram in explore view + drill down
Currently, explore only renders the histogram based on the returned results. Other plugins (Loki, Elasticsearch) generate a histogram for the entire time range based on an aggregation/GROUP BY.
We should add this capability.

slvrtrn:
I cannot create a PR on top of an external branch (see #328), so this one contains both changesets.

To make it simpler for reviewing, here is an identical PR based on my fork where I could use #328 as a base branch: https://github.com/slvrtrn/clickhouse-datasource/pull/1/files

Screenshots:

g1

g2

For obvious reasons, I added two new fields if the format is set to LOGS and the builder mode is Table - `timeField` and `logLevelField`. The only thing I could not figure out is how to restrict showing these fields only when the Explore mode is selected.

I used core ElasticSearch implementation as a reference, and there is a fair amount of copy-paste from the base repository (mainly from [here](https://github.com/grafana/grafana/blob/release-9.4.3/public/app/core/logsModel.ts)).

If we could get

* `LogLevelColor`
* `getThemeColor` 
* `getLogVolumeFieldConfig`
* `getIntervalInfo`

as a part of the public API, the amount of copy-pasted code will be significantly reduced.

Of course, some tests are in order, but I wanted to collect some general feedback and maybe some insights considering the TODOs in the code.

It works the following way: 
* In Explore, if LOGS format is selected and the mode is Table (List), we see two new fields - `timeField` and `logLevelField`. If the `timeField` is set, the supplementary query for logs volumes will be generated and executed.
* Assuming that `timeField` is set to `created_at`, if `logLevelField` is set, the generated query looks like this (abbreviations are taken from `LogLevel` enum):
```sql
SELECT sum(toString("level") IN
           ('critical', 'fatal', 'crit', 'alert', 'emerg', 'CRITICAL', 'FATAL', 'CRIT', 'ALERT', 'EMERG', 'Critical',
            'Fatal', 'Crit', 'Alert', 'Emerg'))                                                           AS critical,
       sum(toString("level") IN ('error', 'err', 'eror', 'ERROR', 'ERR', 'EROR', 'Error', 'Err', 'Eror')) AS error,
       sum(toString("level") IN ('warn', 'warning', 'WARN', 'WARNING', 'Warn', 'Warning'))                AS warn,
       sum(toString("level") IN
           ('info', 'information', 'informational', 'INFO', 'INFORMATION', 'INFORMATIONAL', 'Info', 'Information',
            'Informational'))                                                                             AS info,
       sum(toString("level") IN ('debug', 'dbug', 'DEBUG', 'DBUG', 'Debug', 'Dbug'))                      AS debug,
       sum(toString("level") IN ('trace', 'TRACE', 'Trace'))                                              AS trace,
       sum(toString("level") IN ('unknown', 'UNKNOWN', 'Unknown'))                                        AS unknown,
       toStartOfInterval("created_at", INTERVAL 1 DAY)                                                    AS time
FROM default."logs"
WHERE (created_at >= $__fromTime AND created_at <= $__toTime)
GROUP BY toStartOfInterval("created_at", INTERVAL 1 DAY) AS time
ORDER BY time ASC;
```
NB: here I am avoiding `lower` function usage, so reasonable values for `IN` clauses are generated in advance.

* If `logLevelField` is NOT set, the generated query looks like this:
```sql
SELECT toStartOfInterval("created_at", INTERVAL 1 MINUTE) AS time, count(*) logs
FROM default."logs"
WHERE (created_at >= $__fromTime AND created_at <= $__toTime)
GROUP BY toStartOfInterval("created_at", INTERVAL 1 MINUTE) AS time
ORDER BY time ASC;
```
* In the end, it gives us a single data frame with `time` plus all the log-level fields, which are transformed into a data frame per level.


src/components/queryBuilder/QueryBuilder.tsx
```diff
@@ -197,8 +206,27 @@ export const QueryBuilder = (props: QueryBuilderProps) => {
           timeFieldType={builder.timeFieldType}
           onTimeFieldChange={onTimeFieldChange}
           fieldsList={fieldsList}
+          timeFieldTypeCheckFn={isDateType}
+          labelAndTooltip={selectors.components.QueryEditor.QueryBuilder.TIME_FIELD}
         />
       )}
+      {
+        // Time and LogLevel fields selection for Logs Volume histogram in the Explore mode
+        // TODO: show only in the Explore mode
```
slvrtrn, line 215: Do we have this information available within the plugin?

src/data/logs.ts
```diff
@@ -0,0 +1,264 @@
...
+  scopedVars: ScopedVars,
+  timespanMs: number,
+  timeField = 'created_at'
+): string {
+  let interval = 'DAY';
+  if (scopedVars.__interval) {
+    let intervalMs: number = scopedVars.__interval_ms.value;
+    // below 5 seconds we force the resolution to be per 1ms as interval in scopedVars is not less than 10ms
+    if (timespanMs < SECOND * 5) {
+      // TODO: workaround
```
slvrtrn, line 248: Is it necessary to support millisecond precision, or can we get away with up to second precision?

src/components/queryBuilder/LogLevelField.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { FullField } from 'types';
+
+interface LogLevelEditorProps {
+  fieldsList: FullField[];
+  onLogLevelFieldChange: (logLevelField: string, logLevelFieldType: string) => void;
+}
+
+export const LogLevelFieldEditor = (props: LogLevelEditorProps) => {
+  const { label, tooltip } = selectors.components.QueryEditor.QueryBuilder.LOG_LEVEL_FIELD;
+  // TODO: filter by strings, enums?
```
slvrtrn, line 14: I guess here we might see
* String
* FixedString
* Enum
* LowCardinality
* Nullable of all above

src/components/queryBuilder/TimeField.tsx
```diff
@@ -3,19 +3,20 @@ import { SelectableValue } from '@grafana/data';
...
 import { FullField } from 'types';
-import { isDateType } from './utils';
 
 interface TimeFieldEditorProps {
   fieldsList: FullField[];
   timeField: string;
   timeFieldType: string;
   onTimeFieldChange: (timeField: string, timeFieldType: string) => void;
+  timeFieldTypeCheckFn: (type: string) => boolean;
+  labelAndTooltip: typeof selectors.components.QueryEditor.QueryBuilder.TIME_FIELD;
```
slvrtrn, line 13: I made it a bit more configurable to reuse for Logs Volume time field selection, as we want to see only the `DateTime*` types there and specify different labels and tooltips.

src/data/CHDatasource.ts
```diff
@@ -32,6 +57,129 @@ export class Datasource extends DataSourceWithBackend<CHQuery, CHConfig> {
...
+
+        if (intervalInfo.intervalMs !== undefined) {
+          logsVolumeRequest.intervalMs = intervalInfo.intervalMs;
+          logsVolumeRequest.scopedVars.__interval_ms = {
+            value: intervalInfo.intervalMs,
+            text: intervalInfo.intervalMs,
+          };
+        }
+
+        logsVolumeRequest.hideFromInspector = true;
```
slvrtrn, line 85: I moved this out of `queryLogsVolume` here, as I thought it would be a bit cleaner to mutate the request only in one place

src/data/logs.ts
```diff
@@ -0,0 +1,264 @@
...
+    };
+  });
+}
+
+/**
+ * Take multiple data frames, sum up values and group by level.
+ * Return a list of data frames, each representing single level.
+ */
+function aggregateRawLogsVolume(rawLogsVolume: DataFrame[]): DataFrame[] {
+  if (rawLogsVolume.length !== 1) {
```
slvrtrn, line 125: just to clarify a bit: we already getting a single DataFrame here which looks something like this:

```json
{
    "fields": [
        {
            "name": "time",
            "type": "time",
            "typeInfo": {
                "frame": "time.Time"
            },
            "config": {},
            "values": [
                1680140003000,
                1680140004000,
                1680140027000
            ],
            "entities": {}
        },
        {
            "name": "emerg",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "fatal",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "alert",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "crit",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "critical",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "warn",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "warning",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "err",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "eror",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "error",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "info",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                1,
                1
            ],
            "entities": {}
        },
        {
            "name": "information",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "informational",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "notice",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "dbug",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "debug",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                1,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "trace",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        },
        {
            "name": "unknown",
            "type": "number",
            "typeInfo": {
                "frame": "uint64"
            },
            "config": {},
            "values": [
                0,
                0,
                0
            ],
            "entities": {}
        }
    ],
    "length": 3
}
```

gingerwizard:
Firstly, this rocks. 👏 

Questions:

1. How will this work in plain SQL view? - this is how i expected most users will interact with Logs. 
2. We added the format to address some challenges with traces and logs. The `Format` conflicts somewhat with the `Show as` but _i think_ they are complementary i.e. they let someone build an aggregation but render as logs. `Show As` should probably thus be `Build as`. If this work is going into a 3x release (due to SDK update), we should take the opportunity to clean this all up. 
3. If format is to `Logs`, there is backend code that auto-detects the time field - the first one. Confusingly though also, the front end sets the format if the `log_time` column is present (it does this for timeseries as well) - the explicit format overrides this detection. I think the intention of this was for the SQL view. If we add the explicit options to set the log time column we should remove any auto-detection IMO in the frontend. 
4. The elasticsearch plugin requires the time field to be set at the datasource for example. Did we consider this? it might help with (1). Users could also set the log level here.

bossinc:
Drone is failing on `yarn spellcheck` Please add these words to cspell.config.json

```
 67/82 ./src/data/CHDatasource.ts 57.35ms X
/drone/src/src/data/CHDatasource.ts:71:15 - Unknown word (timespan)
/drone/src/src/data/CHDatasource.ts:72:76 - Unknown word (timespan)
/drone/src/src/data/CHDatasource.ts:89:102 - Unknown word (timespan)
/drone/src/src/data/CHDatasource.ts:119:5 - Unknown word (timespan)
/drone/src/src/data/CHDatasource.ts:134:7 - Unknown word (timespan)

68/82 ./src/data/logs.ts 47.05ms X
/drone/src/src/data/logs.ts:181:57 - Unknown word (timespan)
/drone/src/src/data/logs.ts:186:9 - Unknown word (timespan)
/drone/src/src/data/logs.ts:211:3 - Unknown word (timespan)
/drone/src/src/data/logs.ts:218:9 - Unknown word (timespan)
/drone/src/src/data/logs.ts:241:53 - Unknown word (emerg)
/drone/src/src/data/logs.ts:242:28 - Unknown word (eror)
/drone/src/src/data/logs.ts:245:21 - Unknown word (dbug)
```

mshustov:
Adding a filter when clicking on the amplifier on Logs view doesn't work. @slvrtrn is it something we're going to add in a follow-up?


> In Explore, if LOGS format is selected and the mode is Table (List), we see two new fields - timeField and logLevelField.

Note that `Log level field` and `Time field` aren't reset when changing a table name in `Query builder`

> How will this work in plain SQL view? - this is how i expected most users will interact with Logs.

@gingerwizard  We can add `Time field` and `Log level field` input elements. But I'm not opposed to (4) aligning with elaticsearch plugin behavior, which might be already well-known among users.

> The bucket time width should be dynamic based on the interval - see how the macro $__timeInterval works.

@gingerwizard  Could you elaborate? Don't `getIntervalInfo` & `getTimeFieldRoundingClause` implement bucketing? 

> If we could get
> LogLevelColor
> getThemeColor
> getLogVolumeFieldConfig
> getIntervalInfo

@bossinc @asimpson Would Grafana team accept a PR exporting these from https://github.com/grafana/grafana/blob/ab53772416659fa845effe4913051e1f829d47e2/public/app/core/logsModel.ts? 
This logic seems to be fundamental for any plugin implementing Logging support

src/data/CHDatasource.ts
```diff
@@ -32,11 +32,10 @@ import { cloneDeep, isEmpty, isString } from 'lodash';
 import {
   DEFAULT_LOGS_ALIAS,
   getIntervalInfo,
-  getTimeFieldRoundingClause,
-  LOG_LEVELS,
+  getTimeFieldRoundingClause, LOG_LEVEL_TO_IN_CLAUSE,
```
bossinc, line 35:
```suggestion
  getTimeFieldRoundingClause,
  LOG_LEVEL_TO_IN_CLAUSE,
```

src/data/CHDatasource.ts
```diff
@@ -32,11 +32,10 @@ import { cloneDeep, isEmpty, isString } from 'lodash';
...
   queryLogsVolume,
   TIME_FIELD_ALIAS,
-} from './logs';
+} from './logs'
```
bossinc, line 38:
```suggestion
} from './logs';
```

asimpson:
Chiming in here re:

* `LogLevelColor`
* `getThemeColor` 
* `getLogVolumeFieldConfig`
* `getIntervalInfo`

Let's move forward as is and if/when those pieces get exported we can refactor.

src/data/logs.ts
slvrtrn: I am excluding this for now.

src/components/queryBuilder/LogLevelField.tsx
slvrtrn: Will be done as a follow-up

src/data/CHDatasource.ts
```diff
@@ -32,6 +56,132 @@ export class Datasource extends DataSourceWithBackend<CHQuery, CHConfig> {
...
+
+  private getSupplementaryLogsVolumeQuery(
+    logsVolumeRequest: DataQueryRequest<CHQuery>,
+    query: CHQuery,
+    timespanMs: number
+  ): CHQuery | undefined {
+    if (
+      query.format !== Format.LOGS ||
+      query.queryType !== QueryType.Builder ||
+      query.builderOptions.mode !== BuilderMode.List ||
```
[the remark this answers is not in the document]
slvrtrn: It will be resolved as a part of the re-design.

src/components/queryBuilder/QueryBuilder.tsx
slvrtrn: Found it. `CoreApp` is available from `CHQueryEditorProps` in `CHQueryEditor` and could be passed to `QueryBuilder`

slvrtrn:
I've recreated this PR as https://github.com/grafana/clickhouse-datasource/pull/352
Let's continue the review there.

PR #350, opened 2023-04-04
Add initial feature tracking

### LINKED ISSUE #339: Add initial support for feature tracking
It would be useful to understand what features of the datasource are being used and how they are being used. We can start by tracking some basic features to get a sense of their usage. Below are two basic things to track, we can then expand it out as we see the need for more.

- Track which mode is used, e.g. "SQL Editor" or "Query builder"
- Track how results are viewed in Query Builder: "Table", "Aggregate", or "Time Series"

alyssabull:
**What happened**:

The goal of this task is to add feature tracking to CH so we can get more information on how much certain features are used. This PR adds initial tracking for the type of queries (SQL vs Builder) and if a builder query is used what format is selected (Table, Aggregate or Time Series). With this information we can determine what how much certain features are used. If they are not, we can remove them from the code base, if they are we can improve its UX so they are easier to use.

**This is the list of things we can track so far**:

**Query editor**:
Format: I assume a big percentage of queries will use the "Table" format, since it is the default. We will track how many builder queries use "Table", "Aggregate" or "Time Series".
Builder vs SQL editor: How much is the raw (SQL) editor used? 

Closes #339


aangelisc (dismissed):
Great work @alyssabull, thank you for adding this!

asimpson (approved):
Lovely tests, nice work! :clap:

aangelisc (approved):

PR #363, opened 2023-04-19
Histogram UI fixes

slvrtrn:
- [x] Support filter/filter out logs view actions
- [x] Fix undefined database name by default 
- [x] Reset level and time field properly on table/database change
- [x] Make it possible to clear the level field (so the histogram will render without grouping by level)
- [x] Propagate UI changes from `modifyQuery` to `QueryBuilder` component
- [x] Fix filter value that gets stuck in the UI


slvrtrn:
The last bullet refers to a bug that is reproducible, for example:
- we filter out level by some values, let's say `level != debug AND level != info`
- we filter level by some new value, now it should be `level = error`
- in the UI, either `debug` or `info` will be stuck like `level = debug`, while in the actual QueryBuilder object, everything is correct. The culprit is likely [here](https://github.com/ClickHouse/clickhouse-grafana-datasource/blob/5f25fb466ad046ffe8796eca422590d74ddc0839/src/components/queryBuilder/Filters.tsx#L67), as the `value` in the state becomes stale, trying to find a workaround atm.

src/components/queryBuilder/LogLevelField.tsx
```diff
@@ -7,19 +7,14 @@ import { FullField } from 'types';
 interface LogLevelEditorProps {
   fieldsList: FullField[];
   onLogLevelFieldChange: (logLevelField: string) => void;
+  logLevelField: string | null;
```
bossinc, line 10:
```suggestion
  logLevelField: string;
```
I believe strings are already nullable. Does this work with just string?

bossinc:
@slvrtrn Are you referring to the magnifier between the time selector and run query? 
image

I'm not able to reproduce this. What data set are you using and what queries?

mshustov:
@bossinc I believe Serge referred to the magnifier in Logs panel in Explore

slvrtrn:
@bossinc, as @mshustov correctly suggested, it was about the magnifiers (add a filter or filter out) in the logs panel in Explore.

src/components/queryBuilder/LogLevelField.tsx
slvrtrn: No, null is not assignable to type string, that's why it's explicit here.

mshustov:
> The culprit is likely [here](https://github.com/ClickHouse/clickhouse-grafana-datasource/blob/5f25fb466ad046ffe8796eca422590d74ddc0839/src/components/queryBuilder/Filters.tsx#L67), as the value in the state becomes stale, trying to find a workaround atm.

@slvrtrn https://github.com/grafana/clickhouse-datasource/pull/363/commits/3132f4ffab02ecc8cbad875c50174e65ae8ead2e is supposed to fix the problem, PTAL

slvrtrn:
@mshustov, thanks for the fix. It is working as expected!

@bossinc, I think this one is ready for the final review and merging.

bossinc (commented):
Looks good! It just needs to be updated with main

slvrtrn:
@bossinc, updated

bossinc (approved):

bossinc:
@slvrtrn I'll merge this next!

mshustov:
@bossinc can we merge now?

slvrtrn:
🎉

PR #366, opened 2023-04-28
Allow passing custom ClickHouse settings in datasource

aramperes:
Adds a "[Custom Settings](https://clickhouse.com/docs/en/operations/settings)" section to the ClickHouse datasource config editor.

image

This is my first time diving into a Grafana plugin, so all feedback is appreciated 🍻

### DIFF
--- a/pkg/plugin/driver.go
+++ b/pkg/plugin/driver.go
@@ -7,6 +7,11 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
 	"github.com/ClickHouse/clickhouse-go/v2"
 	"github.com/grafana/clickhouse-datasource/pkg/converters"
 	"github.com/grafana/clickhouse-datasource/pkg/macros"
@@ -17,10 +22,6 @@ import (
 	"github.com/grafana/grafana-plugin-sdk-go/data/sqlutil"
 	"github.com/grafana/sqlds/v2"
 	"github.com/pkg/errors"
-	"os"
-	"strconv"
-	"strings"
-	"time"
 )
 
 // Clickhouse defines how to connect to a Clickhouse datasource
@@ -129,6 +130,13 @@ func (h *Clickhouse) Connect(config backend.DataSourceInstanceSettings, message 
 	if protocol == clickhouse.HTTP {
 		compression = clickhouse.CompressionGZIP
 	}
+	customSettings := make(clickhouse.Settings)
+	if settings.CustomSettings != nil {
+		for _, setting := range settings.CustomSettings {
+			customSettings[setting.Setting] = setting.Value
+		}
+	}
+
 	db := clickhouse.OpenDB(&clickhouse.Options{
 		ClientInfo: clickhouse.ClientInfo{
 			Products: getClientInfoProducts(),
@@ -146,6 +154,7 @@ func (h *Clickhouse) Connect(config backend.DataSourceInstanceSettings, message 
 		DialTimeout: time.Duration(t) * time.Second,
 		ReadTimeout: time.Duration(qt) * time.Second,
 		Protocol:    protocol,
+		Settings:    customSettings,
 	})
 
 	timeout := time.Duration(t)
--- a/pkg/plugin/driver.go
+++ b/pkg/plugin/driver.go
@@ -130,6 +130,12 @@ func (h *Clickhouse) Connect(config backend.DataSourceInstanceSettings, message 
 	if protocol == clickhouse.HTTP {
 		compression = clickhouse.CompressionGZIP
 	}
+	customSettings := make(clickhouse.Settings)
+	if settings.CustomSettings != nil {
+		for _, setting := range settings.CustomSettings {
+			customSettings[setting.Setting] = setting.Value
+		}
+	}
 
 	db := clickhouse.OpenDB(&clickhouse.Options{
 		ClientInfo: clickhouse.ClientInfo{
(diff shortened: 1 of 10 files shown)


aramperes:
One thing I'm looking into doing is allowing variables to be passed as setting values, such that dashboards can tweak settings on the fly.

For instance, passing the Grafana user's login as a setting whenever a query is executed:

```
custom_setting_username: ${user.login}
```

Can someone let me know if that sounds feasible? It looks like the ClickHouse connection is abstracted away as a `sqlstore.DBSession`, which means that settings are only applied when the session is re-created, and context/dashboard variables are not passed to the connection factory.

src/data/CHDatasource.ts
```diff
@@ -218,6 +218,11 @@ export class Datasource
     return {
       ...query,
       rawSql: this.replace(rawQuery, scoped) || '',
+      customSettings: this.settings.jsonData.customSettings?.map(({ setting, value }) => {
+        return {
+          setting, value: this.replace(this.applyConditionalAll(value, getTemplateSrv().getVariables()), scoped) || ''
```
bossinc, line 223: I am not sure if we should evaluate custom template variables. It would be a confusing experience to have your custom settings work for a dashboard and error in Explore view, for example.

src/views/CHConfigEditor.tsx
```diff
@@ -315,6 +327,60 @@ export const ConfigEditor: React.FC<Props> = (props) => {
...
+        {
+          customSettings.map(({ setting, value }, i) => {
+            return <InlineFieldRow key={i}>
+              <InlineField
+                label={`Custom Setting (${i})`}
+                aria-label={`Custom Setting (${i})`}
+              >
+                <Input
+                  value={setting}
+                  placeholder={'my_setting'}
```
bossinc, line 342:
```suggestion
                  placeholder={'Setting'}
```

src/views/CHConfigEditor.tsx
```diff
@@ -315,6 +327,60 @@ export const ConfigEditor: React.FC<Props> = (props) => {
...
+                  onBlur={() => {
+                    onCustomSettingsChange(customSettings);
+                  }}
+                >
+                </Input>
+              </InlineField>
+              <InlineField>
+                <Input
+                  value={value}
+                  placeholder={'value'}
```
bossinc, line 357:
```suggestion
                  placeholder={'Value'}
```

src/views/CHConfigEditor.tsx
```diff
@@ -315,6 +327,60 @@ export const ConfigEditor: React.FC<Props> = (props) => {
...
+      <div className="gf-form-group">
+        <h3>Custom Settings</h3>
+        <br />
+        {
+          customSettings.map(({ setting, value }, i) => {
+            return <InlineFieldRow key={i}>
+              <InlineField
+                label={`Custom Setting (${i})`}
+                aria-label={`Custom Setting (${i})`}
+              >
```
bossinc, line 339: We should have a label for `Setting` and `Value`. 
https://developers.grafana.com/ui/latest/index.html?path=/story/forms-inlinefield--multiple

bossinc (changes requested):
Great work! :+1: 

I just have a few nits. 
I would like thoughts from the team on using template variables at a data source settings level.

src/data/CHDatasource.ts
aramperes: Agreed that it's not a perfect situation, however there is a use-case here for passing user information as a parameter (ROW POLICY often used custom settings). Maybe we could pass this some other way? Evaluating the templates on the server side for a subset of variables?

pkg/plugin/driver.go
```diff
@@ -227,8 +230,15 @@ func (h *Clickhouse) MutateQuery(ctx context.Context, req backend.DataQuery) (co
...
+	customSettings := make(clickhouse.Settings)
+	if dataQuery.CustomSettings != nil {
+		for _, setting := range dataQuery.CustomSettings {
+			customSettings[setting.Setting] = setting.Value
+		}
+	}
+
 	loc, _ := time.LoadLocation(dataQuery.Meta.TimeZone)
-	return clickhouse.Context(ctx, clickhouse.WithUserLocation(loc)), req
+	return clickhouse.Context(ctx, clickhouse.WithUserLocation(loc), clickhouse.WithSettings(customSettings)), req
```
aramperes, line 241: @bossinc In this function `MutateQuery`, is there a way to access:

- Datasource settings (`backend.DataSourceInstanceSettings`)?
- Currently logged-in user to Grafana?

If so, I could remove client-side passing/templating of settings and do this entirely in the backend.

src/data/CHDatasource.ts
aramperes: Ok, for now I decided to simplify this PR and remove this logic. Templated settings can be passed in individual queries like this for now:

```
SELECT * FROM LS_DEMO
SETTINGS ls_profile='${__user.login}';
```

This is just a minor inconvenience for now

aramperes: I now have a fork of `sqlds` and `clickhouse-datasource` that interpolates server-side macros in custom settings, and adds a new macro for `$__userLogin`.

- https://github.com/aramperes/sqlds/compare/main...aramperes:sqlds:user-login-macro
- https://github.com/aramperes/grafana-clickhouse-datasource/compare/custom-clickhouse-settings...aramperes:grafana-clickhouse-datasource:user-macro

bossinc (approved):
Thank you! This is great. Default template variables can be in another PR

jkaflik:
@bossinc, as it's already reported by @aramperes , this feature will not work for native protocol. ClickHouse will reject those settings as unknown.

Please see https://github.com/ClickHouse/clickhouse-go/issues/983

aramperes:
> @bossinc, as it's already reported by @aramperes , this feature will not work for native protocol. ClickHouse will reject those settings as unknown.
> 
> Please see https://github.com/ClickHouse/clickhouse-go/issues/983

It works for ClickHouse client settings that the user can customize normally through clickhouse-client, just not ClickHouse's concept of "custom settings" with custom prefixes.

PR #388, opened 2023-05-15
Fix version bug and add sql format feature tracking

alyssabull:
This PR fixes a bug that was not returning the correct clickhouse version in the initial feature tracking ([PR here](https://github.com/grafana/clickhouse-datasource/pull/350)). It also adds feature tracking to record the format being used in a SQL query. The format options are Auto (default), Table, Logs, Time series and Trace.


aangelisc (approved):
👌

PR #389, opened 2023-05-16
Feature: Add support for the secure socks proxy

stephaniehingtgen:
**Background**
This PR enables support for [the secure socks proxy](https://grafana.com/docs/grafana/next/setup-grafana/configure-grafana/proxy/), which is a new feature in Grafana 10. The feature allows datasource connections to be proxied through a socks5 proxy with TLS. This enables Grafana users to connect to datasources that live in different networks than where Grafana is running when they cannot open up ports to the public internet.

**Implementation notes**
- The feature itself no longer sits behind a feature toggle in Grafana OSS, but a feature toggle is being used in all non-core datasources until the datasource can upgrade the frontend dependency`@grafana/runtime`to at least 10.0.0. At that point, we can switch [this line](https://github.com/grafana/clickhouse-datasource/compare/support-secure-socks-proxy?expand=1#diff-50b041c21a06bf71ca3a091bad78078cae1a4e988cc31a460b28d7c83b2e3955R331) to look at `config.secureSocksDSProxyEnabled`, rather than the feature toggle.
- If in the future, clickhouse is moved to using the sdk to parse the http client settings, the code around parsing the secure socks proxy settings can be removed. Originally, I had tried to just parse the secure socks proxy settings [like this](https://github.com/grafana/clickhouse-datasource/pull/389/commits/682bd64f12e8d38d04f4f7c827b0e8f97a25707f), but that would cause a panic in the tests because `tlsAuth` is passed in as a string 
- Dependency updates:
  - `clickhouse-go`: need to include [this PR](https://github.com/ClickHouse/clickhouse-go/pull/998) to get the http based connection to work. This new version will be released tomorrow, so this PR is in draft until then. 
  - `grafana-plugin-sdk-go`: Needs to run at least [v0.159.0](https://github.com/grafana/grafana-plugin-sdk-go/releases/tag/v0.159.0), where the setup functionality was added 

**Screenshots**
ClickhouseHTTP
ClickhouseNative

Closes https://github.com/grafana/hosted-grafana/issues/3945


bossinc (approved):

bossinc:
Nice! thank you for doing this work

asimpson (approved):
:pray: thank you!

PR #451, opened 2023-07-11
do not create malformed adhoc filters

mshustov:
Sometimes our o11y platform creates malformed links, instead of querying the datasource with an invalid query, we can validate it in place


src/data/adHocFilter.ts
```diff
@@ -26,23 +26,33 @@ export class AdHocFilter {
...
+    if(filters === '') {
+      return sql;
+    }
     // Semicolons are not required and cause problems when building the SQL
     sql = sql.replace(';', '');
     return `${sql} settings additional_table_filters={'${this._targetTable}' : '${filters}'}`;
   }
 }
 
+function validate(filter: AdHocVariableFilter): boolean {
```
mshustov, line 49: Should we log in case of invalid filters?

src/data/adHocFilter.ts
```diff
@@ -26,23 +26,33 @@ export class AdHocFilter {
...
+  return filter.key !== undefined && filter.operator !== undefined && filter.value !== undefined;
```
mshustov, line 50: I didn't add a test for `filter.key: undefined` since here we assume it's a string https://github.com/grafana/clickhouse-datasource/pull/451/files#diff-432fcb747320944461075f7baa45beacdd3b15b65cf530690f35030d978c5e3dR23
Should I add at the very top this condition?
```js
if(filter.key === undefined) return sql
```

asimpson: I like this idea. We should probably have had that to begin with given the `includes` line you linked to.

src/data/adHocFilter.ts
asimpson: A log would be nice just for knowing how many filters aren't valid and potentially debugging issues.
Would notification of an invalid filter be helpful to the user? I'm not sure it would be but if it is we should figure out how to surface that.

asimpson (dismissed):
Looks good to me :clap:. Can you rebase and I'll merge once checks are green? :pray:

src/data/adHocFilter.ts
mshustov: > Would notification of an invalid filter be helpful to the user?

What is the preferred way to create a notification in Grafana? Let's start with adding logs, and we can add another channel later.

asimpson: We could do something like [this](https://developers.grafana.com/ui/latest/index.html?path=/story/overlays-alert-toast--examples). But I would only want to do that if we can give the user a clear action to take based on that notification.

asimpson (approved):

PR #457, opened 2023-07-18
Bump Go version used in CI to 1.20

adamyeats:
This PR fixes an issue surfaced in #455, whereby the CI pipeline was failing due to using a mixture of Go 1.18/1.19. Dependency `grafana-plugin-sdk-go` now requires 1.20 due to the usage of `unsafe.StringData`.

This also fixes formatting inconsistencies, as well as removing the `stable: false` option, which does nothing.

### DIFF
--- a/.github/workflows/run-backend-tests.yml
+++ b/.github/workflows/run-backend-tests.yml
@@ -10,7 +10,7 @@ on:
       - v1
       - main
   schedule:
-    - cron: "0 9 1 * *"
+    - cron: '0 9 1 * *'
 
 jobs:
   run:
@@ -19,7 +19,7 @@ jobs:
       fail-fast: true
       matrix:
         go:
-          - 1.19
+          - '1.20.x'
         clickhouse:
           - 22.3
           - 22.8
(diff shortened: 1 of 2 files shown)


aangelisc (dismissed):

aangelisc (approved):

.github/workflows/run-backend-tests.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       fail-fast: true
       matrix:
         go:
-          - 1.19
+          - '1.20.x'
```
aangelisc, line 22: Do we need the `.x` or can we leave this as just `1.20`?

adamyeats: @aangelisc AFAICT they do do different things, `1.20` locks to the minor version, but `1.20.x` gets the latest patch version. Would we prefer that or do we just want to get the latest patch vesion?

aangelisc: My gut tells me we should lock to the version Grafana uses but I'm open to suggestions

adamyeats: According to the `go.mod` in Grafana, the version is `1.20` but in the case of this file, it would get the latest patch version. In the GitHub Actions config, they are using `1.20.6` or `>=1.20` (which is a wider range than `1.20.x`).

aangelisc: Okay lets do it😊

PR #459, opened 2023-07-19
Fix conditionAll handling

### LINKED ISSUE #445: $__conditionalAll() macro works with the $var syntax but not ${var}
**What happened**:

When using a template variable checking the "All" option and using the `$__conditionalAll(condition, ${templateVar})` macro, the output is not `1=1` when selecting "All". 

I expected the output of the macro to be `1=1` when using curly brackets around the variable name. Or maybe update the documentation to be more explicit on the fact that the [curly brackets syntax](https://grafana.com/docs/grafana/latest/dashboards/variables/variable-syntax/) is not supported with the `$__conditionalAll()` macro.

1. Create a custom template variable `testvar` (values: `1,10,20`) with the option to include all options checked 
2. Create a dashboard with the following query `SELECT 1 WHERE $__conditionalAll( 'condition != All', ${testvar} )` (note the curly brackets)
3. Select the "All" value for the `testvar`
4. Open the `Query inspector` and look at the `request.data.queries[0].rawSql` value, you should see `"SELECT 1 WHERE 'condition != All'"`
5. Update the query to `SELECT 1 WHERE $__conditionalAll( 'condition != All', $testvar )` (no curly brackets around the variable name)
6. Check that the `request.data.queries[0].rawSql` value is `"SELECT 1 WHERE 1=1"`

I think the culprit lies in https://github.com/grafana/clickhouse-datasource/blob/0a013e24406cc3b630f3c3b75440dd7fe0388362/src/data/CHDatasource.ts#L240 where only the first character is removed (the `$` symbol). Maybe [`getVariableName()`](https://github.com/grafana/grafana/blob/30fc075cd7cecb2a7733a74c559b5ecff22fcbd9/public/app/features/templating/template_srv.ts#L156) could be used, but it looks like [it's not exposed](https://github.com/grafana/grafana/pull/65259). Otherwise maybe the documentation could be more explicit on the accepted syntax.

**Environment**:

- Grafana version: 9.4.7
- Plugin version: 3.1.0

aangelisc:
- Ensure any template variable is matched
- Update tests

Fixes #445

### DIFF
--- a/src/data/CHDatasource.test.ts
+++ b/src/data/CHDatasource.test.ts
@@ -7,6 +7,7 @@ import {
   ScopedVars,
   SupplementaryQueryType,
   TimeRange,
+  TypedVariableModel,
   toDataFrame,
 } from '@grafana/data';
 import { DataQuery } from '@grafana/schema';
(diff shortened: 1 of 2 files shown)


asimpson (approved):
:rocket: let's go! Nicely done.

PR #460, opened 2023-07-19
Ensure databases, tables, and columns are escaped correctly 

### LINKED ISSUE #450: Doesn't show table from database with "-" symbol in name
**What happened**:
The datasource doesn't fetch database data if it contains a special symbol in the name (like `-`).

The datasource allows working with database having special symbols in the name. We might need to quote the identifier, [ example](https://github.com/ClickHouse/clickhouse-connect/blob/aa0eeb9415929faabd663d0920f51868cdc588b8/clickhouse_connect/driver/query.py#L336-L341)

1. create a database with `-` in name.
2. create a table in the database 
3. Select the database query builder
4. plugin fails to fetch tables

workaround: escape the database name with `.

**Screenshots**

**Environment**:

- Grafana version: v10
- Plugin version: v3.1.0
- OS Grafana is installed on: running locally from sources

cc @rickysaltzer

aangelisc:
Databases, tables, or columns containing special characters must be escaped in order to be used correctly.

- Update tests

Fixes #450

### DIFF
--- a/src/components/queryBuilder/utils.spec.ts
+++ b/src/components/queryBuilder/utils.spec.ts
@@ -117,41 +117,41 @@ describe('isNumberType', () => {
 });
 
 describe('Utils: getSQLFromQueryOptions and getQueryOptionsFromSql', () => {
-  testCondition('handles a table without a database', 'SELECT name FROM "foo"', {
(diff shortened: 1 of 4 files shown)


asimpson (approved):
Nicely done!

PR #469, opened 2023-07-31
Fix auto formatting by reverting to table correctly

### LINKED ISSUE #467: When pulling 100k records the browser will use excessive amounts of memory and become very laggy
Steps to reproduce:

1. Create datasource that connects to Clickhouse Cloud.
2. Head to Explore
3. Run this query: `SELECT * FROM system.query_log LIMIT 100000`
4. Note the response is hundreds of megabytes and the browser becomes laggy

### Discussed in https://github.com/grafana/clickhouse-datasource/discussions/449

<div type='discussions-op-text'>

<sup>Originally posted by **zey1996** July 11, 2023</sup>
When I was querying a lot of logs. The performance of the grafana was not very good.
Perhaps we can add a paging feature to make it easier to query large amounts of logs or data?</div>

asimpson:
The browser would crash if auto format was selected and a large query was run. We would incorrectly coerce the resulting data into a time-series and graph it even if it wasn't actually a timeseries query. 

This also fixes differences in execution between the Grafana run query button and the clickhouse run query button.

Fixes #467

### DIFF
--- a/src/components/editor.ts
+++ b/src/components/editor.ts
@@ -8,7 +8,7 @@ export const getFormat = (sql: string, selectedFormat: Format): Format => {
     // first field as "time" alias and requires at least 2 fields (time and metric)
     const selectList = getFields(sql);
     // if there are more than 2 fields, index 1 will be a ','
-    if (selectList.length > 2 && isString(selectList[0])) {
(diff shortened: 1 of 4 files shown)


bossinc (approved):
Nice!

PR #486, opened 2023-08-14
Add a converter for the `Point` geo type

### LINKED ISSUE #189: Support Points
ClickHouse points aren't probably returned e.g. `sql: Scan error on column index 0, name "location": unsupported Scan, storing driver.Value type orb.Point into type *string: Could not process SQL results`

We should deliver these in a format which is compatible with geomap (array of floats?) - which works if you do something like 

```
SELECT
    location.1 AS longtitude,
    location.2 AS latitude
FROM noaa
LIMIT 10000
```

where location is a Point.

adamyeats:
Fixes #189.

This PR adds support for the `Point` geo data type. It also cleans up some of the floating `var`s we have for declaring regexes into a `map`, as well as alphabetises some lines.


aangelisc (dismissed):
This LGTM, we could extend the converter to actually validate the schema of a point type which would be a nice check for the users (and maybe give them better errors) but I'll leave that up to you 😊

bossinc (dismissed):
Nice!

adamyeats:
Thanks @aangelisc, added a check and updated tests 🎉

alyssabull (approved):
LGTM! Nice addition of the test 🎉

PR #487, opened 2023-08-15
Fix Table summary and Parts over time panels in Data Analysis dashboard

ethack:
The "Table summary" table was using the compressed bytes field and labeling it as uncompressed. This is now using the uncompressed bytes field.

The "Parts over time..." panel was creating an alias for the `table` column in the SELECT. But with ClickHouse this hides the original `table` column in the WHERE. (See https://github.com/ClickHouse/ClickHouse/issues/23194 for gory details.) This gives an alias to the original table and uses fully qualified column in the WHERE to fix.

Grafana complained about long time series needing to be sorted ASC. This changes the sort so Grafana displays the graph.

Before:
Note that the "Size on disk" and "Uncompressed size" columns are the same, which is not correct.
image

After:
image


aangelisc (approved):
LGTM, ty for the contribution!

PR #497, opened 2023-08-22
`ConfigEditor` updates

aangelisc:
For standardisation across Grafana we're updating datasource config editors to be more consistent. This is a part of https://github.com/grafana/oss-plugin-partnerships/issues/225.

**This is a breaking change as the plugin now has a minimum Grafana dependency of 9.5.0**

The config editor has been updated to replace legacy form component, replace InlineField with Field, and the config editor has been refactored to use the new config editor structural components. There is also now individual validation on required fields.

Node version has also been upgraded to 18.

image

Fixes grafana/oss-plugin-partnerships#235


asimpson:
@aangelisc do we have a timeline for that ui release?

src/views/CHConfigEditor.tsx
```diff
@@ -1,31 +1,20 @@
...
-  InlineField,
-  Input,
-} from '@grafana/ui';
+import { RadioButtonGroup, Switch, Input, Field, SecretInput, Button, InlineField, InlineFieldRow } from '@grafana/ui';
 import { CertificationKey } from '../components/ui/CertificationKey';
 import { Components } from './../selectors';
-import { config } from '@grafana/runtime';
 import { CHConfig, CHCustomSetting, CHSecureConfig, Protocol } from './../types';
 import { gte } from 'semver';
+import { ConfigSection, ConfigSubSection, DataSourceDescription } from '@grafana/experimental';
```
asimpson, line 12: Important to call out that these will end up being accepted into either `@grafana/ui` or another similar package that isn't experimental.

asimpson (dismissed):

src/views/CHConfigEditor.tsx
```diff
@@ -130,133 +116,152 @@ export const ConfigEditor: React.FC<Props> = (props) => {
...
             onChange={onUpdateDatasourceJsonDataOption(props, 'username')}
             label={Components.ConfigEditor.Username.label}
             aria-label={Components.ConfigEditor.Username.label}
             placeholder={Components.ConfigEditor.Username.placeholder}
-            tooltip={Components.ConfigEditor.Username.tooltip}
           />
-        </div>
-        <div className="gf-form">
-          <SecretFormField
+        </Field>
```
bossinc, line 210: The password can be left blank.

bossinc (changes requested):

src/views/CHConfigEditor.tsx
bossinc: It should not be required

src/views/CHConfigEditor.tsx
aangelisc: Would you prefer a comment in the code or on the PR description?

asimpson: Sorry, I was fine just calling it out here :sunglasses:

bossinc (approved):
!!

bossinc (approved):

bossinc (dismissed):

bossinc (approved):

PR #500, opened 2023-08-24
Bump node and go versions in GitHub Actions

adamyeats:
This PR bumps Go and Node.js versions in all CI actions. It also sets the Go version to `1.21` in `go.mod`.


aangelisc:
I think this PR can be closed now? cc @adamyeats

PR #512, opened 2023-09-06
allow to provide an additional url path

### LINKED ISSUE #499: Unable to add a host with proxy
Can we add server address with a path/context appended or is there a way we could do it with custom settings? 
We have our click house deployed behind a proxy, we are unable to create a datasource with our endpoint. 

Endpoint : https://abc.internal.com/clickhouse-proxy/

tatchi:
Fixes #499 

It turns out `clickhouse-go` already support this feature via [an HttpUrlPath option](https://github.com/ClickHouse/clickhouse-go/blob/main/clickhouse_options.go#L142)


aangelisc (commented):
This LGTM in general @tatchi, thank you for the contribution! Would you also be able to update the `CHConfigEditor`, `driver` and `settings` tests?

tatchi:
@aangelisc I added a few tests but I never wrote any `go` previously so I hope what I did makes sense. Please feel free to adjust if necessary.

asimpson (approved):
Thanks for this, it looks great and apologies again for the delay! :trophy:

PR #524, opened 2023-09-20
Query Editor refactor

aangelisc:
Revamp query editor to use newer Grafana UI components. This switches `InlineField`/`InlineFormLabel` components to `EditorField`. A query header has been added as it seems to make sense stylistically (but I'm no design expert so we can always revert this change).

All tests have been updated and the styles that were unused have been removed. I've also slightly refactored some of the logic to simplify.

ClickHouseQueryEditor mov

A part of grafana/grafana#6479 and grafana/oss-plugin-partnerships#225

Fixes grafana/oss-plugin-partnerships#273

### DIFF
--- a/src/components/FormatSelect.tsx
+++ b/src/components/FormatSelect.tsx
@@ -1,48 +1,40 @@
 import React from 'react';
-import { InlineFormLabel, Select } from '@grafana/ui';
 import { selectors } from './../selectors';
 import { Format } from '../types';
-import { styles } from '../styles';
+import { InlineSelect } from '@grafana/experimental';
 
 export type Props = { format: Format; value?: string; onChange: (format: Format) => void };
 
 export const FormatSelect = (props: Props) => {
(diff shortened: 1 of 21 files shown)


bossinc:
I really like this redesign. The only nit I have is I think the format as drop down might be too hidden away.
Also I wonder if we can remove the run query button

aangelisc:
> I really like this redesign. The only nit I have is I think the format as drop down might be too hidden away.
Also I wonder if we can remove the run query button

We could always move it into the top row, it's something we do in quite a few datasources so I mimicked it but it doesn't have to be that way. Maybe we could discuss what we think is a header item vs a main editor item?

I kinda like having the run query button but that's an aesthetics thing. I'm not sure if we should always be running queries on blur/on query change. Can clickhouse queries be expensive/return a lot of data where it may be a problem?

bossinc (approved):

Issue #520, opened 2023-09-14
Allow pass SETTINGS together with SQL statement.


araujobsd:
Add the possibility for AST to understand queries using SETTINGS. Without SETTINGS, it is not possible to make a QUERY to use clickhouse query_cache.

**What happened**:
When run an SQL query that enables query_cache, the AST can't interpret the parameter SETTINGS.

Example of query:
 ```
SELECT count(*) FROM mytable SETTINGS use_query_cache=true, query_cache_ttl=300
```

Using the browser Inspector it is possible to see the following error:
```
Failed to parse SQL statement into an AST: Error: Syntax error at line 13 col 15:

  ) AS subquery SETTINGS
```

**What you expected to happen**:
Clickhouse server will receive the parameter SETTINGS and will place the query on system.query_cache and the subsequent query would have used the query_cache instead to perform a FULL query again.

**How to reproduce it (as minimally and precisely as possible)**:
Just add any dash and create a simple SQL query adding at the end of your SQL statement ```SETTINGS use_query_cache=true, query_cache_ttl=300```

**Anything else we need to know?**:

![ast](https://github.com/grafana/clickhouse-datasource/assets/6048727/55641eed-f2d4-467d-bbc6-a1d3ea419aac)

**Environment**:

- Grafana version: 10.1.0
- Plugin version: 3.2.0
- OS Grafana is installed on: Debian 12
- User OS & Browser: Debian 12 & Firefox

bossinc:
I can reproduce this. We will have a fix soon. Thank you!

araujobsd:
Thank you for the prompt reply, if you need me test it, feel free to send patch on my way and I can do the QA :)

SpencerTorres:
@bossinc do you think this would be used frequently enough to have its own switch on the query builder?

asimpson:
Fixed in #530

PR #530, opened 2023-09-28
Queries with SETTINGS clauses no longer get AST errors

bossinc:
Settings are removed from the SQL before trying to make an ast. We do not need settings with the AST.

Solves https://github.com/grafana/clickhouse-datasource/issues/520


aangelisc (dismissed):

asimpson (approved):
Nice work on this 🏆

araujobsd:
Nice, thanks a lot :)

PR #597, opened 2023-11-15
Remove unused ComplexTypes map

adamyeats:
Removed `ComplexTypes` map that was unused.


asimpson (approved):
Good find, looks like it was left over from a larger piece of work :+1:

PR #605, opened 2023-11-16
V4 Plugin Upgrades

SpencerTorres:
# Grafana ClickHouse Datasource v4

This is a pull request to merge the v4 plugin changes into the upstream repo. This branch includes features and enhancements that were proposed by users and internal teams.

Please feel free to review and comment on the changes, and express any concerns with feature additions/changes.

### Query Builder

The query builder has been updated to support 4 separate views: **Table**, **Logs**, **TimeSeries**, and **Traces**. These are now separate components, which makes it easier to add and maintain view-specific features and behavior. The raw SQL editor is still available for advanced queries.

#### Table

A catch-all query builder. Useful for quickly fetching data. Formats as table in explore view, but works well for selecting columns for custom panels in the dashboard. Supports simple and aggregate query types.

image

#### Logs

A view focused on logs. In explore view this will render the logs panel. Open Telemetry (OTEL) is supported with a switch that automatically selects columns using an OTEL schema.

image

#### Time Series

A view focused on time series queries. In explore view this will show a time series panel. Similar to table view, but with a dedicated time field. Supports simple and aggregate query types.

image

#### Traces

A view for traces. Custom columns can be selected, but the quickest way to use it is by enabling the Open Telemetry (OTEL) table schema. Allows for loading a single trace, or for searching traces.

image

#### Raw SQL Editor

Raw SQL editor for going beyond the capabilities of the query builder. In explore view a format can be selected to change the panel type.

image

### Config View

Configuration options have been refactored to support adding default tables/columns for logs and traces. Open Telemetry (OTEL) is supported for easy selection of columns.

#### Logs Config

image

#### Traces Config

image

### Code Changes

Many upgrades have been made to the plugin code itself, mainly on the frontend app:

- Most of the code has been refactored/reorganized. Removed unused code, made names consistent, etc.
- The SQL generation logic has been updated for generating the more specialized logs/trace queries
- Column selection now uses a `SelectedColumn` object instead of a `string`. This enables more advanced logic in the SQL generator when converting column types, or when finding specialized (hinted) columns such as time/log columns/trace columns
- Query builder state now uses React's `useReducer` to simplify the state changes applied to the top-level `onChange` function
- Majority of the React `useEffect` hooks have been updated to properly use the dependency array (which fixed some state bugs)
- Complex components were moved into smaller components, which included raising the state in some cases
- Tests have been added for previously untested components/logic

### Changes from v3

Any feature enhancements or bug fixes from v3 will be re-added to this v4 branch. Some of them have already been included, but not all of them (such as the query editor component upgrades).


src/components/queryBuilder/DurationUnitSelect.tsx
```diff
@@ -0,0 +1,41 @@
...
+import allLabels from 'labels';
+import { InlineFormLabel, Select } from '@grafana/ui';
+import { SelectableValue } from '@grafana/data';
+import { styles } from 'styles';
+
+interface DurationUnitSelectProps {
+  unit: TimeUnit;
+  onChange: (u: TimeUnit) => void;
+  disabled?: boolean;
+  inline?: boolean;
```
asimpson, line 12: We can probably make a decision on this and remove the need for a prop.

src/components/queryBuilder/FilterEditor.tsx
```diff
@@ -329,38 +332,44 @@ export const FilterEditor = (props: {
     onFilterChange(index, filter);
   };
   return (
-    <EditorFieldGroup>
+    <>
       {index !== 0 && (
         <RadioButtonGroup options={conditions} value={filter.condition} onChange={(e) => onFilterConditionChange(e!)} />
       )}
       <Select
+        disabled={Boolean(filter.hint)}
```
asimpson, line 340: Would it be more clear to have a specific prop/state to handle disabling the Select?

src/components/queryBuilder/views/LogsQueryBuilder.tsx
```diff
@@ -0,0 +1,310 @@
...
+import { LimitEditor } from '../LimitEditor';
+import { FiltersEditor } from '../FilterEditor';
+import allLabels from 'labels';
+import { getColumnByHint } from 'data/sqlGenerator';
+import { columnFilterDateTime, columnFilterString } from 'data/columnFilters';
+import { Datasource } from 'data/CHDatasource';
+import { useBuilderOptionChanges } from 'hooks/useBuilderOptionChanges';
+import { versions as otelVersions } from 'otel';
+import { Alert, VerticalGroup } from '@grafana/ui';
+import useColumns from 'hooks/useColumns';
```
asimpson, line 16: Why are  these separate imports from the hooks below?

asimpson (changes requested):
:exploding_head: This is a _ton_ of work, super well done!! :trophy: 

I clicked around a bit and things seem to work fine. We are missing a closing bracket [here](https://github.com/grafana/clickhouse-datasource/blob/53cd07a93f2b5b31e93209f9fa22e965fe26f250/pkg/plugin/settings.go#L87) which prevents the build from succeeding. I left a few other comments and questions in a few spots. 

I definitely would appreciate @bossinc's :eyes: on this as well since he has more/better context for some of these changes than I do.

pkg/plugin/settings.go
```diff
@@ -13,24 +13,29 @@ import (
...
-	Port               int64  `json:"port,omitempty"`
-	Path               string `json:"path,omitempty"`
-	Username           string `json:"username,omitempty"`
-	DefaultDatabase    string `json:"defaultDatabase,omitempty"`
-	InsecureSkipVerify bool   `json:"tlsSkipVerify,omitempty"`
-	TlsClientAuth      bool   `json:"tlsAuth,omitempty"`
-	TlsAuthWithCACert  bool   `json:"tlsAuthWithCACert,omitempty"`
-	Password           string `json:"-,omitempty"`
-	TlsCACert          string
+	Host     string `json:"host,omitempty"`
```
jkaflik, line 16: Since it's a major update, it's ok, but do you think we can avoid this breaking change?

pkg/plugin/settings.go
```diff
@@ -13,24 +13,29 @@ import (
...
-	Protocol           string          `json:"protocol"`
-	CustomSettings     []CustomSetting `json:"customSettings"`
-	ProxyOptions       *proxy.Options
+
+	Username string `json:"username,omitempty"`
+	Password string `json:"-,omitempty"`
+
+	DefaultDatabase string `json:"defaultDatabase,omitempty"`
+
+	DialTimeout  string `json:"dialTimeout,omitempty"`
```
jkaflik, line 34: Do you think we can introduce a sort of backward compatibility for `timeout` field? We don't need it, but would be nice to make users keep their configuration unchanged.

jkaflik (commented):
It would be great if we can make migration to v4 seamless on configuration schema.

_I did review backend changes only._

pkg/plugin/settings.go
jkaflik: Or we introduce both old and new fields on backward compatibility.

src/components/SqlEditor.test.tsx
```diff
@@ -38,18 +38,18 @@ describe('SQL Editor', () => {
...
-      );
-      expect(screen.queryByText('test')).toBeInTheDocument();
-      await userEvent.click(screen.getByTestId(Components.QueryEditor.CodeEditor.Expand));
-      expect(onChangeValue).toHaveBeenCalledTimes(1);
-    });
+      ));
+
+    expect(screen.queryByText('test')).toBeInTheDocument();
+    await userEvent.click(screen.getByTestId(Components.QueryEditor.CodeEditor.Expand));
+    expect(onChangeValue).toHaveBeenCalledTimes(2);
```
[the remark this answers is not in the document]
SpencerTorres: Good catch, this test was behaving oddly and I'm not sure it was ever correct. This component was using `useEffect` incorrectly and unnecessarily and it revealed another issue with the test. The code editor never gets mounted (or is delayed) and the `onToggleExpand` function has an early return for when it's undefined.

I've simplified the state of the component, just need to adjust the test to wait for the code editor to mount

src/components/SqlEditor.tsx
```diff
@@ -105,26 +124,33 @@ export const SQLEditor = (props: SQLEditorProps) => {
...
-        value={query.rawSql || ''}
-        onSave={onSqlChange}
-        showMiniMap={false}
-        showLineNumbers={true}
-        onBlur={(text) => onChange({ ...query, rawSql: text })}
-        onEditorDidMount={(editor: any) => handleMount(editor)}
-      />
-    </div>
+    <>
+      { app === CoreApp.Explore &&
```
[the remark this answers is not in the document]
SpencerTorres: In explore view we show the query type so that the user can pick which format/panel will appear. This isn't necessary in dashboard view since they manually select their preferred panel

image

src/components/queryBuilder/FilterEditor.tsx
SpencerTorres: I suppose so, but this component is only used within this file, so it would be the same line moved elsewhere. Might make the tests clearer though

src/components/queryBuilder/views/LogsQueryBuilder.tsx
SpencerTorres: Depends on what you mean, I think the imports might have gotten out of order at some point. They're also in different files, although we could group them into one file, especially since some of them aren't default exports

pkg/plugin/settings.go
SpencerTorres: I've updated it so that the struct has only the new fields, but `LoadSettings()` will try to read from the old JSON fields first. If the new fields are present they will be overwritten:

```go
if jsonData["server"] != nil {
	settings.Host = jsonData["server"].(string)
}
if jsonData["host"] != nil {
	settings.Host = jsonData["host"].(string)
}
```

pkg/plugin/settings.go
SpencerTorres: Added this + tests for `Host` (`Server`) and `DialTimeout` (`Timeout`)

pkg/plugin/settings.go
mshustov: What is the purpose of settings renaming? It creates additional friction for the migration, so we should challenge every breaking change.

src/components/SqlEditor.tsx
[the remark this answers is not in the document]
mshustov: It's a good candidate for an additional comment when someone asks a `Why?` question. Please, consider adding a comment

src/components/SqlEditor.tsx
```diff
@@ -105,26 +116,33 @@ export const SQLEditor = (props: SQLEditorProps) => {
...
+      <div className={styles.Common.wrapper}>
+        <a
+          onClick={() => onToggleExpand()}
+          className={styles.Common.expand}
+          data-testid={selectors.components.QueryEditor.CodeEditor.Expand}
+        >
+          <i className={`fa fa-${expand.icon}`}></i>
+        </a>
+        <CodeEditor
+          aria-label="SQL"
```
mshustov, line 134: ni: not very descriptive

src/components/SqlEditor.tsx
```diff
@@ -105,26 +116,33 @@ export const SQLEditor = (props: SQLEditorProps) => {
...
+          height={expand.height}
+          language="sql"
+          value={query.rawSql || ''}
```
mshustov, line 137: In what cases `rawSql` in falsy? the typing says it's `string`

src/components/SqlEditor.tsx
```diff
@@ -1,44 +1,55 @@
...
 
-export const SQLEditor = (props: SQLEditorProps) => {
+export const SqlEditor = (props: SqlEditorProps) => {
   const defaultHeight = '150px';
-  const { query, onRunQuery, onChange, datasource } = props;
+  const { app, query, onChange, datasource } = props;
+  const sqlQuery = query as CHSqlQuery;
   const [codeEditor, setCodeEditor] = useState<any>();
-  const [expand, setExpand] = useState<Expand>({
+  const [expand, _setExpand] = useState<Expand>({
```
mshustov, line 29: nit: let's not break the naming pattern. `setExpand` declared below can have another name (`updateExpand`, `toggleExpand`, etc.)

src/components/SqlEditor.tsx
```diff
@@ -1,44 +1,55 @@
...
 import { CodeEditor } from '@grafana/ui';
-import { Datasource } from '../data/CHDatasource';
+import { Datasource } from 'data/CHDatasource';
 import { registerSQL, Range, Fetcher } from './sqlProvider';
-import { CHQuery, CHConfig, QueryType, CHSQLQuery } from '../types';
-import { styles } from '../styles';
+import { CHConfig } from 'types/config';
+import { CHQuery, EditorType, CHSqlQuery } from 'types/sql';
+import { styles } from 'styles';
 import { fetchSuggestions as sugg, Schema } from './suggestions';
```
mshustov, line 9: nit: Let's avoid unnecessary abbreviations. `fetchSuggestions` is more self-descriptive than `sugg`

src/components/SqlEditor.tsx
```diff
@@ -1,44 +1,55 @@
...
   icon: 'plus' | 'minus';
   on: boolean;
 }
 
-export const SQLEditor = (props: SQLEditorProps) => {
+export const SqlEditor = (props: SqlEditorProps) => {
   const defaultHeight = '150px';
-  const { query, onRunQuery, onChange, datasource } = props;
+  const { app, query, onChange, datasource } = props;
+  const sqlQuery = query as CHSqlQuery;
```
mshustov, line 27: Can we declare `query` type as a union and enforce it via contract, instead of casting manually?

src/components/SqlEditor.test.tsx
```diff
@@ -36,20 +36,20 @@ describe('SQL Editor', () => {
...
-      );
-      expect(screen.queryByText('test')).toBeInTheDocument();
-      await userEvent.click(screen.getByTestId(Components.QueryEditor.CodeEditor.Expand));
-      expect(onChangeValue).toHaveBeenCalledTimes(1);
-    });
+      ));
+
+    expect(result.queryByText('test')).toBeInTheDocument();
+    await userEvent.click(result.getByTestId(Components.QueryEditor.CodeEditor.Expand));
+    expect(onChange).toHaveBeenCalledTimes(0); // TODO: codeEditor isn't mounting and wont call onChange
```
mshustov, line 53: Shouldn't we have a test for `collapse`?

src/views/CHQueryEditor.tsx
```diff
@@ -1,95 +1,91 @@
...
-            app={app}
-          />
-          <Preview sql={newQuery.rawSql} />
-        </div>
-      );
+  /**
+   * Grafana will sometimes replace the builder options directly, so we need to sync in both directions.
+   * For example, selecting an entry from the query history will cause the local state to fall out of sync.
+   * The "key" property is present on these historical entries.
+   */
```
mshustov, line 42: Should we add some tests for this logic?

src/types/sql.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export interface CHQueryBase extends DataQuery {
+  editorType: EditorType;
+  rawSql: string;
+
+  /**
+   * REQUIRED by backend for auto selecting preferredVisualisationType.
+   * Only used in explore view.
+   * src: https://github.com/grafana/sqlds/blob/main/query.go#L36
```
mshustov, line 19: nit: let's link to a specific commit, links to a branch are fragile since branches are mutable.

src/otel.ts
```diff
@@ -0,0 +1,33 @@
...
+  version: string;
+  logColumnMap: Map<ColumnHint, string>;
+  traceColumnMap: Map<ColumnHint, string>;
+  traceDurationUnit: TimeUnit.Nanoseconds;
+}
+
+export const versions: readonly OtelVersion[] = [
+  {
+    name: 'latest',
+    version: '1.26.0',
```
mshustov, line 14: Could you add a link to the schema definition of `1.26.0` version?

src/labels.ts
```diff
@@ -0,0 +1,297 @@
+export default {
+  components: {
```
mshustov, line 2: Thank you for collocating in a single place. it will help with adding i18n [once it's supported by the plugin API](https://github.com/grafana/grafana/issues/60847)

src/hooks/useTables.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export default (datasource: Datasource, database: string): string[] => {
+  const [tables, setTables] = useState<string[]>([]); 
+  
+  useEffect(() => {
+    if (!datasource || !database) {
+      return;
+    }
+
+    let ignore = false;
```
mshustov, line 12: In what cases `ignore` should be true?

src/hooks/useDatabases.ts
```diff
@@ -0,0 +1,21 @@
+import { useState, useEffect } from 'react';
+import { Datasource } from 'data/CHDatasource';
+
+export default (datasource: Datasource): string[] => {
```
mshustov, line 4: nit: Let's return `readonly array` as you did in `useColumns`. 
A rule of thumb: use `readonly array` by default. for cases when Grafana API expects a mutable array, we can make a copy in place `[...readonlyArray]`

src/hooks/useBuilderOptionChanges.ts
```diff
@@ -0,0 +1,33 @@
...
+ * 
+ * @param onChange a function that receives the updated state from the change function
+ * @param prevState the current (previous) state object
+ * @returns a function used to apply changes to individual fields
+ */
+export function useBuilderOptionChanges<T>(onChange: (nextState: T) => void, prevState: T): onOptionChangeFn<T> {
+  return (key: keyof T | Object) =>
+    (nextValue: React.SetStateAction<any>) => {
+    let nextState: T;
+    if (typeof key === 'object') {
```
mshustov, line 19: It will be true for `null` and `Array` as well

src/data/sqlGenerator.test.ts
```diff
@@ -0,0 +1,125 @@
+import { AggregateType, ColumnHint, QueryBuilderOptions, QueryType } from 'types/queryBuilder';
+import { generateSql, getColumnByHint, getColumnIndexByHint, getColumnsByHints, isAggregateQuery } from './sqlGenerator';
+
+describe('SQL Generator', () => {
```
mshustov, line 4: can be extended with tests for filters

src/data/columnFilters.ts
```diff
@@ -0,0 +1,13 @@
+import { SelectedColumn } from "types/queryBuilder";
+
+
+export const columnFilterDateTime = (s: SelectedColumn): boolean => (s.type || '').toLowerCase().startsWith('date');
```
mshustov, line 4: What if it's a `Nullable(Date*)`?

src/components/queryBuilder/views/TraceQueryBuilder.tsx
```diff
@@ -0,0 +1,368 @@
...
+        traceDurationUnit: defaultDurationUnit
+      }
+    }));
+    didSetDefaults.current = true;
+  }, [builderOptions.columns, builderOptions.orderBy, builderOptions.table, builderOptionsDispatch, datasource, isNewQuery]);
+};
+/**
+ * Sets OTEL Trace columns automatically when OTEL is enabled
+ */
+const useOtelColumns = (otelEnabled: boolean, otelVersion: string, builderOptionsDispatch: React.Dispatch<BuilderOptionsReducerAction>) => {
```
mshustov, line 338: Nit: we could split the file into smaller parts.

src/components/queryBuilder/views/LogsQueryBuilder.tsx
```diff
@@ -0,0 +1,310 @@
...
+    }));
+  }, builderState);
+
+  useLogDefaultsOnMount(datasource, isNewQuery, builderOptions, builderOptionsDispatch);
+  useOtelColumns(builderState.otelEnabled, builderState.otelVersion, builderOptionsDispatch);
+  useDefaultTimeColumn(datasource, allColumns, builderOptions.table, builderState.timeColumn, builderState.otelEnabled, builderOptionsDispatch);
+  useDefaultFilters(builderOptions.table, builderState.timeColumn, builderState.filters, builderState.orderBy, builderOptionsDispatch);
+  
+  const configWarning = showConfigWarning && (
+    <Alert title="" severity="warning">
```
mshustov, line 89: can is be closed?

src/data/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+    case QueryType.Table:
+      return 1;
+    case QueryType.Logs:
+      return 2;
+    case QueryType.TimeSeries:
+      return 0;
+    case QueryType.Traces:
+      return 3;
+    default:
+      return 1 << 8; // an unused u32, defaults to timeseries/graph on plugin backend.
```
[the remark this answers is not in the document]
SpencerTorres: [Yes](https://github.com/grafana/sqlds/blob/main/query.go#L22)

However, there is another place in the backend where `PreferredVisualisationType` will default to TimeSeries if `format` isn't provided (or in this case, if the number doesn't match). I wanted to prioritize the backend's logic in case the auto data/panel type detection becomes more advanced in the future.

I also recall trying a simple `-1` but it wanted an unsigned number so thats why I went with an unusably large value. 

v3 used to use `4` to achieve this same effect, but that would cause an issue if they ever added another visualization type

src/components/queryBuilder/views/LogsQueryBuilder.tsx
SpencerTorres: Yes! Checked the docs again and they have `buttonContent` and `onRemove`. Added for Logs and Trace views.

src/views/CHQueryEditor.tsx
```diff
@@ -1,95 +1,91 @@
...
+    const sql = generateSql(builderOptions);
+    onChange({
+      ...query,
+      editorType: EditorType.Builder,
+      rawSql: sql,
+      builderOptions,
+      format: mapQueryTypeToGrafanaFormat(builderOptions.queryType)
+    });
+
+    // TODO: fix dependency warning with "useEffectEvent" once added to stable version of react
```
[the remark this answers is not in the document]
SpencerTorres: Yes it is, but the `useEffectEvent` hook is still experimental and not available in our version:

https://react.dev/learn/separating-events-from-effects#declaring-an-effect-event

This would likely be the best solution to this portion of code, although I would probably need to re-visit it again to be sure. Right now the dependency array isn't being used correctly which can lead to unexpected bugs if someone tries to modify this in the future

src/labels.ts
SpencerTorres: I noticed the `selectors.ts` file contained both `data-test-id` test selectors as well as component labels, so I made this file so that they can be properly separated.

Makes it much easier to rename things 👍

src/components/SqlEditor.tsx
SpencerTorres: When the editor first loads (sql or builder), it only contains `datasource` and `refId`. We could mark all the properties as optional to match this behavior, but it's only true before the first edit to the query. After that all of the required properties are set.

In this case the `value` prop will be fine without the `|| ''` so I've removed it

src/hooks/useTables.ts
SpencerTorres: Added this to prevent race conditions when setting the state with the async table results. Must've shifted a line around, the cleanup effect incorrectly had `ignore = false` when it should've been `ignore = true`

src/data/sqlGenerator.ts
```diff
@@ -0,0 +1,351 @@
...
+}
+
+/**
+ * Generates trace query with columns that fit Grafana's Trace panel
+ * Column aliases follow this structure:
+ * https://grafana.com/docs/grafana/latest/explore/trace-integration/#data-frame-structure
+ */
+const generateTraceQuery = (options: QueryBuilderOptions): string => {
+  const { database, table } = options;
+  const limit = getLimit(options.limit);
```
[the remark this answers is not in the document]
SpencerTorres: Not sure why this is at the top, good nit 👍

src/data/CHDatasource.ts
```diff
@@ -569,6 +671,19 @@ export class Datasource
       throw err;
     }
   }
+
+  // interface DataSourceWithLogsContextSupport
```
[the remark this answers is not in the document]
SpencerTorres: It's already referenced at the top of the class if someone wanted to inspect the interface in-editor, I prefer that over a link. I added this comment as a marker for the 3 methods it requires

src/components/configEditor/LabeledInput.tsx
```diff
@@ -0,0 +1,30 @@
...
+    <div className="gf-form">
+      <InlineFormLabel width={12} className="query-keyword" tooltip={tooltip || label}>
+        {label}
+      </InlineFormLabel>
+      <Input
+        disabled={disabled}
+        width={30}
+        value={value}
+        onChange={e => onChange(e.currentTarget.value)}
+        placeholder={placeholder || label.toLowerCase().replace(/ /g, '_')}
```
mshustov, line 26: 1. why do we need this `replace` call?
2. shouldn't we apply the same logic to the provided input if it's needed?

src/components/configEditor/LogsConfig.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { OtelVersionSelect } from 'components/queryBuilder/OtelVersionSelect';
+import { ColumnHint } from 'types/queryBuilder';
+import { versions as otelVersions } from 'otel';
+import { LabeledInput } from './LabeledInput';
+import { CHLogsConfig } from 'types/config';
+import allLabels from 'labels';
+
+interface LogsConfigProps {
+  logsConfig?: CHLogsConfig;
+  onDefaultDatabaseChange: (v: any) => void;
```
mshustov, line 13: let's add more specific signatures for all the components below
```suggestion
  onDefaultDatabaseChange: (v: string) => void;
```

src/components/configEditor/LogsConfig.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const {
+    onDefaultDatabaseChange, onDefaultTableChange,
+    onOtelEnabledChange, onOtelVersionChange,
+    onTimeColumnChange, onLevelColumnChange, onMessageColumnChange
+  } = props;
+  let {
+    defaultDatabase, defaultTable,
+    otelEnabled, otelVersion,
+    timeColumn, levelColumn, messageColumn
+  } = (props.logsConfig || {}) as CHLogsConfig;
```
mshustov, line 32:
```suggestion
  } = (props.logsConfig || {});
```

src/components/queryBuilder/AggregateEditor.tsx
```diff
@@ -0,0 +1,149 @@
...
+}
+
+const aggregateOptions: Array<SelectableValue<AggregateType>> = [
+  { label: 'Count', value: AggregateType.Count },
+  { label: 'Sum', value: AggregateType.Sum },
+  { label: 'Min', value: AggregateType.Min },
+  { label: 'Max', value: AggregateType.Max },
+  { label: 'Average', value: AggregateType.Average },
+  { label: 'Any', value: AggregateType.Any },
+  // { label: 'Distinct Count', value: AggregateType.Count_Distinct },
```
mshustov, line 23: not supported?

src/components/queryBuilder/ColumnSelect.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    const result = render(
+      <ColumnSelect
+        allColumns={allColumns}
+        selectedColumn={selectedColumn}
+        onColumnChange={() => {}}
+        label={testLabel}
+        tooltip={testTooltip}
+      />
+    );
+    expect(result.container.firstChild).not.toBeNull();
```
mshustov, line 35: Let's test the content

src/components/queryBuilder/ColumnSelect.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export const ColumnSelect = (props: ColumnSelectProps) => {
+  const { allColumns, selectedColumn, onColumnChange, columnFilterFn, columnHint, label, tooltip, disabled, invalid, wide, inline, clearable } = props;
+  const selectedColumnName = selectedColumn?.name;
+  const columns: Array<SelectableValue<string>> = allColumns.
+    filter(columnFilterFn || defaultFilterFn).
+    map(c => ({ label: c.name, value: c.name }));
+
+  // Select component WILL NOT display the value if it isn't present in the options.
+  let staleOption = false;
```
mshustov, line 32: we need a focused test for this logic

src/components/queryBuilder/ColumnsEditor.tsx
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    const columnNames = selectedColumns.map(c => c.name);
+    const customColumns = getCustomColumns(columnNames, allColumns);
+    setCustomColumns(customColumns);
+  }, [allColumns, selectedColumns]);
+
+  const onChange = (selected: Array<SelectableValue<string>>): void => {
+    setIsOpen(false);
+    const selectedColumnNames = new Set<string>(selected.map(s => s.value!));
```
mshustov, line 51: nit: I'd consider extracting this logic into a helper to simplify testing

src/data/migration.ts
```diff
@@ -0,0 +1,191 @@
...
+export type AnyCHQuery = Partial<CHQuery> & {[k: string]: any};
+export type AnyQueryBuilderOptions = Partial<QueryBuilderOptions> & {[k: string]: any};
+
+/**
+ * Takes a CHQuery and transforms it to the latest interface.
+ * Returns undefined if no migration is applied.
+ */
+export const migrateCHQuery = (savedQuery: CHQuery): CHQuery | undefined => {
+  if (savedQuery.rawSql === undefined) {
+    return undefined;
```
mshustov, line 15: Why not return the same `savedQuery` to have a predictable signature `(CHQuery) => CHQuery`?

src/data/sqlGenerator.ts
```diff
@@ -0,0 +1,351 @@
...
+    filterParts.push(')');
+
+    const builtFilter = filterParts.join(' ');
+    builtFilters.push(builtFilter);
+  }
+
+  return builtFilters.join(' ');
+};
+
+const isBooleanType = (type: string): boolean => (type?.toLowerCase().startsWith('boolean'));
```
mshustov, line 338: This file must have better test coverage. can be done in follow-up.

src/components/queryBuilder/QueryTypeSwitcher.tsx
```diff
@@ -0,0 +1,46 @@
...
+  },
+  {
+    label: labels.types.QueryType.timeseries,
+    value: QueryType.TimeSeries,
+  },
+  {
+    label: labels.types.QueryType.traces,
+    value: QueryType.Traces,
+  },
+];
```
mshustov, line 29:
```suggestion
] as const;
```

src/components/queryBuilder/QueryTypeSwitcher.tsx
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Component for switching between the different query builder interfaces
+ */
+export const QueryTypeSwitcher = (props: QueryTypeSwitcherProps) => {
+  const { queryType, onChange, sqlEditor } = props;
+  const { label, tooltip, sqlTooltip } = labels.components.QueryTypeSwitcher;
+
+  return (
+    <span>
+      <InlineFormLabel width={8} className="query-keyword" tooltip={sqlEditor ? sqlTooltip : tooltip}>
```
mshustov, line 40: nit: should be tested

src/components/queryBuilder/OtelVersionSelect.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const { enabled, onEnabledChange, selectedVersion, onVersionChange, defaultToLatest, wide } = props;
+  const { label, tooltip } = selectors.components.OtelVersionSelect;
+  const options: SelectableValue[] = allVersions.map(v => ({
+    label: `${v.version}${v.name ? (` (${v.name})`) : ''}`,
+    value: v.version
+  }));
+
+  useEffect(() => {
+    if (defaultToLatest && selectedVersion === '') {
+      onVersionChange(allVersions[0].version);
```
mshustov, line 26: Will it trigger an update even if the switch is disabled?

src/components/queryBuilder/OrderByEditor.tsx
```diff
@@ -0,0 +1,168 @@
...
+  orderByOptions: Array<SelectableValue<string>>;
+  orderBy: OrderBy[];
+  onOrderByChange: (orderBy: OrderBy[]) => void;
+}
+export const OrderByEditor = (props: OrderByEditorProps) => {
+  const { orderByOptions, orderBy, onOrderByChange } = props;
+  const { label, tooltip, addLabel } = allLabels.components.OrderByEditor;
+
+  const addOrderByItem = () => {
+    const nextOrderBy: OrderBy[] = orderBy.slice();
```
mshustov, line 62: let's extract them in array/helpers and test properly

src/components/queryBuilder/OrderByEditor.tsx
```diff
@@ -0,0 +1,168 @@
...
+  columnOptions: Array<SelectableValue<string>>;
+  index: number,
+  orderByItem: OrderBy;
+  updateOrderByItem: (index: number, orderByItem: OrderBy) => void;
+}
+
+const sortOptions = [
+  { label: 'ASC', value: OrderByDirection.ASC },
+  { label: 'DESC', value: OrderByDirection.DESC },
+];
```
mshustov, line 24:
```suggestion
] as const;
```

src/components/queryBuilder/EditorTypeSwitcher.tsx
```diff
@@ -0,0 +1,108 @@
...
+import isString from 'lodash/isString';
+import { mapQueryTypeToGrafanaFormat } from 'data/utils';
+
+interface CHEditorTypeSwitcherProps {
+  query: CHQuery;
+  onChange: (query: CHQuery) => void;
+  onRunQuery: () => void;
+}
+
+const options: Array<SelectableValue<EditorType>> = [
```
mshustov, line 18:
```suggestion
const options: ReadonlyArray<SelectableValue<EditorType>> = [
```

src/components/queryBuilder/EditorTypeSwitcher.tsx
```diff
@@ -0,0 +1,108 @@
...
+        body={switcher.body}
+        confirmText={switcher.confirmText}
+        dismissText={switcher.dismissText}
+        icon="exclamation-triangle"
+        onConfirm={onConfirmEditorTypeChange}
+        onDismiss={() => setConfirmModalState(false)}
+      />
+      <ConfirmModal
+        title={cannotConvert.title}
+        body={`${errorMessage} \nDo you want to delete your current query and use the query builder?`}
```
mshustov, line 99: shouldn't it be in the labels file?

src/components/queryBuilder/EditorTypeSwitcher.tsx
```diff
@@ -0,0 +1,108 @@
...
+export const EditorTypeSwitcher = (props: CHEditorTypeSwitcherProps) => {
+  const { query, onChange } = props;
+  const { label, tooltip, switcher, cannotConvert } = labels.components.EditorTypeSwitcher;
+  const editorType: EditorType = query.editorType || EditorType.Builder;
+  const [confirmModalState, setConfirmModalState] = useState<boolean>(false);
+  const [cannotConvertModalState, setCannotConvertModalState] = useState<boolean>(false);
+  const [errorMessage, setErrorMessage] = useState<string>('');
+  const onEditorTypeChange = (editorType: EditorType, confirm = false) => {
+    // TODO: component state has updated, but not local state.
+    if (query.editorType === EditorType.SQL && editorType === EditorType.Builder && !confirm) {
```
mshustov, line 35: Does `confirm` mean "require confirmation" or "confirmed switching"? let's rename to show the intention

PR #629, opened 2023-12-11
Fix: remove ad-hoc filter parsing for numeric fields

afagundes:
This PR fixes issue [#655](https://github.com/grafana/clickhouse-datasource/issues/655)


asimpson:
I'm going to defer to @SpencerTorres on this one since he probably knows the implications on Clickhouse for this change.

bossinc:
This is tricky. It would be best to pass the correct type to CH so it doesn't have to parse strings. (if @SpencerTorres says it's not a big deal, then let's go with the current approach)
As an alternative approach, we could allow filter values not to add quotes if quotes are already in the value. So `'123'` wouldn't become `''123''`. What are your thoughts @afagundes

asimpson (approved):
LGTM, thanks! :rocket:

Issue #626, opened 2023-12-08
Add ability to set `HttpHeaders` via plugin options


adamyeats:
### Discussed in https://github.com/grafana/clickhouse-datasource/discussions/625

<div type='discussions-op-text'>

<sup>Originally posted by **ShravanSunder** December  8, 2023</sup>
I wanted to use this with grafana cloud to view data in cloudflare analytics engine that uses clickhouse.  https://developers.cloudflare.com/analytics/analytics-engine/grafana/

They need use to use authorisation header with https.  How can this be done in this plug in?

If not can we request this feature?</div>

SpencerTorres:
This would be useful. Some questions:

1. Are there any headers we would want to restrict adding? (Hop headers?)
2. Should it be a generic input for an array of headers, or should we just add `Authorization` for now?
3. Would any of these fields need to be stored securely? (Such as `Authorization` containing tokens)

ShravanSunder:
@SpencerTorres this is the link to cloudflare's click house offering which would be great to connect to grafana:  https://developers.cloudflare.com/analytics/analytics-engine/grafana/#grafana-plugin-setup
```
Configure the plugin as follows:

URL: https://api.cloudflare.com/client/v4/accounts/<account_id>/analytics_engine/sql. Replace <account_id> with your 32 character account ID (available in the Cloudflare dashboard).
Leave all auth settings off.
Add a custom header with a name of Authorization and value set to Bearer <token>. Replace <token> with suitable API token string (refer to the [SQL API docs](https://developers.cloudflare.com/analytics/analytics-engine/sql-api/#authentication) for more information on this).
No other options need to be set.
```

ShravanSunder:
@adamyeats Any thoughts on when you would be able to support this.  thanks :)

adamyeats:
@ShravanSunder See https://github.com/grafana/clickhouse-datasource/pull/633

SpencerTorres:
As @adamyeats mentioned, we have a PR open to add this to the v4 branch in #633. v4 will be released soon but if you want to build your own branch asap, I recommend using my `v4-fixes` branch from #666 which includes the latest version of v4 (and includes the http header changes)

ShravanSunder:
@SpencerTorres i'm attempting to use this with grafana cloud, so i'll wait for the PR to be merged :) thanks!

SpencerTorres:
[...]

PR #633, opened 2023-12-15
Custom HTTP Headers

SpencerTorres:
Allows adding custom HTTP headers. (Requested in #626)

- All header names are saved to `jsonData`.
- Plain text header values are saved to `jsonData`.
- Secure header values are saved to `secureJsonData`.

Includes frontend + backend changes. 

Also includes some label updates to the config page when selecting a port.

New HTTP Headers section

### DIFF
--- a/pkg/plugin/settings.go
+++ b/pkg/plugin/settings.go
@@ -3,6 +3,7 @@ package plugin
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/ClickHouse/clickhouse-go/v2"
 	"strconv"
 	"strings"
 	"time"
@@ -34,7 +35,8 @@ type Settings struct {
 	DialTimeout  string `json:"dialTimeout,omitempty"`
 	QueryTimeout string `json:"queryTimeout,omitempty"`
 
-	CustomSettings []CustomSetting `json:"customSettings"`
+	HttpHeaders    map[string]string `json:"-"`
+	CustomSettings []CustomSetting   `json:"customSettings"`
 	ProxyOptions   *proxy.Options
 }
 
@@ -42,6 +44,8 @@ type CustomSetting struct {
 	Setting string `json:"setting"`
 	Value   string `json:"value"`
 }
+
+const secureHeaderKeyPrefix = "secureHttpHeaders."
 
 func (settings *Settings) isValid() (err error) {
 	if settings.Host == "" {
@@ -187,8 +191,12 @@ func LoadSettings(config backend.DataSourceInstanceSettings) (settings Settings,
 		settings.TlsClientKey = tlsClientKey
 	}
 
-	// proxy options are only able to be loaded via environment variables
-	// currently, so we pass `nil` here so they are loaded with defaults
+	if settings.Protocol == clickhouse.HTTP.String() {
+		settings.HttpHeaders = loadHttpHeaders(jsonData, config.DecryptedSecureJSONData)
+	}
+
+	// proxy options are currently only able to load via environment variables,
+	// so we pass `nil` here so that they are loaded with defaults
 	proxyOpts, err := config.ProxyOptions(nil)
 
 	if err == nil && proxyOpts != nil {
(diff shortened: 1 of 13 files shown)


adamyeats:
@SpencerTorres Thanks for this! It looks like there's a bunch of extra historical commits on this branch, would you be able to look into this?

SpencerTorres:
This was developed on the v4 branch which is what I'm merging in to. The file changes look correct, but I think all the commits were squashed in that first v4.x PR

pkg/plugin/settings.go
```diff
@@ -203,3 +211,30 @@ func LoadSettings(config backend.DataSourceInstanceSettings) (settings Settings,
...
+func loadHttpHeaders(jsonData map[string]interface{}, secureJsonData map[string]string) map[string]string {
+	httpHeaders := make(map[string]string)
+
+	if jsonData["httpHeaders"] != nil {
+		httpHeadersRaw := jsonData["httpHeaders"].([]interface{})
+
+		for _, rawHeader := range httpHeadersRaw {
+			header := rawHeader.(map[string]interface{})
+			headerName := header["name"].(string)
+			headerValue := header["value"].(string)
```
jkaflik, line 225: It risks panic

```suggestion
                         header, _ := rawHeader.(map[string]interface{})
			headerName, _ := header["name"].(string)
			headerValue, _ := header["value"].(string)
```

jkaflik: Probably trim keys would be good

jkaflik (changes requested):
I've left one detail, but it's quite important.

BTW should this be mentioned in README? Would be great if we mention how to use it in JSON configuration: https://github.com/grafana/clickhouse-datasource?tab=readme-ov-file#with-a-configuration-file

src/components/configEditor/HttpHeadersConfig.tsx
```diff
@@ -0,0 +1,176 @@
...
+    </div>
+  );
+}
+
+/**
+ * Returns a Set of all secured headers that are configured
+ */
+export const useConfiguredSecureHttpHeaders = (secureJsonFields: KeyValue<boolean>): Set<string> => {
+  return useMemo(() => {
+    const secureHeaders = new Set<string>();
```
asimpson, line 168: :heart_eyes: dig it

src/views/CHConfigEditorHooks.ts
```diff
@@ -35,3 +35,59 @@ export const useMigrateV3Config = (options: DataSourceSettings<CHConfig>, onOpti
...
+        header.value = '';
+      }
+    }
+
+    httpHeaders.push(header);
+  }
+
+  const currentSecureJsonFields: KeyValue<boolean> = { ...options.secureJsonFields };
+  for (let key in currentSecureJsonFields) {
+    if (!secureHttpHeaderKeys[key] && key.startsWith('secureHttpHeaders.')) {
```
asimpson, line 71: I don't know a better way but I wish we could have this synced somehow between go/TS so if it gets changed in one place it throws a compiler error or something.

asimpson (changes requested):
Just one UI thing that needs to be fixed.

:rocket: Looks great aside from that

src/components/configEditor/HttpHeadersConfig.tsx
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  const headerValueLabel = secure ? labels.secureHeaderValueLabel : labels.insecureHeaderValueLabel;
+  return (
+    <div data-testid={selectors.headerEditor}>
+      <HorizontalGroup>
+        <Field label={labels.headerNameLabel} aria-label={labels.headerNameLabel}>
+          <Input
+            data-testid={selectors.headerNameInput}
+            width={40}
```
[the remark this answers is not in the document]
SpencerTorres: I originally expanded the width for those long looong custom header names, like `X-Some-Extremely-Long-Custom-Header`, but I see how this fixes the bounds issue. The `Authorization` header fits nicely in the default width, so anyone with a lengthy header will just have to scroll

src/views/CHConfigEditorHooks.ts
SpencerTorres: I tried something with the TS [template literal types](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html), but there's no way to really enforce it considering that other secure keys might be added to the config object.

SpencerTorres:
Fixed the input box width for small screens,  thanks!

asimpson (approved):

asimpson:
Closing since this was included in #666

Issue #599, opened 2023-11-16
ad-hoc filter does not calculate correct target table 


mikhno-s:
**What happened**:
We added ad hoc filter to one of our dashboards and wanted to filter data on dashboards by clicking `filter` on the table cell of one of the tables in that dashboards.
For example I wanted to filter data by column `level` with value `ERROR`

**What you expected to happen**:
Clicking on the `Filter for value` on the table cell where level is `ERROR` and now expecting that `settings additional_table_filters={'app_logs' : ' level = \'ERROR\' '}` will be added to all queries on this dashboard with this datasource.

**How to reproduce it (as minimally and precisely as possible)**:
- Create a table app logs with some columns and column `level` type String in clickhouse and `timestamp` type DateTime64 with some test data within
- Add this clickhouse as a datasource to grafana
- Create a dashboard and dd a adhoc filter on this datasource
- Create a table visualisation and put some default query into it
```sql
SELECT
*
FROM logs_db.app_logs
WHERE $__timeFilter(timestamp)
LIMIT 10 
```
- Enable `Standard options -> Ad-hoc filterable` in grafana table settings
- Clickhouse `Filter for value` with some value, `ERROR` for example in the table
- Checkout the query in `query inspector`:
```sql
SELECT
*
FROM logs_db.app_logs
WHERE timestamp >= '1700137452' AND timestamp <= '1700141052'
LIMIT 10 settings additional_table_filters={'logs_db' : ' level = \'ERROR\' '}
```

The right query should be 
```sql
SELECT
*
FROM logs_db.app_logs
WHERE timestamp >= '1700137452' AND timestamp <= '1700141052'
LIMIT 10 settings additional_table_filters={'logs_db.app_logs' : ' level = \'ERROR\' '}
```

or `settings additional_table_filters={'app_logs' : ' level = \'ERROR\' '}`

I can suggest that plugin incorrectly gets the table name from the query.

**Environment**:

- Grafana version: 10.1.1
- Plugin version: 3.3.0

mikhno-s:
Hi, @SpencerTorres 
Do you have any ETA for this ?

cwurm:
@mikhno-s @vbichov  I reproduced the problem and opened a PR that will hopefully fix it for you: https://github.com/grafana/clickhouse-datasource/pull/650

vbichov:
the issue wasn't fixed:
it seems that the AST parser is not aware of grafana template variables - this causes it to fail 
[...]

PR #650, opened 2024-01-08
Fix default ad hoc filters

cwurm:
According to the README, it should be possible to use ad hoc filters without any special configuration:

https://github.com/grafana/clickhouse-datasource/blob/2573f9453ae7e7a95ced55b661ce965272f25a6a/README.md?plain=1#L211

But that's actually not the case. The current logic sets the target table for the filter to `default` when `$clickhouse_adhoc_query` is not set. This then leads to a query with a setting like `SETTINGS additional_table_filters = {'default':' field = value }` that doesn't work because it's not specifying the correct table. 

This fix removes setting the target table to `default` (which seems weird in any case since there isn't usually a table like that in ClickHouse; there is only a database by that name) and sets it to the table from the query instead if no table has been specified any other way. With that, ad hoc filters work without setting `$clickhouse_adhoc_query` or a default database in the datasource settings.

Fixes https://github.com/grafana/clickhouse-datasource/issues/599.


bossinc:
I am not able to make this work. I am using the default `ChickHouse - Query Analysis` dashboard with an added a adhoc var.
I set the adhoc var to `query_log.query_kind` with a value of `Create`. The query inspector returns `additional_table_filters={'query_log' : ' query_kind = \'Create\' '}` for the adhoc function. It should be `additional_table_filters={'system.query_log' : ' query_kind = \'Create\' '}` It is not filtering without the database before the table.

cwurm:
> This makes sense but, I am not able to make this work. I am using the default `ChickHouse - Query Analysis` dashboard with an added a adhoc var.

@bossinc  How are you adding the adhoc var? With that dashboard, I can add it via the `Query overview` panel: Edit -> Overrides -> Override 20 -> Add override property `Standard options -> Ad-hoc filterable` and then I can click on table cells in the `Type` column and it filters the whole dashboard.

cwurm:
> This makes sense but, I am not able to make this work. I am using the default `ChickHouse - Query Analysis` dashboard with an added a adhoc var. I set the adhoc var to `query_log.query_kind` with a value of `Create`. The query inspector returns `additional_table_filters={'query_log' : ' query_kind = \'Create\' '}` for the adhoc function. It should be `additional_table_filters={'system.query_log' : ' query_kind = \'Create\' '}` It is not filtering without the database before the table.

@bossinc  Ah, I think I know what you mean. You're creating a new variable on the dashboard level, type `Ad hoc filters` and then setting it to `query_log.query_kind` and `Create` while on the dashboard?

Seems that's another case that's not working (and hasn't worked before this PR, either).

I've fixed that as well now. Can you check that it's working for you?

bossinc:
@cwurm I think we are talking about two difference ad-hoc filters. I did not know that you could make overrides for ad-hoc filters in tables like you described. :laughing: 
I am referring to template variable adhoc filters found in dashboard settings. Here are the grafana docs: [adhoc filters](https://grafana.com/docs/grafana/latest/dashboards/variables/add-template-variables/#add-ad-hoc-filters)
In the CH plugin when filtering with adhoc filters, the data source adds the CH function `additional_table_filters`. When filtering with overrides it is filtering the response on the frontend instead of CH doing the filtering

cwurm:
> @cwurm I think we are talking about two difference ad-hoc filters. I did not know that you could make overrides for ad-hoc filters in tables like you described. 😆
> I am referring to template variable adhoc filters found in dashboard settings. Here are the grafana docs: [adhoc filters](https://grafana.com/docs/grafana/latest/dashboards/variables/add-template-variables/#add-ad-hoc-filters)

Ah, crossed wires! This should be working as well now as per my previous comment. 

> In the CH plugin when filtering with adhoc filters, the data source adds the CH function `additional_table_filters`. When filtering with overrides it is filtering the response on the frontend instead of CH doing the filtering

Hm, from what I can see, overrides also change the ClickHouse query.

bossinc:
> Hm, from what I can see, overrides also change the ClickHouse query.
You are right! I am trying to recreate what I thought I saw and the ad-hoc vars seem to be the same. :shrug: 

This PR looks good.

bossinc (approved):

PR #680, opened 2024-02-01
fix: set protocol to "native" by default

SpencerTorres:
Fixes this issue where the `protocol` is undefined. It's a small issue, but confusing for people who are setting up a new datasource.

Also removes an unused/broken dashboard from the preset dashboards.

Includes version increment + changelog

image


asimpson (approved):
:+1: nice nice

PR #682, opened 2024-02-02
fix: v3 config page migration

SpencerTorres:
Fixes the config migration from not running by combining the migration logic and default setting logic into one single hook.

Includes changelog+version increment

### The Problem

The v3 `server` field is not migrating to the v4 `host` field. This causes the UI to show the field as empty, even though the data is still present in the config.

### Why

There are two `useEffect` hooks that call `onOptionsChange`: `useMigrateV3Config` and `useConfigDefaults`.

`useConfigDefaults` overwrites the first call by `useMigrateV3Config`

### Will it still work?

Yes, the plugin backend is still loading the connection from the `server` field, it's backwards compatible even though the UI isn't loading it correctly. The data is also still present in the config state, it just isn't copied into the new field.

If the user decides to re-enter the `host` manually, then it will be updated to the value they enter and the old value will be ignored

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 4.0.2
+
+### Fixes
+
+- Fixed migration script not running when opening an existing v3 config
+
 ## 4.0.1
 
 ### Fixes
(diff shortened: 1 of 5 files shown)


asimpson (approved):
:+1: Nice work

Issue #684, opened 2024-02-06
Unexpected rollback of changes in SQL Editor


glushkovds:
**What happened**:

In SQL Editor tab, while writing a request, part of what was written is deleted. Moreover, if you press Ctrl + Z after the loss, the loss will return.

**What you expected to happen**:

Everything that is written remains.

**How to reproduce it (as minimally and precisely as possible)**:

1. Add new visualization "Time series" to dashboard
2. Go to SQL Editor tab
3. Expand the editor: click the "plus" icon in the top left corner
4. Type anything
5. Exit the editor by clicking outside the editor
6. Click inside the editor 
7. The text will return to the original

**Screenshots**

![grafana_bug2](https://github.com/grafana/clickhouse-datasource/assets/35626465/26f9ad1b-ee87-4e4b-a43d-0336949f8fea)

**Environment**:

- Grafana version: v10.3.1
- Plugin version: 4.0.2
- OS Grafana is installed on: Docker image "grafana/grafana"
- User OS & Browser: Both Win10 & Ubuntu, both Google Chrome & Mozilla Firefox

arniwesth:
I'm seeing similar issues. When clicking "Run Query" any SQL changes are reverted back. Saving does not seem to solve this either as changes are still reverted even when saving.

PR #687, opened 2024-02-07
Do not control SqlEditor height

mshustov:
closes https://github.com/grafana/clickhouse-datasource/issues/684

There is a bug in logic that on every change in the editor calls `saveChanges({expand:true})`, which resets the `rawSql` value back to the original state.
Instead of fixing this bug, I checked how Grafana handles this logic itself. From [this example](https://github.com/grafana/grafana/blob/f59e76d54b0f5ea11e23f2f59eb3281682c7d1d0/public/app/plugins/datasource/tempo/traceql/TraceQLEditor.tsx#L81), it looks like the built-in Grafana editor handles height updates inside `CodeEditor` component, so I removed the external logic to manage component height.

### Changelog
Remove manual expansion of SqlEditor.


src/components/SqlEditor.tsx
```diff
@@ -16,22 +15,25 @@ import { pluginVersion } from 'utils/version';
...
-  on: boolean;
+function setupAutoSize(editor: monacoTypes.editor.IStandaloneCodeEditor) {
+  const container = editor.getDomNode();
+  const updateHeight = () => {
+    if (container) {
+      const contentHeight = Math.max(100, Math.min(1000, editor.getContentHeight()));
+      const width = parseInt(container.style.width, 10);
+      container.style.width = `${width}px`;
+      container.style.height = `${contentHeight}px`;
+      console.log('contentHeight', contentHeight, 'width', width);
```
asimpson, line 26:
```suggestion
```
nit: we don't need this right?

asimpson (changes requested):
:+1: looks good to me, thanks!

Just one nit and I'll approve.

asimpson (approved):

asimpson:
@mshustov Looks like it. The build passed though which is what I care about. Can you merge `main` and I can merge the PR?

mshustov:
@asimpson could you merge?

Issue #720, opened 2024-02-20
Integer division causes loss of precision on trace span durations


SpencerTorres:
The OTel exporter defaults to nanosecond precision, but Grafana requires that span durations be in milliseconds. This causes a lot of rows to be returned with a `0` duration, and spans that run under 1ms to be ruined in the trace visualization.

Broken:
![span durations under 1ms are showing 0us](https://github.com/grafana/clickhouse-datasource/assets/28887171/1c7dbf26-6f49-46b0-b80f-bf2848c32c55)

Correct:
![correct spans, with results under 1ms](https://github.com/grafana/clickhouse-datasource/assets/28887171/fc659d49-351f-4515-b0ee-d61abbc989e3)

See in the correct result how the spans are correctly showing under 1ms

Issue #713, opened 2024-02-19
Clickhouse cpu, memory, and connection are growing abnormally


gfes980615:
**What happened**:
We found the number of ClickHouse connections, CPU, and memory will be increased too higher when we upgraded the Datasource version from 3.3.0 to 4.0.2

We have nearly 2000 queries per 1 min
When we use version 3.3.0 the metrics are normal, but in version 4.0.2 these metrics seems unusual growth.

**What you expected to happen**:
The Clickhouse resource usage will be the same 

**How to reproduce it (as minimally and precisely as possible)**:
Upgrade the version from 3.3.0 to 4.0.2

**Screenshots**

Clickhouse 3.3.0 metrics:
<img width="1438" alt="Screenshot 2024-02-19 at 15 35 46" src="https://github.com/grafana/clickhouse-datasource/assets/25905125/252b4e1f-7f27-48e5-a8d1-1622469af445">
<img width="1439" alt="image" src="https://github.com/grafana/clickhouse-datasource/assets/25905125/913b8bfa-ecef-4fd1-9b78-07f927d89559">

Clickhouse 4.0.2 metrics:
<img width="1440" alt="image" src="https://github.com/grafana/clickhouse-datasource/assets/25905125/83c74323-a013-40e7-95da-9a0c689df535">
<img width="1440" alt="image" src="https://github.com/grafana/clickhouse-datasource/assets/25905125/e12d031e-930b-4a04-9e9d-7aedde544425">

**Anything else we need to know?**:

**Environment**:

- Grafana version: 9.5.15
- Plugin version: 4.0.2
- OS Grafana is installed on: Alpine Linux
- User OS & Browser: Google Chrome
- Others: n/a

mshustov:
@gfes980615 could you check the number of queries in ClickHouse [query_log](https://clickhouse.com/docs/en/operations/system-tables/query_log) before and after the upgrade? Do you see any anomalies? Feel free to contact me on [ClickHouse public Slack](https://clickhouse.com/slack) to investigate the problem deeper

gfes980615:
Hi, @mshustov 

Here is our origin query in Grafana:
```sql
select dt, error / total as error_rate_5xx
from (
  Select dt, sum(value) as total, SUM( CASE WHEN "http_status_code" like '4%' THEN value ELSE 0 END) as error
  from dist_otel_ubt_table
  where 1=1 and 
[...]

Issue #699, opened 2024-02-13
new `$__timeFilter, $__fromTime, $__toTime` macros break `DateTime` primary key indexes


TimSimmons:
**What happened**:

After I upgraded the Grafana plugin, I noticed a lot of my queries got very slow. This was because the new macro logic for time filtering does `toDateTime64(1707855105945 / 1000, 3)` and that seems incompatible with primary key filtering when you order by a `DateTime` column.

**How to reproduce it (as minimally and precisely as possible)**:

```
CREATE TABLE grafana_plugin_repro (
    timestamp DateTime,
    value Int32
) ENGINE = MergeTree()
ORDER BY timestamp;

INSERT INTO grafana_plugin_repro (timestamp, value) SELECT
    now() - number AS timestamp,
    rand() % 100000 AS value
FROM numbers(100000000)

~~~~~

SELECT avg(value)
FROM grafana_plugin_repro
WHERE (timestamp >= toDateTime64(1707851505945 / 1000, 3)) AND (timestamp <= toDateTime64(1707855105945 / 1000, 3))

┌────────avg(value)─┐
│ 49614.35085611295 │
└───────────────────┘

1 row in set. Elapsed: 0.710 sec. Processed 100.00 million rows, 400.02 MB (140.76 million rows/s., 563.07 MB/s.) Peak memory usage: 401.89 MiB.


--- workarounds

SELECT avg(value)
FROM grafana_plugin_repro
WHERE (timestamp >= toUnixTimestamp(toDateTime64(1707851505945 / 1000, 3))) AND (timestamp <= toUnixTimestamp(toDateTime64(1707855105945 / 1000, 3)))

┌────────avg(value)─┐
│ 49625.68138138138 │
└───────────────────┘

1 row in set. Elapsed: 0.005 sec. Processed 6.03 thousand rows, 48.26 KB (1.27 million rows/s., 10.14 MB/s.) Peak memory usage: 4.03 MiB.

SELECT avg(value)
FROM grafana_plugin_repro
WHERE (timestamp >= toDateTime(1707851505945 / 1000, 0)) AND (timestamp <= toDateTime(1707855105945 / 1000, 0))

┌────────avg(value)─┐
│ 49625.68138138138 │
└───────────────────┘

1 row in set. Elapsed: 0.004 sec. Processed 6.03 thousand rows, 48.26 KB (1.58 million rows/s., 12.60 MB/s.) Peak memory usage: 4.03 MiB.

[...]

PR #721, opened 2024-02-20
fix: time macros, trace span durations under 1ms

SpencerTorres:
## Fixes
- #699
- #720

#699 fixed by reverting changes to `$__timeFilter` and instead adding `$__timeFilterMs` for the millisecond use case.
#720 fixed by using `multiply` instead of `intDiv`, resulting in a float for milliseconds instead of an integer.

Fixing #699 also possibly fixes the increased resource usage mentioned in #713

## Features
- `$__dateFilter` now specifically uses a `Date` type instead of inferring the type from a string
- Added new `$__fromTimeMs` macro that outputs a `DateTime(3)` for the "from" time
- Added new `$__toTimeMs` macro that outputs a `DateTime(3)` for the "to" time
- Added new `$__timeFilterMs` macro that outputs a filter using a `DateTime(3)` for the "from" time and "to" time


asimpson (changes requested):
We can remove this I think: https://github.com/grafana/clickhouse-datasource/blob/5a7f12304aba291553ba2bdea3ae6172797f02b6/pkg/macros/macros.go#L18-L23

README.md
```diff
@@ -193,10 +193,13 @@ WHERE $__timeFilter(date_time)
...
+| *$__dateFilter(columnName)*                  | Replaced by a conditional that filters the data (using the provided column) based on the date range of the panel                                                                    | `date >= toDate('2022-10-21') AND date <= toDate('2022-10-23')`                     |
+| *$__timeFilter(columnName)*                  | Replaced by a conditional that filters the data (using the provided column) based on the time range of the panel in seconds                                                         | `time >= toDateTime(1415792726) AND time <= toDateTime(1447328726)`                     |
+| *$__timeFilter_ms(columnName)*                  | Replaced by a conditional that filters the data (using the provided column) based on the time range of the panel in milliseconds                                                         | `time >= fromUnixTimestamp64Milli(1415792726123) AND time <= fromUnixTimestamp64Milli(1447328726456)`                     |
```
mshustov, line 198: Don't forget to update `CHANGELOG.md` file

SpencerTorres:
Removed unused code and updated changelog 👍

Issue #724, opened 2024-02-22
trace query: inspect otel_traces_trace_id_ts table to speed query by traceID


hanjm:
**What happened**:
It is amazing cool that the new verision plugin add trace query support.

but when data amout grow, the query by traceID performance downgrade beacuse the  query sql it not inspect  `otel_traces_trace_id_ts` table.

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/13442#issuecomment-1235284835
> the otel_traces_trace_id_ts table is the key design, it is a small table order by traceID, write by materialized view, so it is very fast to query the start time and end time for a traceID, then query the otel_traces table by traceID with time range effectively.

query sql example: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/clickhouseexporter/README.md?plain=1#L148

**What you expected to happen**:
Query by traceID inspect `otel_traces_trace_id_ts` 
**How to reproduce it (as minimally and precisely as possible)**:
Write many trace spans (more than 1 billion per day), query by traceID.
**Screenshots**

**Anything else we need to know?**:

**Environment**:
- Grafana version: 10.1
- Plugin version: 4.0.0
- OS Grafana is installed on: Linux
- User OS & Browser:
- Others:

SpencerTorres:
This is very interesting, thank you for mentioning this.

Submitted #725 to enable this for Trace ID queries with OTel. I tested it and can already see the performance improvements, even on a much smaller dataset.

However... when I go back further to older trace IDs, it no longer returns the trace data. It seems it's getting filtered out because the trace ID is no longer present in the `otel_traces_trace_id_ts` table. The trace is still there in the `otel_traces` table, so it's not like it's been deleted by a TTL.

This might need to be a separate switch on the UI for Trace ID queries.
If you're using Trace Search we could conditionally enable/disable the setting if the dashboard time is far enough back.

SpencerTorres:
As a reminder, this is now available in [4.0.3](https://github.com/grafana/clickhouse-datasource/releases)

hanjm:
great, Thanks.

PR #725, opened 2024-02-22
Optimize OTel Trace ID queries with skip index

SpencerTorres:
Adds the optimization requested in #724 

This optimization is only included in the SQL when OTel is enabled since the table is specific to OTel.

Here's the new SQL that is generated for Trace ID queries:

```diff
+ WITH
+ '207c6fb96eda8541559a283678587c86' as trace_id,
+ (SELECT min(Start) FROM "default"."otel_traces_trace_id_ts" WHERE TraceId = trace_id) as trace_start,
+ (SELECT max(End) + 1 FROM "default"."otel_traces_trace_id_ts" WHERE TraceId = trace_id) as trace_end
SELECT
"TraceId" as traceID, "SpanId" as spanID, "ParentSpanId" as parentSpanID, "ServiceName" as serviceName,
"SpanName" as operationName, "Timestamp" as startTime, multiply("Duration", 0.000001) as duration,
arrayMap(key -> map('key', key, 'value',"SpanAttributes"[key]), mapKeys("SpanAttributes")) as tags,
arrayMap(key -> map('key', key, 'value',"ResourceAttributes"[key]), mapKeys("ResourceAttributes")) as serviceTags
FROM "default"."otel_traces"
WHERE
- traceID = '207c6fb96eda8541559a283678587c86'
+ traceID = trace_id
+ AND startTime >= trace_start
+ AND startTime <= trace_end 
LIMIT 1000
```

Here's a rough graph of query duration before and after:
image

### DIFF
--- a/src/data/sqlGenerator.ts
+++ b/src/data/sqlGenerator.ts
@@ -1,5 +1,5 @@
 import { BooleanFilter, BuilderMode, ColumnHint, DateFilterWithValue, FilterOperator, MultiFilter, NumberFilter, QueryBuilderOptions, QueryType, SelectedColumn, StringFilter, TimeUnit } from 'types/queryBuilder';
-import { getVersion as getOtelVersion } from 'otel';
+import otel from 'otel';
 
 /**
  * Generates a SQL string for the given QueryBuilderOptions
@@ -148,7 +148,7 @@ const generateTraceIdQuery = (options: QueryBuilderOptions): string => {
 
   // Optimize trace ID filtering for OTel enabled trace lookups
   const hasTraceIdFilter = options.meta?.isTraceIdMode && options.meta?.traceId;
-  const otelVersion = getOtelVersion(options.meta?.otelVersion);
+  const otelVersion = otel.getVersion(options.meta?.otelVersion);
   const applyTraceIdOptimization = hasTraceIdFilter && options.meta?.otelEnabled && otelVersion;
   if (applyTraceIdOptimization) {
     const traceId = options.meta!.traceId;
--- a/src/data/sqlGenerator.ts
+++ b/src/data/sqlGenerator.ts
@@ -1,4 +1,5 @@
 import { BooleanFilter, BuilderMode, ColumnHint, DateFilterWithValue, FilterOperator, MultiFilter, NumberFilter, QueryBuilderOptions, QueryType, SelectedColumn, StringFilter, TimeUnit } from 'types/queryBuilder';
+import { getVersion as getOtelVersion } from 'otel';
 
 /**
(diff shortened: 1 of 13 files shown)


src/otel.ts
```diff
@@ -56,3 +59,10 @@ export const versions: readonly OtelVersion[] = [
 ];
 
 export const getLatestVersion = (): OtelVersion => versions[0];
+export const getVersion = (version: string | undefined): OtelVersion | undefined => {
```
asimpson, line 62: This is a major nit but should we rename this function to `getOtelVersion`? We use `import { versions as otelVersions}` exclusively and I wonder if we should rename that to `otelVersions` as well?

src/data/sqlGenerator.ts
```diff
@@ -145,24 +146,46 @@ const generateTraceIdQuery = (options: QueryBuilderOptions): string => {
   }
   const selectPartsSql = selectParts.join(', ');
 
+  // Optimize trace ID filtering for OTel enabled trace lookups
+  const hasTraceIdFilter = options.meta?.isTraceIdMode && options.meta?.traceId;
+  const otelVersion = getOtelVersion(options.meta?.otelVersion);
+  const applyTraceIdOptimization = hasTraceIdFilter && options.meta?.otelEnabled && otelVersion;
+  if (applyTraceIdOptimization) {
+    const traceId = options.meta!.traceId;
```
asimpson, line 154: 🤘 😏

asimpson (approved):
LGTM 🎉 

Just two nits, up to you on how you want to handle them 😎

SpencerTorres:
Good points on the otel functions, pushed a change to clean up the imports/references 👌

src/otel.ts
```diff
@@ -56,3 +59,16 @@ export const versions: readonly OtelVersion[] = [
...
+  if (!version) {
+    return;
+  }
+
+  return versions.find(v => v.version === version);
+};
+
+export default {
```
asimpson, line 70: :heart_eyes:

asimpson (approved):
:trophy: nicely done

tanner-bruce:
FYI this query scans the trace_id_ts table twice.

Compare with:
```sql
(SELECT (min(start), max(end) + 1) FROM tracing.otel_traces_trace_id_ts WHERE trace_id = ID) as bounds,
...
AND startTime >= bounds.1
AND startTime <= bounds.2 
```

SpencerTorres:
> FYI this query scans the trace_id_ts table twice.

Very good point, this PR was simply using the linked query on the original issue, but this seems like another optimization we can add. Thanks!

PR #731, opened 2024-02-27
fix(dashboard): cluster dashboard latest_fail_time should be at midnight

Defman:


aangelisc (approved):
Thanks for catching this @Defman 😊

aangelisc:
If you could update this with latest main we'll get this merged

PR #733, opened 2024-02-27
fix: Ad-Hoc filters not applied because variables are not recognized by the AST parser

### LINKED ISSUE #714: Ad-hoc does not work, failing grafana template variables parsing
Ad-hoc filter fails to work in queries with grafana's variables due to parse failing.

**What happened**:
- Added ad-hoc variable to dashboard with panels with clickhouse datasource .
- Selected a column to filter by with value

- See the data filter by ad-hoc filter

1 Add a clickhouse datasource
2 Create a dashboard with clickhouse datasource panel
3 Use any grafana's template in query, like SQL comment `/* ${__dashboard} */` 
4 Add variable ad-hoc with that datasource
5 Open JS console in browser
6 Select column from ad-hoc variable and you will see the error message like on this screenshot 
305916863-d4696b85-679a-4d84-8d59-cb8a5dbb0ec3

**Environment**:

- Grafana version: 10.3.3
- Plugin version: 4.0.2
- User OS & Browser: os x sonoma, chrome 120.0.6099.129

SpencerTorres:
Fixes #714 

Changes:
- Fixed the AST from not recognizing variables such as `${var}` `${var.key}` and `${var:singlequote}`. This was causing Ad-Hoc filters to not be applied, and prevented switching back to the Query Builder.
- Switching from SQL Editor back to Query Builder works again. This worked before, but was also broken when using variables. Also it would open an empty query, but now the AST is able to recognize the v4 builder options a little better.
- Fixed an issue where an empty database name would put `FROM "undefined"."table"`
- Multi filters (such as ones using `WHERE .. IN`) now allows for spaces. v3 allowed for `a,b,c` or `a, b, c`, v4 had just the first one but now it supports both. You can also include expressions and it works nicely, for example: `a, b, toString('c')`
- Added new unit tests that specifically focus the AST when using macros and variables.
- Re-added some tests for the AST + SQL generator. These were removed before because the SQL generator was separated from the utils module, but I've added it back. The original intent of these tests seems to be testing consistency when converting sql->builderOptions->sql.
- updated changelog

### DIFF
--- a/src/data/ast.test.ts
+++ b/src/data/ast.test.ts
@@ -15,5 +15,30 @@ describe('ast', () => {
       // this is formatted like this to match how pgsql generates its sql
       expect(toSql.statement(stm)).toEqual('SELECT (count (*) )  FROM foo');
     });
+
+    // https://github.com/grafana/clickhouse-datasource/issues/714
+    it('does not error when brackets/macros/variables are present', () => {
+      const errLog = jest.spyOn(console, 'error');
+      const sql = `
+        SELECT
+          *,
+          \$__timeInterval(timestamp),
+          '{"a": 1, "b": { "c": 2, "d": [1, 2, 3] }}'::json as bracketTest
+        FROM default.table
+        WHERE $__timeFilter(timestamp)
+        AND col != \${variable}
+        AND col != \${variable.key}
+        AND col != \${variable.key:singlequote}
+        AND col != '\${variable}'
+        AND col != '\${__variable}'
+        AND col != ('\${__variable.key}')
+      `;
+
+      const stm = sqlToStatement(sql);
+      const astSql = toSql.statement(stm);
+      expect(errLog).toHaveBeenCalledTimes(0);
+      expect(stm).not.toEqual({});
+      expect(astSql).not.toBeFalsy();
+    });
   });
 });
--- a/src/data/ast.test.ts
+++ b/src/data/ast.test.ts
@@ -20,6 +20,7 @@ describe('ast', () => {
     it('does not error when brackets/macros/variables are present', () => {
       const errLog = jest.spyOn(console, 'error');
       const sql = `
+        /* \${__variable} \${__variable.key} */
         SELECT
           *,
           \$__timeInterval(timestamp),
@@ -32,6 +33,7 @@ describe('ast', () => {
         AND col != '\${variable}'
         AND col != '\${__variable}'
         AND col != ('\${__variable.key}')
+        AND col != \${variable:singlequote}
       `;
 
       const stm = sqlToStatement(sql);
--- a/src/data/ast.ts
+++ b/src/data/ast.ts
@@ -1,30 +1,68 @@
 import { parseFirst, Statement, SelectFromStatement, astMapper, toSql, ExprRef } from 'pgsql-ast-parser';
 
-export function sqlToStatement(sql: string): Statement {
-  const replaceFuncs: Array<{
-    startIndex: number;
-    name: string;
-    replacementName: string;
-  }> = [];
-  //default is a key word in this grammar, but it can be used in CH
-  const re = /(\$__|\$|default|settings)/gi;
+interface ReplacePart {
+  startIndex: number;
+  name: string;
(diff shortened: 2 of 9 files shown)


src/data/ast.ts
```diff
@@ -1,31 +1,69 @@
...
     }
     sql = sql.substring(0, si) + replacementName + sql.substring(si + replaceFuncs[i].name.length);
   }
 
+  return [replaceFuncs, sql];
+}
+
+/**
+ * Replaces Grafana variables such as ${var} ${var.key} ${var.key:singlequote}
+ */
```
mshustov, line 40: let's add a link to appropriate section https://grafana.com/docs/grafana/latest/dashboards/variables/

src/data/ast.ts
```diff
@@ -1,31 +1,69 @@
...
+
+  let regExpArray: RegExpExecArray | null;
+  while ((regExpArray = variableRegex.exec(sql)) !== null) {
+    replaceVariables.push({ startIndex: regExpArray.index, name: regExpArray[0], replacementName: '' });
+  }
+
+  // need to process in reverse so starting positions aren't affected by replacing other things
+  for (let i = replaceVariables.length - 1; i >= 0; i--) {
+    const si = replaceVariables[i].startIndex;
+    const replacementName = 'v' + (Math.random() + 1).toString(36).substring(7);
```
mshustov, line 53: nit: let's extract in a separate function or use uuid

src/data/ast.ts
```diff
@@ -1,31 +1,69 @@
...
+  }
+
+  return [replaceVariables, sql];
+}
+
+export function sqlToStatement(sql: string): Statement {
+  const [replaceVars, variableSql] = replaceMacroVariables(sql);
+  const [replaceFuncs, macroSql] = replaceMacroFunctions(variableSql);
+  const replaceParts = replaceVars.concat(replaceFuncs);
+  sql = macroSql;
```
mshustov, line 65: let's avoid overriding variables. I'd rather rename the function signature
```typescript
export function sqlToStatement(rawSql: string): Statement {
  const [replaceVars, variableSql] = replaceMacroVariables(sql);
  const [replaceFuncs, sql] = replaceMacroFunctions(variableSql);
```

src/data/ast.test.ts
```diff
@@ -15,5 +15,30 @@ describe('ast', () => {
       // this is formatted like this to match how pgsql generates its sql
       expect(toSql.statement(stm)).toEqual('SELECT (count (*) )  FROM foo');
     });
+
+    // https://github.com/grafana/clickhouse-datasource/issues/714
+    it('does not error when brackets/macros/variables are present', () => {
```
mshustov, line 20: Let's add more tests in a follow-up. I'd add more built-in ClickHouse functions, [query params](https://clickhouse.com/docs/en/interfaces/cli#cli-queries-with-parameters)

src/data/ast.test.ts
```diff
@@ -15,5 +15,30 @@ describe('ast', () => {
...
+      const errLog = jest.spyOn(console, 'error');
+      const sql = `
+        SELECT
```
mshustov, line 23:
```suggestion
        /* ${__dashboard} */
        SELECT
```

src/data/sqlGenerator.ts
```diff
@@ -515,12 +515,12 @@ const getColumnIdentifier = (col: SelectedColumn): string => {
...
 
 const getTableIdentifier = (database: string, table: string): string => {
-  const sep = (database === '' || table === '') ? '' : '.';
+  const sep = (!database || !table) ? '' : '.';
   return `${escapeIdentifier(database)}${sep}${escapeIdentifier(table)}`;
 }
 
 const escapeIdentifier = (id: string): string => {
-  return id === '' ? '' : `"${id}"`;
+  return id ? `"${id}"` : '';
```
mshustov, line 523: let's add a few test as well. can be done in follow-up

asimpson (dismissed):
Seems to work well, thanks! :trophy:

asimpson (approved):

PR #734, opened 2024-02-28
fix: re-add query type selector to SQL Editor queries in dashboard view

### LINKED ISSUE #730: breaking dashboards after version upgrade 3=>4
**What happened**:

All dashboards break after upgrading the plugin from 3 => 4

Everything should stay the same or improve

I was able to break it down to the missing selector `Format` in QueryBuilder. 

In version 3 i add a new graph as timeseries and define this query
```sql
select toStartOfMinute(dt) as dt,
       count(*),
       errorcode    as code

from default.test

where (dt >= $__fromTime AND dt <= $__toTime)

group by dt, code
order by dt asc
```
In Format i select "Time Series". The graph is correctly showing a timeseries including my group-by data.

In version 4 the Format selector is missing. The time series graph is always ignoring the group-by data and only showing a wired graph with count() information.

I did check the JSON model of the dashboard and i see that it was marked as ` "queryType": "table"` .  

If i change manually the JSON model to `timeseries` it works again.

How can I make a timeseries graph with group-by work again in plugin version 4?

Version 4:
Screenshot 2024-02-24 at 09 13 40

Version 3:
Screenshot 2024-02-24 at 09 12 29

**Environment**:

- Grafana version: OSS version and Cloud behave the same
- Plugin version: 3-4
- OS Grafana is installed on:
- User OS & Browser: Firefox / Chrome Mac OSx
- Others:

SpencerTorres:
Fixes #730

Migration script works fine, but for some reason the dashboard still requires a `format` to be provided even though the data is the same.

This re-enables the `queryType`/`format` selector on dashboard view. Previously it was only enabled on explore view, but now it is visible on dashboard view again.

query type selector

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Added `$__fromTime_ms` macro that represents the dashboard "from" time in milliseconds using a `DateTime64(3)`
 - Added `$__toTime_ms` macro that represents the dashboard "to" time in milliseconds using a `DateTime64(3)`
 - Added `$__timeFilter_ms` macro that uses `DateTime64(3)` for millisecond precision time filtering
+- Re-added query type selector in dashboard view. This was only visible in explore view, but somehow it affects dashboard view, and so it has been re-added. (#730)
 
 ### Fixes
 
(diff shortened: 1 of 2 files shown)


asimpson (dismissed):

asimpson (approved):
:white_check_mark:

PR #496, opened 2023-08-21
Add Docker image builds and post `docker pull` commands for local testing

adamyeats:
Related to https://github.com/grafana/grafana/issues/73287.

This PR gives us the ability to just spin up a Docker image to test any changes made in a local instance of Grafana. When the CI pipeline is complete, the Github Actions bot will publish a helpful PR comment so that the reviewer can just copy/paste the command into their terminal.

The image is tagged with the long SHA of the most recent commit. At least in the short-term, all plugins will push to the same Docker Hub repo, so we use the long SHA to minimise the possibility of collision (it's also much easier to get that from the build context than the short SHA).

Once this is merged, let's keep this localised to this repo for a week or so, so we can test it fully before rolling it out to other repos.

~**⚠️ NOTE:** At present, you will need to be logged into your Grafana Docker Hub account in order to pull the image. A ticket is open with IT to make the repo public.~ Fixed!

PR #737, opened 2024-02-29
Address docker images for PRs failing to build for external PRs

asimpson:
Currently we build a Docker image and push it to Docker Hub for every PR (#496). This works great. However, this doesn't work for PRs from external forks ([example failed run](https://github.com/grafana/clickhouse-datasource/actions/runs/8075629913/job/22062884089?pr=734)) since secrets are not passed to Github Actions in that context.

There are two things to fix:

1. Prevent PRs from public forks from trying to build and push the image.

2. Come up with a workaround for PRs from external forks so that someone reviewing theh PR can easily spin up a Docker image.

The first point is easy enough to fix by only attempting to push to Dockerhub if the PR _is not_ from a fork.

The second is more difficult and ramps up the complexity. 

I previously attempted to migrate to Github Container Registry since that uses the `GITHUB_TOKEN` which is the only secret that is available in PRs from external forks. However that case only allows [`read` access](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token).

The solution I've settled on is to use [`docker save`](https://docs.docker.com/reference/cli/docker/image/save/) to generate a tarball of the image and then persist that as an artifact. The steps to use that artifact locally are then:

1. Download artifact and unzip it (Github Actions automatically zips all artifacts, see [issue](https://github.com/actions/upload-artifact/issues/39))
2. Use `docker load` to load the tarfile as a Docker image, e.g. `docker load < /path/to/image.tar`
3. `docker run -p 3000:3000 imagename`


aangelisc (commented):
LGTM, a couple questions:

- Is it possible to remove the condition on the build step and then only add a condition for the push? Can build and push be separated? It'll probably make the workflow more readable.
- Shall we add a PR comment detailing that the image has been saved as an artifact and how to run it?

asimpson:
@aangelisc 

Good questions!

1. The step for building also pushes which would then fail for PRs from forks
2. Adding a comment also requires escalated privileges which aren't given for PRs from forks (see [this note](https://github.com/peter-evans/create-or-update-comment?tab=readme-ov-file#action-inputs)). I toyed around with adding a comment using `pull_request_target` and a separate workflow to do that commenting but it didn't seem worth the effort.

aangelisc (approved):
LGTM!

Issue #783, opened 2024-04-01
Error while fetching the data with column type "Array(Nullable(String))"


haudoing:
**What happened**:
The column type "Array(Nullable(String))" is not tested in either [pkg/converters/converters_test.go](https://github.com/grafana/clickhouse-datasource/blob/main/pkg/converters/converters_test.go) or [pkg/plugin/driver_test.go](https://github.com/grafana/clickhouse-datasource/blob/main/pkg/plugin/driver_test.go). I made one test case in file [pkg/plugin/driver_test.go](https://github.com/grafana/clickhouse-datasource/blob/main/pkg/plugin/driver_test.go) on my local then sometimes I found the error below on [line 280](https://github.com/grafana/clickhouse-datasource/blob/main/pkg/plugin/driver_test.go#L277).
```
sql: Scan error on column index 0, name "col1": unsupported Scan, storing driver.Value type []*string into type *string
```
The function I wrote looks like this
```
func TestArrayNullableString(t *testing.T) {
	for name, protocol := range Protocols {
		t.Run(fmt.Sprintf("using %s", name), func(t *testing.T) {
			var val []*string
			conn, close := setupTest(t, "col1 Array(Nullable(String))", protocol, nil)
			defer close(t)
			v := "48"
			val = append(val, &v, nil)
			insertData(t, conn, val)
			checkRows(t, conn, 1, val)
		})
	}
}
```

**What you expected to happen**:
Expecting no error while using the plugin with the column type "Array(Nullable(String))"

**How to reproduce it (as minimally and precisely as possible)**:

**Screenshots**

**Anything else we need to know?**:
This error does not always appear. 

**Environment**:

- Grafana version:
- Plugin version: v4.0.3
- OS Grafana is installed on:
- User OS & Browser: Linux mint 21.3
- Others:

haudoing:
The PR above fixes my problem and all the tests are running fine.

PR #785, opened 2024-04-01
These two regex won't match with the complex column type now.

haudoing:
#783


SpencerTorres:
Thanks for the issue + PR!

I see `Nullabe(IP)` was also changed, is this covered by a test case? I see theres some other ones in there but not specifically for array nullable.

SpencerTorres:
Just noticed there's IPv4 and IPv6, could you add one more test case for IPv6?

Other than that, the PR looks good. I'll run it locally to verify

SpencerTorres:
```go
"Nullable(IP)":              regexp.MustCompile(`Nullable\(IP`),
"Nullable(String)":          regexp.MustCompile(`Nullable\(String`),
```

It's very interesting that of all of the types there, these are the only two without the `^` prefix. I was worried this was intentional, but after seeing the original commits for [`String`](https://github.com/grafana/clickhouse-datasource/blob/0edd06bcbb2f57dbaf02dac9f3f70e04ab8334fa/pkg/converters/converters.go#L26) and [`IP`](https://github.com/grafana/clickhouse-datasource/blob/615978e7dce78e16a5a25da422be484950c1e1c4/pkg/converters/converters.go#L30), it seems this was just a small oversight.

I've run the tests and everything looks good. Thanks for the contribution 👍

bossinc:
Thank you @haudoing and @SpencerTorres !

bossinc (approved):

PR #790, opened 2024-04-05
Bump grafana-plugin-sdk-go to v0.220.0

wbrowne:
This is necessary as v0.217.0 includes a PDC bug.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/ClickHouse/clickhouse-go/v2 v2.21.1
 	github.com/docker/docker v25.0.4+incompatible
 	github.com/docker/go-units v0.5.0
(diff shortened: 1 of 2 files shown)


wbrowne:
@grafana/partner-datasources it looks like we have a release with the `v0.217.0` SDK version - can you confirm if I got that wrong or not? Since this plugin supports PDC so we should remove this release from the catalog in case customers end up running into this issue (we can make a release here, move customers off the latest version onto the new version and remove the problematic version from the catalog)

aangelisc:
Hey @wbrowne, that's accurate. I'm going to approve this and then we can look at releasing

aangelisc (approved):

wbrowne:
> Hey @wbrowne, that's accurate. I'm going to approve this and then we can look at releasing

Looks like I can't merge, so work away. Thank you! ❤️

PR #799, opened 2024-04-15
Make it possible to use PDC request values

wbrowne:
- Updates `sqlds` as this makes it possible to access request values via `context`
- Uses new SDK function to get request values rather than relying on environment variables https://github.com/grafana/grafana-plugin-sdk-go/pull/955


aangelisc (approved):
LGTM, just some lint related issues that need fixing before this could be merged

aangelisc (approved):
Thanks @wbrowne!

PR #804, opened 2024-04-18
Dashboards: Fix query in Data Analysis dashboard

### LINKED ISSUE #803: Dashboards: query failing in the "ClickHouse - Data Analysis" dashboard
**What happened**:
Dashboard "ClickHouse - Data Analysis" is not loading properly. Panel "Database summary" fails because it is unable to find data_uncompressed_bytes column.

All panels in the dashboard loading with no errors.

**How to reproduce it**:

1. Install this plugin.
2. Create a datasource that uses this plugin.
3. Go to the datasource config -> Dashboards and click on the Import button on "ClickHouse - Data Analysis".
4. Open the dashboard and notice the error.

**Screenshots**
image

- Grafana version: 10.2.3
- Plugin version: 4.0.6

oserde:
Similar to the issue fixed by #487, there is a missmatch in the "Database summary" panel, as it uses data_uncompressed_bytes in the outer query (this seems to be the wanted columns, as it is defined in the "Organize fields by name" transformation) but uses data_compressed_bytes in the inner query, causing an error.

This PR sets the right column in the inner query and fixes #803


aangelisc (approved):
LGTM, thank you for the contribution @oserde!

PR #814, opened 2024-04-29
Refactor `GetConverter` and handle `LowCardinality()` type

### LINKED ISSUE #805: Regression in `LowCardinality(Nullable(String))` type processing in v4.0.5
**What happened**:

Trying to select a column of aforementioned type returns an error:

```
sql: Scan error on column index 0, name "%column_name%": unsupported Scan, storing driver.Value type *string into type *string: Could not process SQL results

```

It seems to be a regression of [#785](https://github.com/grafana/clickhouse-datasource/pull/785), wherein the regexp for `"Nullable(String)"` Converter [has been changed](https://github.com/grafana/clickhouse-datasource/pull/785/commits/310eeaed1cafcecffc3cf088aea40d6338a01a07#diff-ee17992c987df01645f07024591cfd9212594ea65f8d7f7a06446d49523fa7d5L37) to match on the beginning of string only, thus not matching `LowCardinality(Nullable(String))` anymore (and in this case default Converter is used and that is probably the cause of the error).

I've downgraded datasource version to **v4.0.4** to corroborate that and the error indeed disappeared.

<br>

No errors.

<br>

In Explore view:

```
select
    toNullable('test')
;
```

OK

```
select
    toLowCardinality('test')
;
```

OK

```
select
    toLowCardinality(
        toNullable('test'))
;
```

Error:

```
sql: Scan error on column index 0, name "toLowCardinality(toNullable('test'))": unsupported Scan, storing driver.Value type *string into type *string: Could not process SQL results
```

<br>

**Environment**:

v4.0.6, grafana v10.4.1

adamyeats:
Fixes #805. This PR does a few things:

- Handles `LowCardinality()` data type correctly and adds tests
- Adds a converter for `String` to handle the recursion in `GetConverter` correctly (is there a better way to do this? i guess there was a reason why there wasn't a converter there for `String` already, but adding one does allow us to do some recursiveness here)
- Modifies `defaultConvert` to handle strings and add better handling for pointers and dereferencing them
- Refactors `GetConverter` to make it a bit nicer (?)
- Uses `reflect.PointerTo` instead of `reflect.PtrTo`, which is now deprecated

### DIFF
--- a/pkg/converters/converters.go
+++ b/pkg/converters/converters.go
@@ -2,6 +2,7 @@ package converters
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"math/big"
 	"net"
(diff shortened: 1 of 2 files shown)


aangelisc (approved):
LGTM! Great stuff @adamyeats 👍

PR #837, opened 2024-05-17
fix: isStringType for Nullable(String) and FixedString(N) (#830)

### LINKED ISSUE #830: IN operator not correctly constructed in Query Builder Filter
**What happened**:
When constructing a query using Query Builder, if using the 'IN' operator for a filter, the generated SQL is not correct

the generated SQL for XXX IN (1,2,3) would be expected to be
XXX IN (1,2,3)
Instead it produces 
XXX IN '(1,2,3)'

1. Go to Query Builder, Query Type: Time Series
2. Create a Filter using the 'IN' operator
3. View the generated SQL Preview

n/a

**Environment**:

- Grafana version:
- Plugin version: grafana-clickhouse-datasource v4.0.7
- OS Grafana is installed on:
- User OS & Browser:
- Others:

SpencerTorres:
Fixes #830 

The `IN` operator was broken because the column type was `Nullabe(String)` instead of `String`.
- Fixed issue
- added tests for `String`, `Nullable(String)`, and `FixedString`
- Updated changelog

### DIFF
--- a/src/data/sqlGenerator.ts
+++ b/src/data/sqlGenerator.ts
@@ -747,7 +747,7 @@ const numberTypes = ['int', 'float', 'decimal'];
 const isNumberType = (type: string): boolean => numberTypes.some(t => type?.toLowerCase().includes(t));
 const isDateType = (type: string): boolean => type?.toLowerCase().startsWith('date') || type?.toLowerCase().startsWith('nullable(date');
 // const isDateTimeType = (type: string): boolean => type?.toLowerCase().startsWith('datetime') || type?.toLowerCase().startsWith('nullable(datetime');
-const isStringType = (type: string): boolean => type.toLowerCase() === 'string' && !(isBooleanType(type) || isNumberType(type) || isDateType(type));
+const isStringType = (type: string): boolean => (type.toLowerCase() === 'string' || type.toLowerCase().startsWith('nullable(string') || type.toLowerCase().startsWith('fixedstring')) && !(isBooleanType(type) || isNumberType(type) || isDateType(type));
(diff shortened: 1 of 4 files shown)


src/data/sqlGenerator.ts
```diff
@@ -747,7 +747,7 @@ const numberTypes = ['int', 'float', 'decimal'];
 const isNumberType = (type: string): boolean => numberTypes.some(t => type?.toLowerCase().includes(t));
 const isDateType = (type: string): boolean => type?.toLowerCase().startsWith('date') || type?.toLowerCase().startsWith('nullable(date');
 // const isDateTimeType = (type: string): boolean => type?.toLowerCase().startsWith('datetime') || type?.toLowerCase().startsWith('nullable(datetime');
-const isStringType = (type: string): boolean => type.toLowerCase() === 'string' && !(isBooleanType(type) || isNumberType(type) || isDateType(type));
+const isStringType = (type: string): boolean => (type.toLowerCase() === 'string' || type.toLowerCase().startsWith('nullable(string') || type.toLowerCase().startsWith('fixedstring')) && !(isBooleanType(type) || isNumberType(type) || isDateType(type));
```
adamyeats, line 750: Would this have the same issue as #805? It doesn't look like this would work if the type is wrapped by `LowCardinality()`. Is that a concern in this case?

adamyeats (commented):
Thanks @SpencerTorres , just one comment from me, and also please rebase/merge this branch! 🙏

src/data/sqlGenerator.ts
SpencerTorres: Huge concern. Is an issue. Just fixed it though 😎 
Thanks!

adamyeats (approved):

adamyeats:
🚀

Issue #833, opened 2024-05-16
Clickhouse datasource plugin returns error in Grafana explore page when query with a LowCardinality(Nullable(String)) column


linjunzhe:
**What happened**: 
It returned an error if a table contains a LowCardinality(Nullable(String)) column. 
```
sql: Scan error on column index 0, name "src_tor_name": unsupported Scan, storing driver.Value type *string into type *string: Could not process SQL results
```

**What you expected to happen**:
It should return the data same as clickhouse-client. 

**How to reproduce it (as minimally and precisely as possible)**:
1. Create a table with a LowCardinality(Nullable(String)) column, for example 
```
CREATE TABLE pingmesh.test ON CLUSTER staging
(
    `src_tor_name` LowCardinality(Nullable(String)),
    `loss_count` Int64,
    `unixtime` DateTime('Etc/UTC') CODEC(DoubleDelta)
)
ENGINE = ReplicatedMergeTree('/clickhouse/{installation}/{cluster}/tables/{shard}/pingmesh/test', '{replica}')
PARTITION BY toYYYYMMDD(unixtime)
PRIMARY KEY unixtime
ORDER BY unixtime
SETTINGS index_granularity = 8192
```
2. Insert some data
```
INSERT INTO pingmesh.test (*) VALUES ('test', 0, 0)
```
3. Query the data from clickhouse client. It works fine.
```
> SELECT * FROM "pingmesh"."test" LIMIT 10
SELECT *
FROM pingmesh.test
LIMIT 10

Query id: fd6dea33-12b2-4ae2-aaa4-c2e8601742cf

┌─src_tor_name─┬─loss_count─┬────────────unixtime─┐
│ test         │          0 │ 1970-01-01 00:00:00 │
└──────────────┴────────────┴─────────────────────┘

1 row in set. Elapsed: 0.001 sec.
```
4. Put the same query in Grafana Explore with Clickhouse datasource. 
![image](https://github.com/grafana/clickhouse-datasource/assets/1384807/5fb8b903-b21f-448a-916d-89caaa19a8ba)

**Screenshots**
![image](https://github.com/grafana/clickhouse-datasource/assets/1384807/5fb8b903-b21f-448a-916d-89caaa19a8ba)
**Anything else we need to know?**:
1. I was upgrade Grafana recently to the newest 10.4.3. The previous version I used is 10.2.3, which was working fine. 
[...]

PR #857, opened 2024-06-07
Temporaily patch converters to fix support for LowCardinality strings

adamyeats:
This is a temporary fix for #833. At the moment, the `sqlutil` package in the Plugin SDK does not correctly handle the `LowCardinality` data types. I thought we'd fixed this in #814, but it turns out that the `GetConverter()` function is not actually called by anything within the SDK or out code as expected. 

Until we can patch this upstream, I've added two extra converters for now, that can be removed when we can introduce more generic handling of these types.


SpencerTorres:
I think `LowCardinality` is mainly used with strings so this should be fine for now. Maybe we can just strip this modifier before calling `GetConverter`? I don't think it affects the type scanning/reflection once it gets to the client, so if we strip it, it should be effectively the same as `String`.

adamyeats:
@SpencerTorres 

> Maybe we can just strip this modifier before calling GetConverter? I don't think it affects the type scanning/reflection once it gets to the client, so if we strip it, it should be effectively the same as String.

That is more or less how [the original fix](https://github.com/grafana/clickhouse-datasource/blob/main/pkg/converters/converters.go#L276) worked, however the `GetConverter()` function is not actually called by anything (maybe it's just a test helper function? not sure), so we needed to just add a new converter for the time being. As you mention above, we figured that as only strings are supported by `LowCardinality` out-of-the-box, that we'd be fine with just covering these cases for now.

Unfortunately, this is blocked on some failing tests in CI at the moment, but the test failures are intermittent and do not seem to occur locally, making this quite a tricky one to debug. We'll have more info ASAP, but for now I'm going to move this back to draft.

aangelisc (dismissed):

SpencerTorres (approved):
I have thoroughly poked through the plugin and sqlds code, and this is clearly the easiest fix for now. We have users who need `LowCardinality(Nullable(String))` so we need to get this merged ASAP.

In the future, I would like to see more flexibility with `Converter` type in sqlds. Ultimately what we need is a way to strip `LowCardinality(` + `)` from the type string BEFORE it goes to sqlds. This could possibly be done in `clickhouse-go`, or maybe by making a porxy interface for `clickhouse-go`'s SQL driver, but this is all so complicated. The easiest solution is to add these converters. Perhaps automatically.

It wouldn't hurt to dynamically build the array of converters and add a copy of each converter that is wrapped in `LowCardinality`. Low Cardinality does not affect clients, only how the data is stored on the server.

A temporary workaround is remove `LowCardinality` via `CAST()` in the user's SQL.

TL;DR: This is good, we need this fix ASAP. Not sure what CI issues are failing but let me know if I can help.

SpencerTorres (approved):
Looks good. Hopefully we can get a better sqlds API for fixing this, but this works for now. Most `LowCardinality` types are `String`.

pkg/converters/converters.go
```diff
@@ -253,10 +254,20 @@ var Converters = map[string]Converter{
 		scanType:   reflect.TypeOf((*interface{})(nil)).Elem(),
 	},
 	"Point": {
-		convert:    pointConverter,
-		fieldType:  data.FieldTypeNullableJSON,
-		matchRegex: matchRegexes["Point"],
```
aangelisc, line 258: I think this line has been unintentionally removed?

aangelisc (dismissed):
LGTM - just one nit in the comments!

aangelisc (approved):

PR #872, opened 2024-06-21
feature: add `$__dateTimeFilter()` macro

SpencerTorres:
## TL;DR

Adds a `$__dateTimeFilter(dateColumn, timeColumn)` that expands into:
```
WHERE
  (dateColumn >= toDate('2022-10-21') AND dateColumn <= toDate('2022-10-23'))
  AND
  (timeColumn >= toDateTime(1415792726) AND timeColumn <= toDateTime(1447328726))
```

## Why?

A common pattern in ClickHouse is having an `ORDER BY` that looks like this:
```sql
PARTITION BY toYYYYMM(TimestampDate)
ORDER BY (TimestampDate, TimestampTime)
```

Where  `TimestampDate` is a `Date` and `TimestampTime` is a `DateTime`.

When querying this, it is common to do `$__dateFilter(TimestampDate) AND $__timeFilter(TimestampTime)`. This macro makes it shorter to type, by being able to do `$__dateTimeFilter(TimestampDate, TimestampTime)`.

With this macro it is faster and more efficient to query tables that have this pattern of partitioning, since it references both time columns.


asimpson (approved):
LGTM, thanks! :rocket:

PR #873, opened 2024-06-21
cleanup dependencies

SpencerTorres:
Changes:
- Added dependencies for `semver` and `@grafana/schema`, which were used but not added as direct dependencies
- Removed `@grafana/experimental` dependency since the repo is being deprecated 
- Copied over components + tests that were used from `@grafana/experimental`


asimpson (approved):
:rocket: nice!

Issue #900, opened 2024-07-10
Trace Detail no Longer Works


forestsword:
Hello, after the recent update the trace detail has stopped working. I'm using the latest grafana and this plugin.

After I click on a trace to show it from the trace id list nothing is returned. The multiply function added in https://github.com/grafana/clickhouse-datasource/pull/890 I think was not fully implemented, the [`trace_start` and `trace_end` ](https://github.com/grafana/clickhouse-datasource/blob/13014f723e4ba27cb1566f06d21e7ce7a2b524d2/src/data/sqlGenerator.ts#L158-L159) also need the function added in order for the where comparison to work. Manually adding them to the generated SQL I get the trace returned.

SpencerTorres:
Looking into this, thanks for the info

aangelisc:
Fixed by #890 😊

forestsword:
Thanks! Could we have a patch release of the plugin?

NicolasBissig:
@SpencerTorres @aangelisc When is this planned to be released?

PR #890, opened 2024-07-03
Update trace queries to better handle time fields

aangelisc:
Traces are currently displayed with the wrong timing due to discrepancies between how the trace visualisation handles times and how the Javascript `Date` type functions (it doesn't correctly handle sub-millisecond precision).

Fixes #888 by converting the time field to nanoseconds and then adjusting it to milliseconds.

PR #904, opened 2024-07-11
Fix trace_start and trace_end changes from #890

SpencerTorres:
fixes (#900)

changes:
- Reverted Trace Search back to original timestamp, since that is intended to be human readable
- Fixed Trace ID mode to use the original DateTime type instead of the float type, which fixes the time range filtering
- Updated changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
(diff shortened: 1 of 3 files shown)


aangelisc (approved):
Thank you @SpencerTorres! I've tested locally and this all LGTM.

image

rudebono:
Thank you so much ! ⸜(｡˃ ᵕ ˂ )⸝♡

PR #923, opened 2024-07-26
Code-gen dashboard from `system.dashboards` table

SpencerTorres:
Fixes #886 

ClickHouse has built-in dashboards defined in the system table [`dashboards`](https://clickhouse.com/docs/en/operations/system-tables/dashboards).

This can be seen natively with the ClickHouse HTTP server: https://clickhouse.com/docs/en/operations/monitoring#built-in-advanced-observability-dashboard

This change includes a script that fetches the queries from that table, transforms them for Grafana, and then puts them into a dashboard that can be loaded in the plugin.

generated dashboard

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Added preset dashboard from `system.dashboards` table
+
 ### Fixes
 
 - Fix trace start times in trace ID mode (#900)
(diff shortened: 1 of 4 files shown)


asimpson (approved):
Looks and works great :trophy:

SpencerTorres:
Thanks! Hopefully we can improve it in the future with more specialized units/colors, but this matches the built-in dashboard almost exactly

PR #933, opened 2024-08-08
Skip unnecessary json.Unmarshal in converter

SpencerTorres:
An optimization when loading object/array-like types dependent on `jsonConverter`. We can skip this marshal+unmarshal combo and simply marshal it and cast the result to the required `json.RawMessage`.

Looks like this was added to convert `[]byte` to `json.RawMessage`, but both types are already `[]byte` so we can simply cast it.

Changes:
- optimized `jsonConverter`
- updated changelog


aangelisc (approved):

aangelisc:
@SpencerTorres this just needs main merged and then it's good to go 😊

aangelisc:
Thanks @SpencerTorres!

PR #939, opened 2024-08-13
Add warning for missing `uid` in provisioned datasources

SpencerTorres:
Some users reported confusion with data linking not working between logs/traces.
This is because a `uid` must be set for the datasource. All UI-configured datasources generate one automatically, but not all provisioned datasources have one. I have added a warning to the config page to remind users to add one.

Changes:
- added warning banner when `uid` is not set
- updated changelog

Example of what this warning looks like:
example warning

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixes
+
+- Added warning for when `uid` is missing in provisioned datasources.
+
 ## 4.3.2
 
 ### Fixes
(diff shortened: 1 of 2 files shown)


aangelisc (dismissed):

aangelisc:
Thank you @SpencerTorres! Has this been reported internally or is there an open issue related to this?

SpencerTorres:
This was something found internally with one of our users. There's no alternative fix so it seemed best to simply add a warning. I believe this would affect all datasources, perhaps a warning could be added at the Grafana level at some point

aangelisc (approved):
Awesome, thank you @SpencerTorres!

PR #942, opened 2024-08-14
fix: map key filtering from showing column name

SpencerTorres:
This fixes a regression from #862 that caused the column name to show up in the map key filter dropdown instead of showing the actual key being filtered. The filter still works but it's a confusing visual bug

changes:
- Map key now correctly shows in filter
- Updated changelog

broken:
before fix

fixed:
after fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


aangelisc (approved):
LGTM, thanks @SpencerTorres!

Issue #187, opened 2022-09-01
Better error reporting on timeout


gingerwizard:
Currently, if the query deadline is exceeded (as controlled by QueryTimeout) in 2.0, the go driver returns a `error querying the database: context deadline exceeded` error. This isn't particularly clear and basically means the user needs to increase the query timeout (default 60s).

We should catch this specific error and wrap with something to indicate as such.

![image](https://user-images.githubusercontent.com/12695796/187908205-6f8ddeb8-4ac7-44c1-9699-a2c4727330c6.png)

gingerwizard:
note QueryTimeout is in 2.0 only - see https://github.com/grafana/clickhouse-datasource/pull/186

previous versions hardcoded at 30s.

gingerwizard:
This is a client issue. Ill wrap this error in the go-client and PR the plugin when we release the client.

aangelisc:
@gingerwizard are you able to create a PR for this? Was it fixed?

alyssabull:
Hey @gingerwizard - confirming if this issue was fixed and if so, is there a corresponding PR that we can link this issue to?

SpencerTorres:
@jkaflik yes, I can catch this error and add a clearer message 👍

dheerajathrey:
@SpencerTorres @jkaflik Was this issue fixed? If not any plan of picking this up?

aangelisc:
Hi @SpencerTorres,

I'm not sure about the fix for this one. The context timeout could be reached due to a number of reasons. Is there any way of us identifying from the ClickHouse driver if a query is long-running which could lead to this being hit.

I'm concerned that we could wrap this and mislead users into thinking that increasing the timeout may solve their issue when really the timeout could be for a number of reasons.

adamyeats:
Fixed in #1776. Please re-open if the issue still persists.

PR #951, opened 2024-08-19
Add error source

aangelisc:
Add errorsource to any errors that are downstream and may be returned at some point in the QueryData endpoint. A downstream error is anything that is not due to the plugin directly.

The majority of this is handled by `sqlds` directly but I've tried to account for any cases where it isn't handled.


adamyeats (approved):

SpencerTorres (approved):
Awesome! Hope this improves error messages on the query and config pages.

SpencerTorres:
I wonder if this solves or improves https://github.com/grafana/clickhouse-datasource/issues/187

bossinc (approved):

aangelisc:
Hey @SpencerTorres, thanks for the review. For now I don't think it'll fix that issue but we can look into that soon 😊

PR #954, opened 2024-08-19
Update/fix missing `system.dashboards` dashboard

SpencerTorres:
This dashboard was added in #923 but not added to the list in `plugin.json`, so it never appeared in the UI. I have fixed this and fetched the latest dashboards from the latest version of ClickHouse.

Changes:
- Fetched latest system dashboards from ClickHouse
- Added dashboard to `plugin.json` so that it appears in the UI
- Added newline to dashboard code gen
- Updated changelog


aangelisc (approved):
Thank you @SpencerTorres!

Issue #262, opened 2022-12-14
Allow empty Inputs for $__conditionalAll


fionera:
**Describe the bug**
As a user I would expect an empty Input field to be a selection for all

**Versions**
Grafana: v9.2.0
Plugin: v2.0.5

**Query**
-

**To Reproduce**
-

**Expected behavior**
Currently an empty Input triggers the $__conditionalAll and renders the Value

**Screenshots**
-

**Additional context**
-

fionera:
Additionally it seems that Variables inside the $__conditionalAll are not evaluated

fionera:
Ok the evaluation works but it needs parentheses around it

PR #960, opened 2024-08-23
Allow empty string values for `$__conditionAll` macro

### LINKED ISSUE #947: [Bug?] Macro $__conditionalAll sets `1=1` in empty variable values
**What happened**:
We have a Grafana selector that allows empty string values (''), which are valid in our context. However, when using the macro $__conditionAll, it is being replaced with 1=1, even though an empty string should be considered a valid selection.

For reference, the current documentation of the macro is as follows (README.md):

| Macro                                        | Description                                                                                                                                                                         | Output example                                                                                        |
|----------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------|
| *$__conditionalAll(condition, $templateVar)* | Replaced by the first parameter when the template variable in the second parameter does not select every value. Replaced by the 1=1 when the template variable selects every value. | `condition` or `1=1`                                                                                  |

If the Grafana selector has an empty string value `''`, the macro `$__conditionAll` should not be replaced, as this value can be supported and is valid for applying the specific condition.

Code:
https://github.com/grafana/clickhouse-datasource/blob/e4fce05dd18cc64112202155aa230c7b5111b21a/src/data/CHDatasource.ts#L268

**Questions**: 

- Is there a specific reason why an empty string ('') is treated as all?

**Environment**:

- Grafana version: 11.1.3
- Plugin version: 4.0.8
- OS Grafana is installed on: - 
- User OS & Browser: -
- Others: -

SpencerTorres:
Fixes #947 but may also break #262.

It makes sense to allow empty strings for this macro, anyone who wishes to exclude the macro can use `$__all` in their query to get the `1=1` functionality.

Changes:
- Empty string will no longer replace the macro with `1=1`
- Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Added warning for when `uid` is missing in provisioned datasources.
 - Map filters in the query builder now correctly show the key instead of the column name
+- Allow empty string values for `$__conditionAll` macro (#947)
 
 ## 4.3.2
 
--- a/src/data/CHDatasource.ts
+++ b/src/data/CHDatasource.ts
@@ -265,7 +265,7 @@ export class Datasource
       if (templateVar) {
         const key = templateVars.find((x) => x.name === templateVar[0]) as any;
         let value = key?.current.value.toString();
-        if (value === '' || value === '$__all') {
+        if (value === '$__all') {
           phrase = '1=1';
         }
       }


aangelisc (commented):
This LGTM also, are we expecting that this will be a potential breaking change?

sbengo:
Hi @SpencerTorres , is there any chance the PR will be merged and available in the upcoming releases?

SpencerTorres:
@sbengo I agree we need to be careful with changing the behavior of this macro, I think we'll follow the suggestion of making a new one for the new behavior

PR #966, opened 2024-08-28
opentelemetry-clickhouse.json: fix "Duration" in "Traces" list

jberryman:
formerly this was the sum of all (nested) spans which isn't a meaningful value. `max` is also potentially not quite right, as it assumes one top-level containing span which I'm not sure is a valid assumption

### DIFF
--- a/src/dashboards/opentelemetry-clickhouse.json
+++ b/src/dashboards/opentelemetry-clickhouse.json
@@ -670,7 +670,7 @@
           },
           "pluginVersion": "4.0.6",
           "queryType": "table",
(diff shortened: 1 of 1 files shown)


jberryman:
actually the whole traces section is pretty bad and should be redone with some care imo

SpencerTorres:
> actually the whole traces section is pretty bad and should be redone with some care imo

Do we still want to merge this or should we open an issue to collaborate on an improved version?

With pre-made dashboards it can be hard to properly cover all use cases, these are more of a foundation for users to expand on to fit their specific needs.

jberryman:
sorry my last comment wasn't very constructive. I will try to post a list of other things that can be changed or PRs if I have space, but I'm not sure if I will. I think merging this would be an improvement if you have time

it was definitely helpful having this as a starting point

SpencerTorres:
All good, let's get a changelog entry and sign the CLA and we can merge. We can open a discussion/issue to talk more improvements. Thanks!

SpencerTorres (approved):

aangelisc:
Hi @jberryman, thank you for the contribution! Are you able to sign the CLA and add a CHANGELOG entry for this PR? 😊

SpencerTorres:
Closing in favor of #983 which has merged

PR #983, opened 2024-09-12
Fix: Update Duration value in example traces dashboard

aangelisc:
Similar to #966 the duration displayed will now be the maximum value. I'm also explicitly using the `divide` function for readability.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Added warning for when `uid` is missing in provisioned datasources.
 - Map filters in the query builder now correctly show the key instead of the column name
 - Updated and fixed missing `system.dashboards` dashboard in list of dashboards
+- Updated the duration value in example traces dashboard to provide useful information
 
 ## 4.3.2
 
--- a/src/dashboards/opentelemetry-clickhouse.json
+++ b/src/dashboards/opentelemetry-clickhouse.json
@@ -670,8 +670,7 @@
           },
           "pluginVersion": "4.0.6",
           "queryType": "table",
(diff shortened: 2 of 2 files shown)


alyssabull (dismissed):

alyssabull (approved):

alyssabull (approved):

SpencerTorres:
Just a note, division may not be the best operator here:

Issue: https://github.com/grafana/clickhouse-datasource/issues/720
Related code: https://github.com/ClickHouse/clickhouse-grafana-datasource/blob/280af29668f9977e9d4e779c033c4a297589905c/src/data/sqlGenerator.ts#L561-L575

aangelisc:
> Just a note, division may not be the best operator here:
> 
> Issue: #720 Related code: https://github.com/ClickHouse/clickhouse-grafana-datasource/blob/280af29668f9977e9d4e779c033c4a297589905c/src/data/sqlGenerator.ts#L561-L575

That's interesting @SpencerTorres, thank you for picking up on that. Would you suggest truncating the data to millisecond precision instead?

SpencerTorres:
I suppose it depends on how it's displayed, I want to avoid rounding to 0 and having that affect filtering

aangelisc:
I see, my understanding of the ClickHouse [documentation](https://clickhouse.com/docs/en/sql-reference/functions/arithmetic-functions#divide) is that `divide` should always yield a float value (and I've tested this on ClickHouse play with an integer value for the duration). I believe this should be fine but we can switch to `multiply` instead if you'd prefer to be cautious. Let me know what you think!

image

SpencerTorres:
I see the problem, I was previously using ` intDivOrZero`, which is much different...

Either one works then, should be fine as long as it preserves the float value 👍 

Thanks!

Issue #883, opened 2024-06-28
Ad hoc filter values ​​are incorrectly escaped when used with the IN operator


Kenterfie:
**What happened**:
We switched some weeks ago from the Altinity plugin for clickHouse to the official grafana clickhouse source plugin and we noticed that something was not working anymore as expected. We use extensively the in operator in our ad-hoc filters, which works fine with the previous plugin, but now the values are incorrectly escaped.

`settings additional_table_filters={'table' : ' IPv4NumToStringClassC(ipv4Dst) IN \'('1.1.1.xxx','2.2.2.xxx')\' '}`

**What you expected to happen**:
`settings additional_table_filters={'table' : ' IPv4NumToStringClassC(ipv4Dst) IN (\'1.1.1.xxx\',\'2.2.2.xxx\')'}`

**Anything else we need to know?**:

**Environment**:

- Grafana version: v10.4.1 
- Plugin version: 4.1.0

SpencerTorres:
Could you add some more detail about where the adhoc filter is getting its values? Also what the query looks like? Is it a SQL editor query, or query builder?

bossinc:
@Kenterfie I'm closing this issue. If it is still an issue, please reopen it with answers to Spencer's questions. Thank you!

PR #986, opened 2024-09-12
Adds support for IN operator as adhoc filter

Kenterfie:
This merge request adds a missing feature, which we have actively used in the past with the Altinity Grafana datasource. In general it allows adhoc filters which make use of the IN operator.

This issue is also addressed in https://github.com/grafana/clickhouse-datasource/issues/883

Example adhoc filter
```
key = ipv4Dst
operator = IN
value = ('1.1.1.xxx','2.2.2.xxx')
```

Expected additional filter
`settings additional_table_filters={'table' : ' ipv4Dst IN (\'1.1.1.xxx\',\'2.2.2.xxx\')'}`

Current additional filter output
`settings additional_table_filters={'table' : ' ipv4Dst IN \'('1.1.1.xxx','2.2.2.xxx')\' '}`


SpencerTorres:
Thanks for the contribution! Make sure to sign the CLA and add a CHANGELOG entry. Could you also edit your PR description to include a screenshot or example adhoc value for the `IN` operator? Is it comma separated or something else?

src/data/adHocFilter.test.ts
```diff
@@ -152,6 +152,28 @@ describe('AdHocManager', () => {
     );
   });
 
+  it('apply ad hoc filter IN operator with string values', () => {
+    const ahm = new AdHocFilter();
+    ahm.setTargetTableFromQuery('SELECT * FROM foo');
+    const val = ahm.apply('SELECT stuff FROM foo WHERE col = test', [
+      { key: 'key', operator: 'IN', value: '(\'val1\', \'val2\')' },
```
SpencerTorres, line 159: If we're adding a new `IN` operator, perhaps `value` could be more tailored to this use case.

For example, instead of having the user write out this complex escaped string `('val1', 'val2')`, it could instead just be comma separated `val1, val2` and then we format it in the code.

This would be similar to how we handle `IN` filters in the query builder:
- https://github.com/grafana/clickhouse-datasource/blob/7b50309aefba12ef9f2b980e2f9f255001b98db0/src/components/queryBuilder/FilterEditor.tsx#L93
- https://github.com/grafana/clickhouse-datasource/blob/7b50309aefba12ef9f2b980e2f9f255001b98db0/src/data/sqlGenerator.ts#L736-L738

src/data/adHocFilter.ts
```diff
@@ -58,6 +58,14 @@ function isValid(filter: AdHocVariableFilter): boolean {
   return filter.key !== undefined && filter.operator !== undefined && filter.value !== undefined;
 }
 
+function escapeValueBasedOnOperator(s: string, operator: AdHocVariableFilterOperator): string {
+  if (operator === 'IN') {
+    return `${s}`.replace(/'/g, "\\'");
```
SpencerTorres, line 63: `s` is already `string`, why is it wrapped as
```javascript
`${s}`
```

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Added "Labels" column selector to the log query builder
 - Datasource OTel configuration will now set default table names for logs and traces.
+- Added support for IN operator in adhoc filters
```
SpencerTorres, line 9: Thanks!

SpencerTorres (changes requested):
Few notes/questions

SpencerTorres (approved):

Kenterfie:
Thanks for your fast reaction on that @aangelisc, but one question. I can find the change in the CHANGELOG.md, but can't find it in release 4.4.0? What happened to it?

Kenterfie:
Ok, grafana broke it with its newer versions and now they not support anymore "custom" operators, which makes it nearly impossible to use it properly with databases.

PR #1002, opened 2024-09-26
Fix AST parsing when switching back to query builder

SpencerTorres:
The AST module we use for parsing SQL is still not fully compatible with ClickHouse SQL, but there are things we can fix.

The button to switch from "SQL" to "Query Builder" has not been working as expected the past few releases, this aims to fix and enhance its capabilities.

old button (dramatized):
broken query builder button

new behavior (query is actually somewhat parsed correctly, columns are put into their associated boxes):
slightly better query builder button

I have added some logic to help properly map log/trace column names to their correct boxes based on the SQL column names/aliases.

Changes:
- Switching from SQL editor to Query Builder is statistically less likely to destroy your query
- SQL columns will be matched (best effort) to the correlated column selection box. This uses the default column mappings from the datasource config, including OTel if enabled.
- Added tests
- Changelog

Other notes:
- The AST does not like backticks for column names
- The AST will refuse to parse map keys with bracket syntax (`ResourceAttributes['host']`)
- You must use quotes for your column name to preserve the case (`"Timestamp"`), otherwise it will be forced to lowercase. This behavior comes from the postgres AST module.
- The `ORDER BY` should also used a quoted identifier, else it will try to use the lowercased identifier and likely throw an error when executing the query.
- I have tried parsing ClickHouse's native `EXPLAIN AST`, but this does not include the required information to rebuild the query. Perhaps a hybrid approach would work?


aangelisc (approved):
Amazing work on this @SpencerTorres, especially with the testing! Thank you!

PR #1041, opened 2024-11-01
Fix failing bench github action

jalevin:
* bump bench version
* remove default auth variables
* remove comments
* fix playwright setup command
* run bench on PR and Push

### DIFF
--- a/.github/workflows/grafana-bench.yml
+++ b/.github/workflows/grafana-bench.yml
@@ -3,6 +3,8 @@ on:
   push:
     # Only run on push to the main branch
     branches: [main, master]
+  pull_request:
+    branches: [main, master]
 
 jobs:
   test:
--- a/.github/workflows/grafana-bench.yml
+++ b/.github/workflows/grafana-bench.yml
@@ -39,30 +39,21 @@ jobs:
         with:
           compose-file: './docker-compose.yml'
 
-      - name: Wait for Grafana to start
+      - name: Ensure Grafana is running
         run: |
           curl http://localhost:3000
-          #RESPONSE=$(curl -s -o /dev/null -w "%{http_code}" http://localhost:3000)
-          #echo $response
-          #if [ $RESPONSE -ne 200 ]; then
-          #  sleep 5;
-          #  echo "Grafana is not accessible"
-          #  exit 1
-          #fi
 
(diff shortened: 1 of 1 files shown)


.github/workflows/grafana-bench.yml
```diff
@@ -3,6 +3,8 @@ on:
   push:
     # Only run on push to the main branch
     branches: [main, master]
+  pull_request:
```
adamyeats, line 6: I'm not sure if I see this as something that needs to run on every PR, but I'm happy to be shown  otherwise! What are your thoughts?

.github/workflows/grafana-bench.yml
```diff
@@ -39,30 +41,21 @@ jobs:
         with:
           compose-file: './docker-compose.yml'
 
-      - name: Wait for Grafana to start
+      - name: Ensure Grafana is running
         run: |
           curl http://localhost:3000
```
adamyeats, line 46:
```suggestion
      - name: Wait for Grafana to start
        run: |
          timeout 180 bash -c 'until curl -s -o /dev/null -w "%{http_code}" http://localhost:3000 | grep -q "200"; do sleep 1; done && echo "Grafana is up!"'
```

This would allow us to guard against Grafana not being fully ready to server requests before proceeding to the next step. Wdyt?

.github/workflows/grafana-bench.yml
jalevin: If we don't run on a PR then the only chance to debug it is on merge which is tedious, especially if I need a review. I want to make sure tests are passing as often as possible (these were committed broken) and give feedback early and often.

.github/workflows/grafana-bench.yml
jalevin: It's fine for now- honestly I think this is or should be built into bench. I need to review our docs and we can remove this step entirely

.github/workflows/grafana-bench.yml
SpencerTorres: There's not too many PRs on this repo, I agree it would be good to validate early. Perhaps conditionally, as long as the simpler unit tests pass

adamyeats:
Thanks @jalevin ! 🚀

adamyeats (approved):

Issue #977, opened 2024-09-10
references (span links) not showing for traces in Grafana


Sup3rGeo:
Hi,

When we query traces from Clickhouse as a data source we don't see the span references listed, which is different from when I check traces from Tempo, where the reference links are in fact listed in the trace view.

I was able to somewhat fix it by adding another transformation to the sql query:

`arrayMap(x,y -> map('spanID', x, 'traceID', y), Links.SpanId, Links.TraceId) AS references`

This makes the links appear, but we are missing link attributes which are useful to identify/name each of the links. 

Unfortunately trying to add the Links.Attributes column as tags (as it seems to be the case with Tempo) wouldn't work:

`arrayMap(x,y,z -> map('spanID', x, 'traceID', y, 'tags', z), Links.SpanId, Links.TraceId, Links.Attributes) AS references`

the query fails because clickhouse requires all map values to be of the same type, and in this case we have (String, String, Map).

Is there a way to have Links.Attributes showing up as span references in the Trace view in Grafana?
Thanks!

**Environment**:

- Grafana version: 11.2.0
- Plugin version: 4.3.2

tatchi:
> the query fails because clickhouse requires all map values to be of the same type, and in this case we have (String, String, Map).


I'm hitting the same issue with trace events: https://github.com/grafana/clickhouse-datasource/discussions/840#discussioncomment-10706180. I wonder what we should do in this case ?

SpencerTorres:
I've experienced this error before with the array types. We have some types that may be able to handle this, but I'm not sure how they behave in an array. Perhaps some form of `Variant`, `Dynamic`, or `JSON` would work. For now you may be able to cast the map to a string by serializing it as JSON via [toJSONString](https://clickhouse.com/docs/en/sql-reference/functions/json-functions#tojsonstring). It may not look pretty in the UI, but it wouldn't break.

If we find a nice way to `SELECT` this we can add it to the default trace UI

tatchi:
Thanks for your input. It doesn't beak the UI but `attributes` aren't shown. I believe because it expects an `array` while it is a `string`

![Image](https://github.com/user-attachments/assets/f92f0d0d-6f2b-47b8-af8d-de7100766d59)

tatchi:
I found a solution for `events`, see https://github.com/grafana/clickhouse-datasource/discussions/840#discussioncomment-10799129. I guess we can apply the same solution for `Links.Attributes` ?

[...]

PR #1043, opened 2024-11-01
Add support for kind, status, instrumentation library, links, events and state data for traces

benceszikora:
This PR adds support for the kind, status, instrumentation library, links, events and state fields in the trace panel in Grafana. This will bring the plugin to feature parity with what Tempo can offer today.


src/otel.ts
```diff
@@ -49,6 +49,13 @@ const otel129: OtelVersion = {
     [ColumnHint.TraceTags, 'SpanAttributes'],
     [ColumnHint.TraceServiceTags, 'ResourceAttributes'],
     [ColumnHint.TraceStatusCode, 'StatusCode'],
+    [ColumnHint.TraceKind, 'SpanKind'],
+    [ColumnHint.TraceStatusMessage, 'StatusMessage'],
+    [ColumnHint.TraceInstrumentationLibraryName, 'InstrumentationLibraryName'],
```
SpencerTorres, line 54: I noticed theres a few columns defined in `otel.ts` that don't exist on the [ClickHouse OTel exporter default schema](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/eac16d39fcf5d201894b9babae3eff79dce4e58c/exporter/clickhouseexporter/example/default_ddl/traces.sql). The intent of these definitions is to match as closely as possible to this table schema for a better out of the box experience.

Should we look into adding these columns to the otel exporter?

Also note that `State` is defined too, but on the table it is called `TraceState`.

SpencerTorres (changes requested):
Thank you for the contribution! Glad to see expanded functionality for the trace columns. I've added a note regarding the intended table schema

Sup3rGeo:
Also AFAIU In the default schema Links is a Nested table and you need to select the fields individually:

```
 arrayMap(
  (traceid, spanid, attributes) -> tuple(
                     traceid,
                     spanid,
                     arrayMap(
                       key -> map('key', key, 'value', attributes[key]),
                       mapKeys(attributes)
                     )
                   )::Tuple(traceID String, spanID String, tags Array(Map(String, String)))
                   , Links.TraceId, Links.SpanId, Links.Attributes
  ) AS references
```

jwhitaker-gridcog:
> Also AFAIU In the default schema Links is a Nested table and you need to select the fields individually

@Sup3rGeo not disputing the table schema, but `arrayMap(l -> ..., Links)` does work here - is it possible this is a newer clickhouse feature?

@SpencerTorres  @benceszikora  is it worth splitting the bits of this PR that can work with the current schema into something that's mergeable now? In my case I'm itching for span events to be supported, and this achieves that out of the box which is really nice.

SpencerTorres:
> is it worth splitting the bits of this PR that can work with the current schema into something that's mergeable now?

Yes, there's some parts of this that need changing. I don't mind picking apart the code myself for it, but I want to make sure the author is okay with that (@benceszikora)

src/data/sqlGenerator.ts
```diff
@@ -149,9 +149,44 @@ const generateTraceIdQuery = (options: QueryBuilderOptions): string => {
   if (traceStatusCode !== undefined) {
     selectParts.push(`if(${escapeIdentifier(traceStatusCode.name)} IN ('Error', 'STATUS_CODE_ERROR'), 2, 0) as statusCode`);
   }
+
   const traceEventsPrefix = getColumnByHint(options, ColumnHint.TraceEventsPrefix);
   if (traceEventsPrefix !== undefined) {
-    selectParts.push(`arrayMap((name, timestamp, attributes) -> tuple(name, toString(toUnixTimestamp64Milli(timestamp)), arrayMap( key -> map('key', key, 'value', attributes[key]), mapKeys(attributes)))::Tuple(name String, timestamp String, fields Array(Map(String, String))),${escapeIdentifier(traceEventsPrefix.name)}.Name, ${escapeIdentifier(traceEventsPrefix.name)}.Timestamp, ${escapeIdentifier(traceEventsPrefix.name)}.Attributes) AS logs`);
```
harryjph, line 154: There are two ways in ClickHouse to define a Nested Table, controlled by the [`flatten_nested`](https://clickhouse.com/docs/en/operations/settings/settings#flatten-nested) setting at column creation time.

The query as previously written does not work when `flatten_nested=0`, as ClickHouse does not recognise the sub-columns such as `Events.Names` in this mode.

The rewritten query should work with either mode, as ClickHouse will recognise `Events` as one column even when it has been flattened.

src/components/queryBuilder/views/TraceQueryBuilder.tsx
```diff
@@ -102,6 +124,18 @@ export const TraceQueryBuilder = (props: TraceQueryBuilderProps) => {
     }));
   }, builderState);
 
+  // A function to assemble custom column definitions for the prefixed Events/Links columns
+  const onColumnPrefixChange = (name: keyof TraceQueryBuilderState, columnHint: ColumnHint) => {
```
harryjph, line 128: This change is necessary since at the moment, you cannot type into the "Events Prefix" field in the column selection form.

This function generates a `SelectedColumn` struct for the arbitrary input, which is necessary since when `flatten_nested=1`, the `ColumnSelect` component recognises the `Events` column as multiple sub-columns.

harryjph (commented):
Hey everyone! I've updated the code to be based on the [OTel Collector Schema](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/eac16d39fcf5d201894b9babae3eff79dce4e58c/exporter/clickhouseexporter/example/default_ddl/traces.sql) as requested, and rebased the PR onto main to incorporate https://github.com/grafana/clickhouse-datasource/pull/1128, which was merged last week. I've also updated the test cases to cover code merged in that PR.

I've left a couple of comments explaining the changes which aren't simple column addition.

benceszikora:
@SpencerTorres with @harryjph's changes this PR should now conform to the same schema as the OTel Collector. Are there any changes you would suggest before this can be merged?

src/views/CHConfigEditor.tsx
```diff
@@ -445,6 +445,13 @@ export const ConfigEditor: React.FC<ConfigEditorProps> = (props) => {
           onTagsColumnChange={c => onTracesConfigChange('tagsColumn', c)}
           onServiceTagsColumnChange={c => onTracesConfigChange('serviceTagsColumn', c)}
           onEventsColumnPrefixChange={c => onTracesConfigChange('eventsColumnPrefix', c)}
+          onLinksPrefixColumnChange={c => onTracesConfigChange('linksColumnPrefix', c)}
```
ValentinLvr, line 448: Typo here I guess
```suggestion
          onLinksColumnPrefixChange={c => onTracesConfigChange('linksColumnPrefix', c)}
```

ValentinLvr:
Just tested out this fork locally and note that span link redirection is not working so far.

When you click on the link it basically redirects you to the same trace instead of the one linked.
Echoes to [this issue](https://github.com/grafana/clickhouse-datasource/issues/977)

Screenshot 2025-03-25 at 16 08 17

SpencerTorres:
Hello! I appreciate all the changes and comments everyone contributed here.

The main value of this PR was solving how to properly decode these `Nested` columns, and it seems like this was solved for multiple cases (`flatten_nested` `=` `0`/`1`). Thank you for testing that

There were some large structural changes I wanted to address before merging this into the main branch, and instead of writing a bunch of suggestions I decided to fix them myself in a different PR: #1208

These are mainly related to how the components are used, along with fixing some test cases and adding flexibility. If you have any more comments related to the code of this PR, I would recommend writing them over there where I can address them asap.

Hopefully we can get this merged soon now that everything is working smoothly 👌

SpencerTorres:
> Just tested out this fork locally and note that span link redirection is not working so far.
> 
> When you click on the link it basically redirects you to the same trace instead of the one linked. Echoes to [this issue](https://github.com/grafana/clickhouse-datasource/issues/977)
> 
> Screenshot 2025-03-25 at 16 08 17

@ValentinLvr Hey! I finally found a sample row that shows this. Let's put it into a new issue, this might have something to do with the linking. Maybe even a conflict with our existing `TraceId` link. I'll ask the Grafana team

PR #1073, opened 2024-11-28
remove hard coded database name on opentelemetry-clickhouse dashboard

hoangphuocbk:
Data source with database name, which is not "default", will not work on opentelemetry-clickhouse dashboard.


SpencerTorres (approved):
This is good, now the datasource's default will apply. Thanks!

aangelisc (approved):

PR #1127, opened 2025-01-20
improve dev experience with devcontainer and debugger breakpoint support

OneCyrus:
this PR adds [DevContainer](https://code.visualstudio.com/docs/devcontainers/containers) support to simplify the dev environment setup. the devcontainer install all needed dependencies to build and run the plugin.

I also added a debug configuration to easily run the plugin inside the devcontainer.

image

with the added debug configuration comes also the mapping of sourcemaps to the local files. this allows to set breakpoints inside vscode and directly inspect the variables in the typescript code.

image


SpencerTorres:
Hello! Thanks for submitting this PR. There's already some guides/tools in place for Grafana plugin development, I feel like it may confuse contributors to have this added in this form. It adds another dependency, another config to be maintained, and there's also some quirks with how the plugin frontend/backend gets reloaded for code changes. I'm not sure this container solves those workflows the same way as the official guides recommend.

I personally use a direct clone of the Grafana repo and point it to my local plugin fork, it automatically reloads and falls back to the latest plugin binary when I'm not running it in debug mode in my IDE. Same for the frontend. The plugin architecture makes this work surprisingly seamlessly.

@bossinc wdyt? Perhaps I'm being too cautious about adding any new build/dev tools like this

aangelisc:
Thank you for the comment @SpencerTorres. I concur, we wouldn't be willing to accept this contribution at this time. As we have a standard method of developing our plugins that is consistent across all of them I wouldn't want to introduce a divergent method for ClickHouse.

Thank you for the contribution @OneCyrus but unfortunately we can't accept this at this time 😊

PR #1133, opened 2025-01-24
Add JSON paths to column selection

SpencerTorres:
When viewing a table with `JSON` columns, the paths are now included in the column suggestions. Since the path types are also included, the filters automatically adapt to the underlying type.

json path suggestions in query builder

Changes:
- updated tests
- updated changelog

### DIFF
--- a/src/data/CHDatasource.test.ts
+++ b/src/data/CHDatasource.test.ts
@@ -249,6 +249,73 @@ describe('ClickHouseDatasource', () => {
     });
   });
 
+  describe('fetchPathsForJSONColumns', () => {
+    it('sends a correct query when database and table names are provided', async () => {
+      const ds = cloneDeep(mockDatasource);
+      const frame = new ArrayDataFrame([
+        JSON.stringify({ keys: 'a.b.c', values: ['Int64'] }),
+        JSON.stringify({ keys: 'a.b.d', values: ['String'] }),
+        JSON.stringify({ keys: 'a.b.e', values: ['Bool'] })
+      ]);
+      const spyOnQuery = jest.spyOn(ds, 'query').mockImplementation((_request) => of({ data: [frame] }));
+      await ds.fetchPathsForJSONColumns('db_name', 'table_name', 'jsonCol');
+      const expected = { rawSql: 'SELECT arrayJoin(distinctJSONPathsAndTypes(jsonCol)) FROM "db_name"."table_name"' };
+
+      expect(spyOnQuery).toHaveBeenCalledWith(
+        expect.objectContaining({ targets: expect.arrayContaining([expect.objectContaining(expected)]) })
+      );
+    });
+
+    it('sends a correct query when only table name is provided', async () => {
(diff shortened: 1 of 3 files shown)


adamyeats (commented):
LGTM! Thanks for this. Just a small suggestion for one of the lines where I've switched the style to be more functional rather than imperative, but I left `fetchPathsForJSONColumns` as is as array order seems to be important here.

src/data/CHDatasource.test.ts
```diff
@@ -249,6 +249,73 @@ describe('ClickHouseDatasource', () => {
...
+      const expected = { rawSql: 'SELECT arrayJoin(distinctJSONPathsAndTypes(jsonCol)) FROM "table.name"' };
+
+      expect(spyOnQuery).toHaveBeenCalledWith(
+        expect.objectContaining({ targets: expect.arrayContaining([expect.objectContaining(expected)]) })
+      );
+    });
+
+    it('returns correct json columns', async () => {
+      const ds = cloneDeep(mockDatasource);
+      const frame = new ArrayDataFrame([
```
adamyeats, line 303:
```suggestion
      const frame = arrayToDataFrame([
```

`ArrayDataFrame` is deprecated, `arrayToDataFrame` in `@grafana/data` is the replacement.

src/data/CHDatasource.ts
```diff
@@ -550,12 +595,20 @@ export class Datasource
...
       picklistValues: [],
     }));
+
+    const jsonColumnNames = columns.filter(c => c.type.startsWith('JSON')).map(c => c.name);
+    for (let jsonColumnName of jsonColumnNames) {
+      const jsonColumns = await this.fetchPathsForJSONColumns(database, table, jsonColumnName);
+      columns.push(...jsonColumns);
+    }
+
+    return columns;
```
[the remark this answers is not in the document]
SpencerTorres: Good suggestion, this would be better. My only concern would be sending too many queries at once, but it should be fine. Hopefully nobody has more than a few JSON columns on their table 👀

src/data/CHDatasource.test.ts
SpencerTorres: Updated this and the rest of the references in this file 👍

SpencerTorres:
@adamyeats thanks for the review, updated according to your notes. I agree the `fetchPathsForJSONColumns` could've been less verbose but I was getting some incredibly odd results by doing the shorter `filter` and `map` functions. Perhaps something related to the data frame view's underlying iterator. Also wanted to avoid errors for `undefined` and such, so that's why that function looks so... _formal_

Thanks!

adamyeats (approved):
LGTM! 🚀 

> I was getting some incredibly odd results by doing the shorter filter and map functions

I noticed the same thing while doing some quick tests here... I'll look into this, but for now the code that's here is more than serviceable enough for the time being. Thanks again!

adamyeats (approved):

PR #1154, opened 2025-02-04
Refactor `clickhouse.Connect` to improve context cancellation handling

adamyeats:
This PR removes an unnecessary goroutine for `db.PingContext(ctx)`, preventing potential leaks when timeouts occur. It properly propagates `context.WithTimeout` throughout the function, ensuring `db.PingContext` and the PDC dialler respects the timeout. It also removes the `chErr` channel and simplifies the `select` statement for cleaner and safer connection handling.

### DIFF
--- a/pkg/plugin/driver.go
+++ b/pkg/plugin/driver.go
@@ -107,6 +107,7 @@ func (h *Clickhouse) Connect(ctx context.Context, config backend.DataSourceInsta
 	if err != nil {
 		return nil, err
 	}
+
 	var tlsConfig *tls.Config
 	if settings.TlsAuthWithCACert || settings.TlsClientAuth {
 		tlsConfig, err = getTLSConfig(settings)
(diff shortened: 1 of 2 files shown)


pkg/plugin/driver.go
```diff
@@ -195,25 +200,19 @@ func (h *Clickhouse) Connect(ctx context.Context, config backend.DataSourceInsta
...
+	default:
+		err := db.PingContext(ctx)
 		if err != nil {
-			// sql ds will ping again and show error
 			if exception, ok := err.(*clickhouse.Exception); ok {
 				log.DefaultLogger.Error("[%d] %s \n%s\n", exception.Code, exception.Message, exception.StackTrace)
 			} else {
 				log.DefaultLogger.Error(err.Error())
 			}
 			return db, nil
```
aangelisc, line 214: Not in your PR but I've just noticed this return, if the error is `nil` should we be returning the error?

aangelisc (approved):

pkg/plugin/driver.go
SpencerTorres: I assume the original intent of this is to log the non critical error and allow the driver to handle any potential reconnects. Grafana also has retry logic though, so maybe initial connection failure should be fatal?

SpencerTorres (approved):
Thank you for looking into this 🙏

pkg/plugin/driver.go
adamyeats: @aangelisc This is a great point, wdyt?

aangelisc: I think we can keep this as is for now 😊 If we decide the failing connection should be fatal we can make it so pretty easily!

PR #1155, opened 2025-02-04
Refactor `MutateResponse` function and PDC dialler creation

adamyeats:
This refactors the `MutateResponse` function and PDC dialler creation in `driver.go` to be a little cleaner and more idiomatic.

### DIFF
--- a/pkg/plugin/driver.go
+++ b/pkg/plugin/driver.go
@@ -55,6 +55,30 @@ func getTLSConfig(settings Settings) (*tls.Config, error) {
 	return tlsConfig, nil
 }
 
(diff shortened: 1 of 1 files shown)


beejeebus (approved):
LGTM

SpencerTorres (approved):
Looks good, appreciate the code cleanup in `MutateResponse` too!

I'm curious if the PDC dialer is able to properly switch between TLS and non-TLS dialers. This is an issue we had recently with a customer where they were using the wrong dialer by default. If it were a problem, it would be noticeable when someone is using a self-hosted (insecure) ClickHouse instance vs ClickHouse Cloud

aangelisc (approved):

PR #1156, opened 2025-02-07
Prevent usage of failed connections and improve connection management

adamyeats:
This PR improves connection management and error handling to ensure that failed connections are not returned to the pool, preventing the use of invalid connections.

- Ensures failed `PingContext(ctx)` connections are not returned to the pool. If `PingContext(ctx)` fails, `db` is no longer returned to prevent an unusable connection from being used.
- Moves timeout creation (`context.WithTimeout`) closer to `PingContext(ctx)`. Prevents premature cancellations that could occur during connection setup.
- Ensures that the timeout only applies to the connection attempt itself. 
- Adds explicit connection pool settings for better efficiency.
  - Sets `MaxOpenConns` to 50.
  - Sets `MaxIdleConns`  to 25.
- The `select` check was moved back to before `PingContext(ctx)` to avoid connecting if `ctx.Done()` is already triggered. The error from `PingContext(ctx)` is now handled outside of the `select` as  `PingContext(ctx)` is context aware.


aangelisc (dismissed):
This LGTM! The only suggestion I would make is to have connection lifetime, max idle connections, and max open connections configurable as we do in other data sources. It'd be nice to do that in this PR but feel free to also do it later on.

I definitely think this change should explicitly be called out in the changelog with the previous defaults vs the current defaults. I'm wondering if it constitutes a breaking change but I'm not sure if anyone would be reliant on these settings. Wdyt?

SpencerTorres (approved):
Always good to see this kind of async logic cleaned up 👍 

I agree with @aangelisc, more configuration is nice, even if we don't write the frontend UI for it yet

adamyeats:
@aangelisc @SpencerTorres Thanks both, totally agreed on your point about having this configurable in the UI. @aangelisc I think we talked about this previously too. I've added this in 723e9fd.

pkg/plugin/driver.go
```diff
@@ -215,21 +210,44 @@ func (h *Clickhouse) Connect(ctx context.Context, config backend.DataSourceInsta
...
 	}
 
+	ctx, cancel := context.WithTimeout(ctx, time.Duration(t)*time.Second)
+	defer cancel()
+
 	db := clickhouse.OpenDB(opts)
 
+	// Set connection pool settings
+	if i, err := strconv.Atoi(settings.ConnMaxLifetime); err == nil {
+		db.SetConnMaxLifetime(time.Duration(i) * time.Minute)
```
aangelisc, line 220: Nit: I'd suggest having this parsing logic in settings.go instead just so once we're in the driver then the settings object is ready to go without needing to carry out any more operations against it. Feel free to disregard!

adamyeats: Let's fix this up in a separate PR... `DialTimeout` and `QueryTimeout` are done the same way and we should fix this up in all cases, and I really want to get this PR out, so let's come back to this later.

PR #1158, opened 2025-02-07
feat: Enable CtrlCmd + Enter keyboard shortcut to Run Query

### LINKED ISSUE #1157: Enable Shift+Enter keyboard shortcut to Run Query
**What happened**:
When a user is on the Datasource Explore page and using the SQL Editor, and wants to use the usual keyboard shortcut "Shift + Enter" instead of pressing the "Run Query" button with the mouse, a new line is introduced.

When a user is on the Datasource Explore page and using the SQL Editor, and wants to use the usual keyboard shortcut "Shift + Enter" instead of pressing the "Run Query" button with the mouse, the query should be executed.

Access the SQL Editor and try to execute the query with any keyboard shortcut.

This is a standard and expected capability. It is available for instance in the Loki and Tempo Datasource plugins and something many users get used to use and miss when using the Clickhouse Datasource plugin. 

**Environment**:
- Grafana version: any
- Plugin version: v4.7.0
- OS Grafana is installed on: linux
- User OS & Browser: Linux Firefox

igor-borisoglebski:
Fixes #1157

### DIFF
--- a/src/components/SqlEditor.test.tsx
+++ b/src/components/SqlEditor.test.tsx
@@ -8,39 +8,39 @@ import { EditorType } from 'types/sql';
 
 // Mock the Monaco editor types
 const mockMonaco = {
-    KeyMod: {
-        CtrlCmd: 2048,
-        Shift: 1024,
-    },
-    KeyCode: {
-        Enter: 3
-    },
-    languages: {
-        getLanguages: () => [],
-        register: jest.fn(),
-        setMonarchTokensProvider: jest.fn(),
-        registerCompletionItemProvider: jest.fn(),
-    }
+  KeyMod: {
+    CtrlCmd: 2048,
+    Shift: 1024,
+  },
+  KeyCode: {
+    Enter: 3
+  },
+  languages: {
+    getLanguages: () => [],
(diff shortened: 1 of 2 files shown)


igor-borisoglebski:
Tested locally to confirm it works.

SpencerTorres:
Shift + Enter is often used for newlines, also I think there may be a way to add a jest test for this

Should this be Ctrl/Cmd + Enter?

igor-borisoglebski:
> also I think there may be a way to add a jest test for this
Unit test added 

> Should this be Ctrl/Cmd + Enter?
Changed it to `CtrlCmd + Enter`. I also find it to be more appropriate. I have initially set it to Shift to keep it consistent with the Grafana loki/prometheus/tempo datasource plugins - which all use `Shift + Enter`

> Edit: I just submitted this comment with Cmd + Enter 👀
I can't wait to be able to use it also in the ClickHouse datasource :laughing:

SpencerTorres (approved):
Thanks for changing the shortcut + tests 👍

adamyeats (approved):
LGTM! Thanks @igor-borisoglebski 🤝

PR #1190, opened 2025-02-27
update grafana dependencies to 11.5.2

OneCyrus:
updates the dependencies to 11.5.2 and adapts tests/mocks.


aangelisc:
Hi @OneCyrus, thank you for the contribution. Would you be able to add some context as to why this is needed at the moment?

OneCyrus:
@aangelisc to add more context. with the updated grafana UI library we get access to new controls which would improve the usability of this plugin. especially the new combobox.

PR #1204, opened 2025-03-21
SQL Editor Suggestions

SpencerTorres:
Implements SQL suggestions

suggestions sample for JSON

Here's an example of it suggesting columns, json paths, macros, and server functions for the given subquery.

## Details

This is an absolute hack of getting LSP-like logic running in our SQL Editor component. It's not a proper implementation, but it's massively better than what we have now (which is nothing).

This code is made to do these exact requirements, nothing more:
- Show column suggestions where expected (After a `WITH`, `SELECT`, `WHERE`, etc.)
- Show database/table suggestions where expected (After a `FROM` or `JOIN`)
- Show server functions such as `toDateTime64()` with documentation
- Show plugin macros such as `$__fromTime`
- Show Grafana dashboard variables such as `${my_variable}`
- Handle sub-query context. So if you're inside `SELECT (SELECT subq_col FROM B) FROM A` then table `B`'s column suggestions will show.
- Cache results so we don't spam the database

And it struggles even with this depending on the edge case.

I have a strong desire to write a proper ClickHouse SQL analyzer that can be used in TypeScript, but for now this is a very functional QoL feature for our SQL editor.

I did try other parsing tools, but I found that they were getting stuck on non-ClickHouse SQL (If you're familiar with our past problems with the adhoc filters, you'll know what I mean). This implementation is unsophisticated enough to simply parse the areas required for basic suggestions.

### Architecture:
1. There is none
2. Tokenize the SQL
3. Parse it into nodes (try to track incomplete tokens for suggestions)
4. Build cursor context using parsed nodes (Find where the cursor is within those nodes, and what type of SQL clause is expected)
5. Convert cursor context into suggestions

### Other notes:
- The lexer is translated from ClickHouse's C++ for maximum compatibility.
- Everything else is the product of lots of trial and error and iterations, and unfortunately does not make much sense

Let me know if you have any comments/suggestions. Thanks!


bossinc (approved):
I've been playing around with this for a while, and it works great! The code structure is nicely done. There isn't anything worth calling out that needs to change. I'd rather merge this and iterate as we find things.
Thank you for your hard work! :tada:

SpencerTorres:
> The code structure is nicely done.

Uhhhhhhhhhhh


Sure 👉 😎 👉 

*(I hope to improve it someday)*

bossinc:
drone is upset about spellcheck. `subquery` needs to be added to cspell.config.json

SpencerTorres:
> drone is upset about spellcheck. `subquery` needs to be added to cspell.config.json

Interesting. Spellcheck is passing when I run it, but I've added `subquery` just to be safe

bossinc (approved):

Issue #1103, opened 2025-01-09
Logs: can't see/quick-filter by SpanName, ResourceAttributes, ScopeAttributes


jwhitaker-gridcog:
When using the Logs query view, I think it's pretty important to be able to see the SpanName field (in python this is your logger name), and the ResourceAttributes field (this will have things like `deployment.environment.name`, again imo a basic need).

Is it possible to get these displaying + working in quick filter?

![Image](https://github.com/user-attachments/assets/9b7c4785-db50-444c-ac02-046c8ff94542)

SpencerTorres:
What does the SQL for this query look like?

jwhitaker-gridcog:
That's where I was unsure - does grafana give you hooks to send them back in separate columns? Or would they need to all be smushred together to whatever is backing `fields`?

SpencerTorres:
commented on the PR

PR #1214, opened 2025-04-05
RFC: Use ResourceAttributes and ScopeAttributes for log labels as well

jwhitaker-gridcog:
Heya, have just been trying to get something working on #1103 - currently when Exploring logs, you can see labels from LogAttributes, but not from ResourceAttributes or ScopeAttributes. This is pretty important functionality - "I have a random log line from somewhere in my k8s cluster, which pod did it come from?"

To query, I'm merging all the Attributes maps together and prefixing their keys so I can track where they came from.

I wasn't 100% sure on the best way to model this - I flipped and flopped between having a special `.labelColumns` field added on `QueryBuilderOptions` , or in a special `.magicMapForLabels` bool on `SelectedColumn` - I went with the latter.

This seems to work OK!
image

SQL Preview
```sql
SELECT Timestamp as "timestamp", Body as "body", SeverityText as "level", mapConcat(mapApply((k, v) -> ('attr.' || k,  v), "LogAttributes"),mapApply((k, v) -> ('res.' || k,  v), "ResourceAttributes"),mapApply((k, v) -> ('span.' || k,  v), "ScopeAttributes")) as "labels", TraceId as "traceID" FROM "otel"."otel_logs" WHERE ( timestamp >= $__fromTime AND timestamp <= $__toTime ) AND ( ResourceAttributes['k8s.namespace.name'] = 'sys-nginx-gateway' ) AND ( ResourceAttributes['k8s.node.name'] = 'i-04fb08d3a1fcd7eed.ap-southeast-2.compute.internal' ) ORDER BY timestamp DESC LIMIT 1000
````

Could I ask for a pointer if this is an ok approach to take to achieve this? I can't promise I'd be able to finish it off but I can probably get pretty far at least.

### DIFF
--- a/src/components/queryBuilder/ColumnsEditor.tsx
+++ b/src/components/queryBuilder/ColumnsEditor.tsx
@@ -12,6 +12,8 @@ interface ColumnsEditorProps {
   onSelectedColumnsChange: (selectedColumns: SelectedColumn[]) => void;
   disabled?: boolean;
   showAllOption?: boolean;
+  label?: string;
+  tooltip?: string;
 }
 
 function getCustomColumns(columnNames: string[], allColumns: readonly TableColumn[]): Array<SelectableValue<string>> {
@@ -24,7 +26,10 @@ function getCustomColumns(columnNames: string[], allColumns: readonly TableColum
 const allColumnName = '*';
 
 export const ColumnsEditor = (props: ColumnsEditorProps) => {
-  const { allColumns, selectedColumns, onSelectedColumnsChange, disabled, showAllOption } = props;
+  const {
+    allColumns, selectedColumns, onSelectedColumnsChange, disabled, showAllOption,
+    label = labels.components.ColumnsEditor.label, tooltip = labels.components.ColumnsEditor.tooltip
+  } = props;
   const [customColumns, setCustomColumns] = useState<Array<SelectableValue<string>>>([]);
   const [isOpen, setIsOpen] = useState(false);
   const allColumnNames = allColumns.map(c => ({ label: c.label || c.name, value: c.name }));
@@ -32,7 +37,6 @@ export const ColumnsEditor = (props: ColumnsEditorProps) => {
     allColumnNames.push({ label: allColumnName, value: allColumnName });
   }
   const selectedColumnNames = (selectedColumns || []).map(c => ({ label: c.alias || c.name, value: c.name }));
-  const { label, tooltip } = labels.components.ColumnsEditor;
 
   const options = [...allColumnNames, ...customColumns];
 
--- a/src/components/queryBuilder/views/LogsQueryBuilder.tsx
+++ b/src/components/queryBuilder/views/LogsQueryBuilder.tsx
@@ -32,7 +32,7 @@ interface LogsQueryBuilderState {
   timeColumn?: SelectedColumn;
   logLevelColumn?: SelectedColumn;
   messageColumn?: SelectedColumn;
-  labelsColumn?: SelectedColumn;
+  labelsColumns: SelectedColumn[];
   // liveView: boolean;
   orderBy: OrderBy[];
   limit: number;
@@ -78,8 +78,9 @@ export const LogsQueryBuilder = (props: LogsQueryBuilderProps) => {
     if (next.messageColumn) {
       nextColumns.push(next.messageColumn);
     }
-    if (next.labelsColumn) {
-      nextColumns.push(next.labelsColumn);
+
+    for (const c of next.labelsColumns) {
+      nextColumns.push({ ...c, mapForLabels: true })
     }
 
     builderOptionsDispatch(setOptions({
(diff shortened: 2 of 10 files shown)


jwhitaker-gridcog:
specific design things I was wanting some guidance on:

- Is ramming all of these inside one `labels` column really needed? My limited understanding of Grafana's requirements is that it is, however if Grafana allowed some official way of providing multiple `labels` cols that I'm unaware of, this would be both less hacky, and may appear separated in the UI as well, both of which would be desirable.

- Assuming we do need to put everything inside one `labels`, is prefixes the way to do it? I've somewhat arbitrarily gone with `attr.${key}` for LogAttributes, `res.${key}` for ResourceAttributes, and `scope.${key}` for ScopeAttributes. This is user-exposed so I wanted it to be reasonably succinct - something else I wondered about was using a valid sql identifier, e.g. `ResourceAttributes['${key}']` etc. This is probably too verbose to be exposed in the `labels` block though. N.B. I should probably add some comment to this effect, but this namespacing system is not just for UX, it's also wired back through the Filters system so the quick +/- buttons work, so needs to be reversible.

lasermoth:
I just came across this, as I was also going down the rabbit hole of wanting ResourceAttributes as log labels. 

My current work around was to disable the "use Otel" flag and run `mapConcat(LogAttributes, ResourceAttributes)` on the labels field which is rather ugly. 

I believe the `labels` field is special. I can at least see is used [in the loki datasource](https://github.com/grafana/grafana/blob/eb4de388c3cf197a97a490de4e038a0018307203/public/app/plugins/datasource/loki/datasource.ts#L1070) this way. It does also appear [exclusively referenced](https://github.com/search?q=repo%3Agrafana%2Fgrafana+path%3A%2F%5Epublic%5C%2Fapp%5C%2Ffeatures%5C%2Fexplore%5C%2FLogs%5C%2F%2F++labels&type=code). 

Looking forward to this contribution.

lasermoth:
Is this still being worked on, it should we look to open a new PR?

SpencerTorres:
Hello! Thanks for submitting this. This is a known issue, but there isn't really a clear way to solve it. I'll elaborate by answering your questions:

> Is ramming all of these inside one labels column really needed? My limited understanding of Grafana's requirements is that it is; Grafana needs the resulting dataframe to have a single column of labels. However, if Grafana allowed some official way of providing multiple labels cols that I'm unaware of, this would be both less hacky, and may appear separated in the UI as well, both of which would be desirable.

Yes, Grafana looks for a `labels` field on the data frame. There's no official way to have multiple label fields, but we did discuss the possibility of merging them on the backend. Something like `labels1` `labels2` `labels3` would be merged by either the backend or frontend processing logic. I was hesitant to add this since it's very opinionated and specific to our datasource, I didn't want to break any workflows that coincidentally use the same field names as these.

> Assuming we do need to put everything inside one labels, is prefixes the way to do it? I've somewhat arbitrarily gone with attr.${key} for LogAttributes, res.${key} for ResourceAttributes, and scope.${key} for ScopeAttributes. This is user-exposed so I wanted it to be reasonably succinct - something else I wondered about was using a valid sql identifier, e.g. ResourceAttributes['${key}'] etc. This is probably too verbose to be exposed in the labels block though. N.B. I should probably add some comment to this effect, but this namespacing system is not just for UX, it's also wired back through the Filters system so the quick +/- buttons work, so needs to be reversible.

This is where the label merging idea gets a bit murky. Yes I would go with prefixing, but once we start aliasing these as `labels1` `labels2`, etc. it gets confusing on how we enable filtering on these values.

In the logs panel there are hooks whenever you "filter in" and "filter out" a value. All we know is the actual field name. So if your field name is `pod.name`, we don't really know whether that actually came from `ResourceAttributes` or `labels1` or some other column, it's just whatever's in `labels`. If you were to click filter on this I believe it would add a filter for something like `labels['pod.name'] = 'some value'`. All we know is that the value came from the `labels` field. 

We could add prefixes for `ResourceAttributes` so that it appears as `ResourceAttributes.pod.name` and then we can check for whether this prefix matches a selected column and then auto convert it, making `ResourceAttributes.pod.name` turn into `ResourceAttributes['pod.name']`, but this requires that the labels were merged, and that the `ResourceAttributes` was selected. There are a lot of edge cases here. This same problem applies to our Ad-Hoc filters feature. I believe there's already some logic in there that does something like this, but it clearly doesn't cover the full use case.

If you have any more ideas on how to handle this, I'm glad to explore them. I'll bring this up again internally to see if we can come up with a clean solution for this. We might just have to make the `Labels` dropdown accept multiple map/json columns, or maybe have them pre-configured per datasource.

jwhitaker-gridcog:
Thanks for looking at this @SpencerTorres . Yeah, this PR does nearly what you're suggesting with ResourceAttributes.pod.name, but I introduced the convention `res.pod.name` as this is user-exposed and brevity might be appreciated. It does seem to work, however it probably doesn't cope with some of the edge cases you're talking about.

I wonder if an approach to make it less flaky on if merged labels appear in the query or not, might be to add an alias column to the otel logs table itself in the schema? That would be more of a guarantee that merged labels are around, and additionally could help nail down the specific prefixes you want to use.

jwhitaker-gridcog:
that looks awesome Spencer, thankyou!!

PR #1215, opened 2025-04-07
Add support for the general `row_limit` config setting

### LINKED ISSUE #1134: clickhouse datasource is not respecting dataproxy config row_limit
**What happened**:
Hello, 
There is a config in grafana:
https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#row_limit

> row_limit - Limits the number of rows that Grafana will process from SQL (relational) data sources. Default is 1000000.

This prevents Grafana from being killed in case someone did not provide explicit LIMIT in query. 

In postgres datasource it looks like this:

Image

While in clicklhouse datasource it does not respect the parameter:

Image

clickhouse datasource would fail to query more than 3000 rows

Set:
```
   dataproxy:
      logging: "true"
      timeout: "1200"
      row_limit: "3000"
```
Then make a query with LIMIT 3001 and see the result.

**Environment**:

- Grafana version: Grafana v11.2.2 (dbf571744d)
- Plugin version: v4.6.0 
- OS Grafana is installed on: linux
- User OS & Browser: MacOs Chrome

aangelisc:
This PR adds support for the general `row_limit` configuration setting.

If set, this will override the `limit` set in the query builder.

Fixes #1134

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+# Unreleased
+
+### Features
+
+- Add support for the Grafana `row_limit` [configuration setting](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#row_limit).
+
 ## 4.8.2
 
 ### Fixes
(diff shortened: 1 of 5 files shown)


bossinc (approved):
Nice!

SpencerTorres (approved):
Thanks!

PR #1234, opened 2025-05-02
change `host` to `server` in yaml config example in README

wshirey:
The `host` YAML property was not being recognized by grafana, however `server` is recognized.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ datasources:
     jsonData:
       defaultDatabase: database
       port: 9000
-      host: localhost
+      server: localhost
       username: username
       tlsSkipVerify: false
       # tlsAuth: <bool>


SpencerTorres:
server side:
https://github.com/grafana/clickhouse-datasource/blob/7bfce721533031badbdcd5bfeda4ee8d5cb66744/pkg/plugin/settings.go#L19

server side compatibility change:
https://github.com/grafana/clickhouse-datasource/blob/7bfce721533031badbdcd5bfeda4ee8d5cb66744/pkg/plugin/settings.go#L75-L81

client side:
https://github.com/grafana/clickhouse-datasource/blob/7bfce721533031badbdcd5bfeda4ee8d5cb66744/src/views/CHConfigEditor.tsx#L229-L232

client side compatibility change:
https://github.com/grafana/clickhouse-datasource/blob/7bfce721533031badbdcd5bfeda4ee8d5cb66744/src/views/CHConfigEditorHooks.ts#L76-L82


Can you share more info about what version you're using and what exactly is in your config? This change has been around for a while and has tests to verify it works. I haven't heard of this not working, I'm wondering if you're still using v3 and copied a v4 config?

aangelisc:
+1 to @SpencerTorres's reply above. Are you able to provide us with more information @wshirey , otherwise we'll close this out 😊

SpencerTorres:
Closing; see related comments. Feel free to re-open if this is still an issue. Thanks!

PR #1241, opened 2025-05-13
Fix ad-hoc filter application with templated target tables

srstrickland:
Previously, ad-hoc filters could fail when applied to queries containing templated table names because filters were being applied before template variables were resolved. This change ensures template variables are properly resolved before ad-hoc filters are applied, making the filtering system more reliable when working with dynamic table references.

In particular, we have a lot of fully generic dashboards where a user can select a database, and then from that database a table to explore.  The query in the panel looks like:

```
SELECT ...
FROM ${database}.${table}
WHERE ...
```

This breaks ad-hoc filters, because it's looking for a the literal `somedb.sometbl` string in the raw, pre-interpolated SQL.  This just changes the order of operations to do interpolation first.


adamyeats (approved):
Thanks @srstrickland , apologies for the wait here!

PR #1265, opened 2025-06-05
Ad-hoc queries: Allow to filter by values inside the map

aiakovlev-twilio:
If we follow opentelemetry conventions we rely heavily on maps (ResourceAttributes, LogAttributes, ScopeAttributes)
But Clickhouse plugin for Grafana do not allow us to filter data inside these maps dynamically.

Behaviour before the change:
Screenshot 2025-06-05 at 22 17 02

Behaviour after the change:
This PR allows us to query resource attribute keys if they added to query builder this way:
Screenshot 2025-06-09 at 23 17 08

After that we can click to add these fields to ad-hoc filters and it will work:
452071708-87093772-e090-4480-a0f2-3388071d1009

Statistics result:
Screenshot 2025-06-05 at 22 13 33

This is important because default behavior works only for top-level fields without dots in names and per my opinion not compatible with opentelemetry format.

I've added tests so it shouldn't break previous logic, but add new feature to slice&dice for any field in logs.

Please let me know if it doesn't make any sense and there is good standard way to solve my problem.

P.S. This PR also solving our another issue that plugin couldn't process columns with "." in key name. It's not so common in top level column names, but naming with dots is opentelemetry semantic convention and almost all keys have dots in names.


SpencerTorres (approved):

SpencerTorres:
I'm always cautious with changes to the ad-hoc filtering, it's notoriously broken and hard to fit every use case. This looks good though, I appreciate the detailed PR.

adamyeats (approved):
LGTM, thanks @aiakovlev-twilio!

zdyj3170101136:
this is really very strange.

PR #1268, opened 2025-06-09
Add push and publish Github Action files

beejeebus:
These files mostly copy-pasta from the instructions in Plugins Platform (excellent) documentation for migrating from Drone to GHA.


bossinc:
We are running tests using GHA now. We should remove those in favor of this GHA

beejeebus:
> We are running tests using GHA now. We should remove those in favor of this GHA

I think we should remove the `push.yml` from this PR. we only need the `publish.yml` bits to be able to turn off Drone. I *think*.

bossinc (dismissed):

aangelisc (approved):

PR #1294, opened 2025-06-26
Configuration fixes related to row limit implementation

### LINKED ISSUE #1256: v4.9.0 seems broken with grafana 10.1.0
**What happened**:

Since 4.9.0 released yesterday, I haven't been able to query ClickHouse from Grafana. Any query, even just `SHOW TABLES` results in the error `SQLDatasourceMaxOpenConnsDefault not set in config`. 

Screenshot:
Image

The logs include lines like
```logger=plugin.grafana-clickhouse-datasource t=2025-05-23T03:32:10.237184461Z level=error msg="Partial data response error" dsName=ClickHouse dsUid=clickhouse endpoint=queryData error="SQLDatasourceMaxOpenConnsDefault not set in config" uname= pluginId=grafana-clickhouse-datasource refID=0.322196045463234 status=500 statusSource=downstream```

The same thing happens if I open the data source page, and click the "Test" button.

Queries work as before

- Run grafana 10.1.0 with `GF_INSTALL_PLUGINS=grafana-clickhouse-datasource`
- Open grafana in chrome
- Open the developer console
- Open "Explore" and select the ClickHouse data source
- The last request in the developer console fails with the above error.
- Trying to do queries results in the same error.

I can't easily upgrade the version of Grafana that we deploy. Also, since I'm using an older version, the `GF_PLUGINS_PREINSTALL` env variable isn't supported. That's unfortunate since it allows pinning a plugin to a specific version.

Instead, I need to use `GF_INSTALL_PLUGINS`, which makes Grafana fail to start with `Plugin not found` if I try to add the suffix `@4.8.2`.

Also, If I locally run a later version of Grafana, like 11.6.0, with the plugin version 4.9.0, I get a different error: `error querying the database: code: 164, message: Cannot modify 'limit' setting in readonly mode`. Screenshot: 

Image

This error goes away if I roll back to plugin version 4.8.2 using the env variable `GF_PLUGINS_PREINSTALL=grafana-clickhouse-datasource@4.8.2`.

**Environment**:

- Grafana version: 10.1.0
- Plugin version: 4.9.0
- OS Grafana is installed on: Ubuntu
- User OS & Browser: Ubuntu & Chrome
- Others:

### PR DESCRIPTION
**Pull request by aangelisc**
#1215 introduced two separate issues in the plugin:

- Older versions of Grafana wouldn't function as expected because the SQL configuration properties did not exist before v11.0.0.
- Clickhouse requires additional permissions to support setting the `limit` property at the driver level.

This PR adds a toggle in the data source config to support enabling/disabling the row limit setting. It defaults to `false` to avoid breaking data source configurations.

If the toggle is set to `true` and the user is on a version of Grafana >= 11.0.0 with the correct permissions then the limit will be applied correctly.

If the toggle is set to `true` and the user is on a version of Grafana >= 11.0.0 with incorrect permissions then the query will fail.

If the toggle is set to `true` and the user is on a version of Grafana < 11.0.0 then the data source configuration will fail.

If the toggle is set to `false` the limit will never be set and the data source will not attempt to read the SQL configuration values from the Grafana configuration (essentially the same as before #1215 was merged).

Fixes #1256

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,22 @@
+name: Plugins - CI
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+permissions: {}
+
+jobs:
+  ci:
+    name: CI
+    uses: grafana/plugin-ci-workflows/.github/workflows/ci.yml@main # zizmor: ignore[unpinned-uses]
+    permissions:
+      contents: read
+      id-token: write
+    with:
+      plugin-version-suffix: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || '' }}
+      go-version: '1.24'
+      golangci-lint-version: '2.1.6'
+      run-playwright: false

### CODE REVIEW

File: pkg/plugin/settings.go [hunk 1]
```diff
@@ -186,6 +187,19 @@ func LoadSettings(ctx context.Context, config backend.DataSourceInstanceSettings
 		}
 	}
 
+	settings.EnableRowLimit = false
+	if jsonData["enableRowLimit"] != nil {
+		if enableRowLimitString, ok := jsonData["enableRowLimit"].(string); ok {
+			settings.EnableRowLimit, err = strconv.ParseBool(enableRowLimitString)
+			if err != nil {
+				backend.Logger.Warn("Failed to parse enableRowLimit value, defaulting to false", "error", err)
+				settings.EnableRowLimit = false
```
beejeebus: `settings.EnableRowLimit` doesn't need to be explicitly set to false on 190, as bool type zero value is false.

also, `strconv.ParseBool()` returns `false` in the error [case](https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/strconv/atob.go;l=17), so we don't need 196 either.

pkg/plugin/settings.go
```diff
@@ -186,6 +187,19 @@ func LoadSettings(ctx context.Context, config backend.DataSourceInstanceSettings
...
+			}
+		} else {
+			settings.EnableRowLimit = jsonData["enableRowLimit"].(bool)
```
beejeebus, line 197: are we sure of the type of the `enableRowLimit` key? if it's an int, this will [panic](https://go.dev/play/p/5sBWxFLzsm7).

beejeebus (dismissed):
left one nit and one question, otherwise LGTM.

bossinc (changes requested):
I like the lint changes buuuut I think we should make this a separate PR. Its a lot of noise for this issue

pkg/plugin/settings.go
aangelisc: Updated

pkg/plugin/settings.go
aangelisc: We control the type of that property in the frontend 😊 I added the string check to be overly cautious in the provisioning case but otherwise it shouldn't be anything but a `string` or `bool`.

beejeebus: righto.

beejeebus (approved):

SpencerTorres (approved):

Issue #1307, opened 2025-07-14
Nullable JSON column doesn't work


mvrhov:
**What happened**:
Error is returned. `clickhouse [Decode]: data unsupported JSON serialization version for decode: 18446744073709551615: Could not process SQL results`

**What you expected to happen**:
Query returns data in a table

**How to reproduce it (as minimally and precisely as possible)**:

Create a table with some data:
 ```
CREATE TABLE IF NOT EXISTS example (
			time DateTime64(3) CODEC(ZSTD),
			data Nullable(JSON) CODEC(ZSTD)
		) Engine = MergeTree;

INSERT INTO example (time, data) VALUES (now(), null);
INSERT INTO example (time, data) VALUES (now(), '{"duration":355668,"packet":{"id":1,"size":8,"type":"control"}}');
```

Go to grafana and do explore on the CH datasource with example table.
Select database and example table. Then select `*` under columns and pres Run Query button.

**Environment**:

- Grafana version: 12.x
- Plugin version: 4.10.1

PR #1326, opened 2025-07-30
Fix large JSON objects + complex JSON types

SpencerTorres:
There are some bugs with the JSON type in the plugin:
- Large JSON/Dynamic objects will break with decoding errors
- `Nullable(JSON)` does not work
- `JSON(x LowCardinality(Nullable(String)))` typed path with prefixed types such as `LowCardinality` do not work
- Plugin cannot decode strings if the server forces JSON to be sent as strings
- Plugin cannot decode new JSON/Dynamic serialization version for handling large JSON objects
- HTTP was not sending proper timezone

All of these have been addressed in this update.

Closes https://github.com/grafana/clickhouse-datasource/issues/1307

## Large JSON Objects

If you're using our new OTel JSON exporter its likely you'll run into large JSON objects. This will sometimes break the plugin since the driver did not support the extended portion of data for these large objects. 

This can be solved in a few ways:
- Use ClickHouse v25.6+ and enable `output_format_native_use_flattened_dynamic_and_json_serialization`. This will use a new object encoding that can handle the full JSON/Dynamic object.
- Use ClickHouse v24.10+ and enable `output_format_native_write_json_as_string`. This will enable string encoding. Use this option if you're unable to update to ClickHouse v25.6+. (This only fixes JSON, if your Dynamic types have low type limits they can still break)
- Cast your `JSON` or `Dynamic` types to `String` or another preferred type when querying, or select specific fields on the JSON object.

Eventually one these settings will be enabled by default either by the server or the plugin, but for now you must manually add them during the migration period.

These settings can be added to your datasource under the "Custom Settings" section.

custom settings

For more details about this limitation you can follow the links in this PR: https://github.com/ClickHouse/clickhouse-go/pull/1590

## Complex JSON types

Typed paths and dynamic paths should now be working for complex types such as `Nullable(JSON)` and `JSON` with `LowCardinality` types within it.

* *Note that `Nullable(JSON)` is only available on ClickHouse v25.2+* *

proper decoding of new complex JSON types

This upgraded logic should work whether you're using string or object encoding at the driver level.

## Other changes in this PR

There were some other issues I found while adding this, I have decided to fix them to help organize the code:
- Updated to clickhouse-go v2.40.1 for latest driver fixes
- Changed the SQL type `Converters` map to an array since maps have inconsistent access. I want the converters to have a controllable order in which they are checked.
- Simplified the definition of some regexes. I added these incorrectly before. Nothing broken, just duplicated code.
- Corrected the package name for `converters_test.go`, now we can test unexported functions
- Updated broken tests for types dependent on JSON conversion (JSON, Map, Tuple, etc.)
- Updated FieldType from the deprecated nullable json to the preferred `FieldTypeJSON`
- Simplified logic and added test cases for checking `LowCardinality` types
- The jsonConverter function can now handle raw strings and similar types, this was required for handling JSON in string and object format.
- Added test cases for JSON object and string converters
- Fixed HTTP timezone test since driver now supports timezones being sent over HTTP/Native


beejeebus:
I did a scan of this PR, looks good!

> Changed the SQL type Converters map to an array since maps have inconsistent access. I want the converters to have a controllable order in which they are checked.

If the order of definition of these converters is now a load-bearing concept, I'd add a test. would be easy for something as dumb as a formatting rule to change the order, without the dev understanding the implications.

beejeebus (dismissed):
LGTM, :ship: with a test asserting the order of converters.

SpencerTorres:
> If the order of definition of these converters is now a load-bearing concept, I'd add a test. would be easy for something as dumb as a formatting rule to change the order, without the dev understanding the implications.

There are no cases where this occurs right now, I can't think of a test where the order matters. This is more of a cleanup to prevent issues in the future as more types are added. I myself have copy pasted converters and regular expressions in the past, and it would be very easy to make a mistake with it.

Instead of a test I have added comments warning about the ordering and potential overlap.

I could also revert it and make it a map again since there's not a functional need for this change right now, it was just something I noticed that could cause incredibly obscure bugs later

See these commnets I added:
```go
// matchRegexes is a mapping of regular expressions.
// When adding entries, try to prevent overlap in regular expressions.
// For example `^Date` and `^DateTime` could conflict when matching `DateTime64`
var matchRegexes = map[string]*regexp.Regexp{

// Converters defines a list of type converters.
// When a converter is looked up by name or regex, it will be in the order they are defined below.
// This is important for regular expressions that may overlap or conflict.
var Converters = []Converter{
```

adamyeats:
@SpencerTorres Can you try giving this a rebase? I'm hoping that resolves the pipeline fails here. 🤞

PR #1330, opened 2025-08-02
Fix Ad-Hoc filters for variable datasources

SpencerTorres:
The current function for fetching ad-hoc variables was deprecated over 2 years ago.
To fetch ad-hoc filters, it required the name of the datasource.

```js
const adHocFilters = (templateSrv as any)?.getAdhocFilters(this.name);
```

Some dashboards use variable datasources such as `${datasource}`, so the filters would show as empty and never be applied.

I updated this function to use the new ad-hoc variable source, and now filters are applied even when the datasource is a variable.

```js
// filters comes from the request.filters, which is the new recommended pattern
// for fetching ad-hoc filters from the dashboard
applyTemplateVariables(query: CHQuery, scoped: ScopedVars, filters: AdHocVariableFilter[] = [])
```

filter example

There are still other bugs and limitations associated with filters, but this makes the behavior more consistent for some dashboards.

### DIFF
--- a/src/data/CHDatasource.ts
+++ b/src/data/CHDatasource.ts
@@ -238,7 +238,7 @@ export class Datasource
     rawQuery = this.replace(rawQuery, scoped) || '';
 
     if (!this.skipAdHocFilter) {
-      if (this.adHocFiltersStatus === AdHocFilterStatus.disabled && filters?.length > 0) {
+      if (this.adHocFiltersStatus === AdHocFilterStatus.disabled && filters.length > 0) {
         throw new Error(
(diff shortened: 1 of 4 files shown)


src/data/CHDatasource.ts
```diff
@@ -136,7 +136,7 @@ export class Datasource
       return undefined;
     }
 
-    
+
```
adamyeats, line 139: Very minor,  but can we remove these two additional lines?

adamyeats (approved):
LGTM! 👍

PR #1335, opened 2025-08-11
Add additional tracking events for v1 config

alyssajoyner:
This PR adds additional tracking events we want to capture for the v1 configuration page. Specifically, tracking the server and port input fields allow for better insights into the full user flow.


bossinc (changes requested):
The tracking events should be in the on change functions not in the HTML

SpencerTorres (approved):
I forgot the exact syntax but it would be nice to collapse that import list into some kind of `import * as track`. Not a blocker though. Looks good, thanks!

bossinc (dismissed):

bossinc (approved):

PR #1341, opened 2025-08-13
Fix bug where switched to logs query type errored

bossinc:
When switching to logs before `contextColumns` was defined, it would throw an `not an array` error.


SpencerTorres (approved):

alyssajoyner (approved):
Nice catch 💯

Issue #1262, opened 2025-05-30
Span events & links not selected when querying trace from table view


ValentinLvr:
**What happened**:

When I'm jumping to a TraceId link from a *table* result, events & links are not in the `SELECT` clause of the generated SQL.

In more details:
1. I'm doing a basic query with the *table* query type
```sql
SELECT TraceId 
FROM default.otel_traces
LIMIT 10
```
![Image](https://github.com/user-attachments/assets/5316fe5a-ab3a-47e5-8f3d-2675acb36248)

2. Then, I click on the `TraceId` embedded link to view the trace on a second panel

![Image](https://github.com/user-attachments/assets/c3d7263c-a16d-4245-8f85-345348be3f71)

```sql
WITH
  'XXXXXXX' as trace_id,
  (
    SELECT
      min(Start)
    FROM
      "default"."otel_traces_trace_id_ts"
    WHERE
      TraceId = trace_id
  ) as trace_start,
  (
    SELECT
      max(End) + 1
    FROM
      "default"."otel_traces_trace_id_ts"
    WHERE
      TraceId = trace_id
  ) as trace_end
SELECT
  "TraceId" as traceID,
  "SpanId" as spanID,
  "ParentSpanId" as parentSpanID,
  "ServiceName" as serviceName,
  "SpanName" as operationName,
  multiply (toUnixTimestamp64Nano ("Timestamp"), 0.000001) as startTime,
  multiply ("Duration", 0.000001) as duration,
  arrayMap (
    key -> map ('key', key, 'value', "SpanAttributes" [key]),
    mapKeys ("SpanAttributes")
  ) as tags,
  arrayMap (
    key -> map ('key', key, 'value', "ResourceAttributes" [key]),
    mapKeys ("ResourceAttributes")
  ) as serviceTags,
  if (
    "StatusCode" IN ('Error', 'STATUS_CODE_ERROR'),
    2,
    0
  ) as statusCode,
  "SpanKind" as kind,
  "StatusMessage" as statusMessage,
  "TraceState" as traceState
FROM
  "default"."otel_traces"
WHERE
  traceID = trace_id
  AND "Timestamp" >= trace_start
  AND "Timestamp" <= trace_end
LIMIT
  1000
```

As you can see, events & links are not added by default on the `SELECT` clause.
I either have to add it myself OR unselect then select again the **Use Otel 1.29** button 

**What you expected to happen**:

[...]

PR #1345, opened 2025-08-19
Correctly display events and links for table view queries

mcorbin:
Fix https://github.com/grafana/clickhouse-datasource/issues/1262

Pass the events and links configuration to the query and use otel columns if opentelemetry is enabled.

This allows displaying events in the trace panel.

I tested it locally: the events are correctly displayed when coming from a table trace ID.

### DIFF
--- a/src/data/utils.ts
+++ b/src/data/utils.ts
@@ -188,11 +188,7 @@ export const transformQueryResponseWithTraceAndLogLinks = (datasource: Datasourc
       };
 
       if (otelConfig?.traceColumnMap) {
-        const columns: SelectedColumn[] = [];
-        otelConfig.traceColumnMap.forEach((name, hint) => {
-          columns.push({ name, hint });
-        });
-        options.columns = columns;
+        options.columns = Array.from(otelConfig.traceColumnMap, ([hint, name]) => ({ name, hint }));
       } else {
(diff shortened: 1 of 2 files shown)


src/data/utils.ts
```diff
@@ -178,12 +182,22 @@ export const transformQueryResponseWithTraceAndLogLinks = (datasource: Datasourc
...
 
-      const defaultColumns = datasource.getDefaultTraceColumns();
-      for (let [hint, colName] of defaultColumns) {
-        options.columns!.push({ name: colName, hint });
+      if (otelConfig?.traceColumnMap) {
+        const columns: SelectedColumn[] = [];
+        otelConfig.traceColumnMap.forEach((name, hint) => {
+          columns.push({ name, hint });
+        });
+        options.columns = columns;
```
adamyeats:
```suggestion
        options.columns = Array.from(otelConfig.traceColumnMap, ([hint, name]) => ({ name, hint }));
```

adamyeats (approved):
LGTM! One minor comment below.

src/data/utils.ts
mcorbin: Thank you for your review.
I pushed a commit with your suggestion.

adamyeats (approved):

PR #1369, opened 2025-09-17
Merge OTel labels

SpencerTorres:
The Logs UI currently has a "Labels" field for logs. By default it selects `LogAttributes` to show in the logs explore panel, but this excludes important information from `ResourceAttributes` and `ScopeAttributes`.

This PR adds logic to auto merge these columns (JSON or Map type) into a single `labels` field for easy viewing and filtering. To use this, simply use Logs visualization with at least one of `ResourceAttributes`, `LogAttributes`, or `ScopeAttributes` selected. The labels will not be merged if there's an existing `labels` field selected, so you may have to clear the "Labels" input on the query builder.

Example:
example label merge

Note that the fields are prefixed with the associated column. Instead of showing only `a`, it is displayed as `ResourceAttributes.a`. This is done to prevent label conflicts, and to give context for filtering logic.

## Filtering

Filtering is also enhanced with this. Because the labels are prefixed, we can easily convert these filters into the correct syntax for `Map` and `JSON` types.

### JSON Filter Example

Top-level JSON columns have a `{}` appended at the end in the column selector to show the type.

JSON filter

Generated SQL:
```sql
ResourceAttributes.`env` = 'staging'
```

Note that the path is escaped with back ticks to prevent errors for complex path names:
```sql
ResourceAttributes.`path with spaces`.`a`.`b` = 'staging'
```

Another important detail is that JSON paths are not yet flattened. I expect this to be a server side setting at some point. You can manually type in the path as a workaround if the filter does not apply correctly. In a future PR we can add a setting to let the driver do the flattening, but this is risky due to potential path conflicts.

### Map Filter Example

Top-level Map columns have a `[]` appended at the end in the column selector to show the type.

Map filter

Generated SQL:
```sql
ResourceAttributes['env'] = 'staging'
```

### DIFF
--- a/pkg/plugin/driver.go
+++ b/pkg/plugin/driver.go
@@ -311,6 +311,72 @@ func (h *Clickhouse) MutateQuery(ctx context.Context, req backend.DataQuery) (co
 // except for specific visualizations (traces, tables, and logs).
 func (h *Clickhouse) MutateResponse(ctx context.Context, res data.Frames) (data.Frames, error) {
 	for _, frame := range res {
+		if frame.Meta.PreferredVisualization == data.VisTypeLogs {
+			var attrFields []*data.Field
+			for _, field := range frame.Fields {
+				if field.Type() != data.FieldTypeJSON {
+					continue
+				}
+
+				if field.Name == "ResourceAttributes" || field.Name == "ScopeAttributes" || field.Name == "LogAttributes" {
+					attrFields = append(attrFields, field)
+				}
+			}
+
+			rowLen, err := frame.RowLen()
+			if err != nil {
+				return nil, err
+			}
+
+			allLabelsValues := make([]map[string]any, rowLen)
+
+			for _, field := range attrFields {
+				for j := 0; j < rowLen; j++ {
+					currentVal := allLabelsValues[j]
+					if currentVal == nil {
+						currentVal = make(map[string]any)
+					}
+
+					val := field.At(j).(json.RawMessage)
+					if val != nil {
+						var valMap map[string]any
+						err := json.Unmarshal(val, &valMap)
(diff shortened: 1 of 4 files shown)


src/data/CHDatasource.ts
```diff
@@ -579,7 +589,11 @@ export class Datasource
         continue;
       }
 
-      const kv = JSON.parse(x[0]);
+      let kv = x[0];
+      if (typeof kv === 'string') {
+        kv = JSON.parse(x[0]);
+      }
```
adamyeats:
```suggestion
      const kv = typeof kv === 'string' ? JSON.parse(x[0]) : x[0];
```

pkg/plugin/driver.go
```diff
@@ -411,3 +418,75 @@ func extractForwardedHeadersFromMessage(message json.RawMessage) (map[string]str
 
 	return httpHeaders, nil
 }
+
+func mergeOpenTelemetryLabels(frame *data.Frame) error {
```
adamyeats, line 422: Would you be able to add some unit tests for this code?

SpencerTorres:
@adamyeats added tests to the suggested function. It looks like an unrelated test is failing due to docker/network/connection issue, but all other tests are passing. Not sure if that's just my machine with a conflicting port

aiakovlev-twilio:
>ResourceAttributes.a
Such naming convention will break ad-hoc filters because they will think that "ResourceAttributes" is a table name.
Is this known issue or am i missing something?

SpencerTorres:
Updated the PR to flatten JSON paths

aiakovlev-twilio:
Hello, this new functionality doesn't work for map type, only for the json

PR #1371, opened 2025-09-17
Update log and trace column names to match ClickHouse otel defaults

avleentwilio:
In ClickHouse the logs and trace tables use the order:
`ServiceName, TimestampTime, Timestamp`
This is set by the clickhouse exporter for opentelemetry-collector-contrib.

Currently in this plugin the defaults when use_otel is enabled sets the timestamp column to `Timestamp`. As a result ClickHouse has to search much more data, making queries against very large tables quite slow.
On my 1.5TB logs table, `Timestamp` queries take 94.5s, whereas the same `TimestampTime` queries take 10.1s.

The correct timestamp column to use would be `TimestampTime`.

Ref: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/clickhouseexporter/internal/sqltemplates/logs_table.sql

I'm unsure if there are more changes I need to make here, any advice would be appreciated!

### DIFF
--- a/src/otel.ts
+++ b/src/otel.ts
@@ -26,7 +26,7 @@ const otel129: OtelVersion = {
   specUrl: 'https://opentelemetry.io/docs/specs/otel',
   logsTable: defaultLogsTable,
   logColumnMap: new Map<ColumnHint, string>([
-    [ColumnHint.Time, 'Timestamp'],
+    [ColumnHint.Time, 'TimestampTime'],
     [ColumnHint.LogMessage, 'Body'],
     [ColumnHint.LogLevel, 'SeverityText'],
     [ColumnHint.LogLabels, 'LogAttributes'],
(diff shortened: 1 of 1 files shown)


adamyeats:
> I'm unsure if there are more changes I need to make here, any advice would be appreciated!

@SpencerTorres Are you able to comment on this?

avleentwilio:
Yeah I get that. But surely the default option should be the fast and
mostly accurate one?

Or we should flip the short order for the table in ClickHouse to have
Timestamp first and just use that.

But the current state is the worst of all of the options :-)

adamyeats:
@SpencerTorres Any follow-up on this one?

vstariradev:
I may be missing something obvious from the discussion here so please excuse/ignore me if I'm way off the point.

`Time` field can be kept as `Timestamp` for accuracy and keep queries fast if we put `TimestampTime` in the the WHERE/Filters clause. This means decoupling Time from the association of it being what you display and what you filter by. In other words, the datasource settings could auto-fill (set default) Time as `Timestamp` but in WHERE/Filters to have TimestampTime because that's what the default otel schema (dictated by clickhouseexporter) is optimized for.
Ordering (ORDER BY) is not even necessary to be set because Grafana itself offers sorting by Newest first / Oldest First, it's just hidden until any results come in.
Some may be worried about incomplete results but as long as Timestamp is within the TimestampTime (seconds) it will get picked up in the result set and sorted regardless by grafana according to the Time field.
image

bossinc:
We should make the default timestamp `TimestampTime`, as @avleentwilio is suggesting. We can add a new column hint called `FilteredTimestamp` that uses `Timestamp` for ordering.

PR #1396, opened 2025-10-13
Updates for testing in CI

adamyeats:
This PR migrates the old, handwritten GitHub Action for E2E tests to the new GitHub Action from `plugin-ci-workflows`. Unit tests and integration tests have now been split out into separate workflows, and integration tests can now be run by using `go test -tags=integration`. 

This PR also updates Grafana Bench to the latest version.


.github/workflows/grafana-bench.yml
```diff
@@ -44,19 +44,20 @@ jobs:
...
+        with:
+          url: http://localhost:3000
 
       - name: Run Grafana Bench tests
         run: |
-          docker run --rm \
+          docker run \
+            --rm \
             --network=host \
-            --volume="./:/home/bench/tests/" \
```
aangelisc, line 54: Does this work for forks?

adamyeats: I believe so - at worst it should be a no-op as it's not really a functional change from what was there to begin with.

PR #1400, opened 2025-10-14
Run E2E against Grafana Cloud

adamyeats:
Closes https://github.com/grafana/data-sources/issues/673, https://github.com/grafana/data-sources/issues/667 and https://github.com/grafana/data-sources/issues/668.

This PR adds a Github Action to run our E2E tests against Grafana Cloud, on a timed `cron` job. It also pushes `main` to the plugin catalog on PR merge.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -24,7 +24,7 @@ jobs:
   publish-latest-to-catalog:
     name: Publish `main` to Dev Catalog
     if: github.ref == 'refs/heads/main'
-    uses: grafana/plugin-ci-workflows/.github/workflows/cd.yml@main # zizmor: ignore[unpinned-uses]
+    uses: grafana/plugin-ci-workflows/.github/workflows/cd.yml@ci-cd-workflows/v5.1.0
     permissions:
       attestations: write
       contents: write


bossinc (approved):

.github/workflows/push.yml
```diff
@@ -12,10 +12,29 @@ jobs:
   ci:
     name: CI
     uses: grafana/plugin-ci-workflows/.github/workflows/ci.yml@main # zizmor: ignore[unpinned-uses]
+    if: github.ref != 'refs/heads/main'
```
aangelisc, line 15: This will stop CI from running on `main` won't it? I don't think we want this.

.github/workflows/push.yml
```diff
@@ -12,10 +12,29 @@ jobs:
...
     permissions:
       contents: read
       id-token: write
     with:
+      go-version: '1.24'
+      golangci-lint-version: '2.1.6'
       plugin-version-suffix: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || '' }}
+
+  publish-latest-to-catalog:
```
aangelisc, line 24: I assume this will publish with the version specified in `package.json` - will this break because the version won't have changed?

.github/workflows/push.yml
adamyeats: The `cd.yml` workflow actually runs the the `ci.yml` workflow too - this prevents it from running twice.

.github/workflows/push.yml
adamyeats: Good catch - added `plugin-version-suffix` to the `publish-latest-to-catalog` action to ensure the version is distinct.

.github/workflows/push.yml
aangelisc: Will the `needs` fail then if this workflow doesn't run? IHAF it'll be skipped but I'm not sure

adamyeats: I've already remvoed the `needs` clause. 👍

.github/workflows/cron.yml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Get secrets from Vault
+        id: get-secrets
+        uses: grafana/shared-workflows/actions/get-vault-secrets@get-vault-secrets/v1.3.0
+        with:
+          # Grafana auth (used by @grafana/plugin-e2e)
+          # + Prometheus creds for Bench metrics reporting
+          common_secrets: |
+            PLAYWRIGHT_GRAFANA_PASSWORD=data-sources/e2e:grafana-pw
+            PLAYWRIGHT_GRAFANA_USERNAME=data-sources/e2e:grafana-username
+            PROMETHEUS_PASSWORD=grafana-bench:prometheus_token
```
aangelisc, line 43: Which Prom does this point to?

adamyeats: I believe it's `prometheus-ops-03-ops-eu-south-0`.

PR #1410, opened 2025-10-21
Update glob

bossinc:


adamyeats:
As discussed in standup, this package is a stub and can now just be removed

PR #1422, opened 2025-10-27
Config page design adjustments

alyssajoyner:
UI updates to the Config page:

- Remove numbers in CollapsableSection headers
- Fix margin mismatch in DatabaseCredentialsSection description
- Update spacing and margins on LeftSidebar
- Update placeholder text

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,5 @@
     "semver": "7.7.3",
     "sql-formatter": "^15.6.10",
     "tslib": "2.8.1"
-  },
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+  }
 }
--- a/src/views/config-v2/AdditionalSettingsSection.tsx
+++ b/src/views/config-v2/AdditionalSettingsSection.tsx
@@ -1,12 +1,20 @@
 import { ConfigSubSection } from 'components/experimental/ConfigSection';
 import allLabels from './labelsV2';
-import React, { ChangeEvent, useState } from 'react';
+import React, { ChangeEvent, useMemo, useState } from 'react';
 import {
   DataSourcePluginOptionsEditorProps,
   onUpdateDatasourceJsonDataOption,
   onUpdateDatasourceJsonDataOptionChecked,
 } from '@grafana/data';
-import { AliasTableEntry, CHConfig, CHCustomSetting, CHLogsConfig, CHSecureConfig, CHTracesConfig } from 'types/config';
+import {
+  AliasTableEntry,
+  CHConfig,
+  CHCustomSetting,
+  CHLogsConfig,
+  CHSecureConfig,
+  CHTracesConfig,
+  defaultCHAdditionalSettingsConfig,
+} from 'types/config';
 import { AliasTableConfig } from 'components/configEditor/AliasTableConfig';
 import { DefaultDatabaseTableConfig } from 'components/configEditor/DefaultDatabaseTableConfig';
 import { LogsConfig } from 'components/configEditor/LogsConfig';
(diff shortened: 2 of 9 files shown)


src/views/config-v2/LeftSidebar.tsx
```diff
@@ -12,37 +12,36 @@ export const LeftSidebar = ({ pdcInjected }: LeftSidebarProps) => {
...
+              <LinkButton
+                style={header.isOptional ? { padding: '5px 15px', height: '50px', width: '225px' } : {}}
+                variant="secondary"
+                fill="text"
+                onClick={(e) => {
+                  e.preventDefault();
+                  const target = document.getElementById(header.id);
+                  if (target) {
+                    const y = target.getBoundingClientRect().top + window.scrollY - 60;
+                    window.scrollTo({ top: y, behavior: 'smooth' });
```
bossinc: Is this the "normal" way to handle scroll to? If so, i think we should pull this out of the html and make a function. I would argue we shouldnt have condisionals here

bossinc (changes requested):
Just a few questions but it looks great! (Totally valid for you to ignore all my suggestion/questions)

src/views/config-v2/CHConfigEditor.tsx
```diff
@@ -28,7 +28,7 @@ export const ConfigEditor: React.FC<ConfigEditorProps> = (props) => {
         <div className={styles.requiredFields}>
           <Alert
             severity="info"
-            title="You are viewing a new design for the Clickhouse configuration settings."
+            title="You are viewing a new design for the ClickHouse configuration settings."
```
SpencerTorres, line 31: Consistency 🤌

src/views/config-v2/LeftSidebar.tsx
adamyeats:
```suggestion
                    const top = target.getBoundingClientRect().top + window.scrollY - 60;
                    window.scrollTo({ top, behavior: 'smooth' });
```

adamyeats: I agree with @bossinc here 👍

src/views/config-v2/DatabaseCredentialsSection.tsx
```diff
@@ -77,7 +77,10 @@ export const DatabaseCredentialsSection = (props: Props) => {
               onBlur={trackClickhouseConfigV2DatabaseCredentialsUserInput}
             />
           </Field>
-          <Field label={labels.password.label} description={labels.password.tooltip}>
+          <Field
+            label={labels.password.label}
+            description={<div style={{ marginTop: '5px' }}>{labels.password.tooltip}</div>}
```
adamyeats: Can we use the design tokens (e.g. `theme.spacing(value)`) here instead?

src/views/config-v2/AdditionalSettingsSection.tsx
```diff
@@ -110,6 +110,72 @@ export const AdditionalSettingsSection = (props: Props) => {
     });
   };
 
+  // if any of the settings in this section are set, open the section by default
```
bossinc: This check could be simplified if we added all the settings in an additional settings to an object, and then checked to see if the object has changed from its default state (which will be a new default object). This way, if we add another additional setting, we won't have to remember to add it here

package.json
```diff
@@ -88,5 +88,6 @@
     "semver": "7.7.3",
     "sql-formatter": "^15.6.10",
     "tslib": "2.8.1"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
bossinc: No longer using yarn

src/views/config-v2/AdditionalSettingsSection.tsx
```diff
@@ -52,6 +52,69 @@ export const AdditionalSettingsSection = (props: Props) => {
   useConfigDefaults(options, onOptionsChange);
 
   const [customSettings, setCustomSettings] = useState(jsonData.customSettings || []);
+  const ADDITIONAL_SETTINGS_DEFAULTS = {
```
bossinc: lets talk about this

src/module.ts
```diff
@@ -10,7 +10,7 @@ import { analyzeQueries, trackClickhouseDashboardLoaded } from 'tracking';
 import pluginJson from './plugin.json';
 import clickhouseVersion from '../package.json';
 
-// ConfigEditorV2 is the new design for the Clickhouse configuration page
+// ConfigEditorV2 is the new design for the ClickHouse configuration page
 const configEditor = config.featureToggles['newClickhouseConfigPageDesign' as keyof FeatureToggles]
```
aangelisc, line 14: Is the `as keyof` required here?

src/views/config-v2/AdditionalSettingsSection.tsx
aangelisc: I'd love to hear what you all decide here. I think if we need this it should be pulled out somewhere else at the very least 😊

src/views/config-v2/AdditionalSettingsSection.tsx
```diff
@@ -52,6 +52,69 @@ export const AdditionalSettingsSection = (props: Props) => {
...
+      flattenNested: jsonData.traces?.flattenNested,
+      traceEventsColumnPrefix: jsonData.traces?.traceEventsColumnPrefix,
+      traceLinksColumnPrefix: jsonData.traces?.traceLinksColumnPrefix,
+    },
+    aliasTables: jsonData.aliasTables && jsonData.aliasTables.length > 0 ? jsonData.aliasTables : undefined,
+    enableRowLimit: jsonData.enableRowLimit,
+    enableSecureSocksProxy: jsonData.enableSecureSocksProxy,
+  };
+
+  const hasDefinedValue = (obj: any): boolean => {
```
aangelisc: I'd suggest pulling this out as a utility function. Also, can `obj` truly be `any`? I wonder if we could restrict the types here a little bit.

src/views/config-v2/AdditionalSettingsSection.tsx
```diff
@@ -52,6 +52,69 @@ export const AdditionalSettingsSection = (props: Props) => {
...
+    if (obj == null) {
+      return false;
+    }
+    if (typeof obj !== 'object') {
+      return obj !== '' && obj !== undefined && obj !== null;
+    }
+    return Object.values(obj).some((v) => hasDefinedValue(v));
```
aangelisc: Nit: Is an object invalid if nested properties don't have a defined value? I think this function could do with a little documentation

src/module.ts
alyssajoyner: Yes - I believe so. `newClickhouseConfigPageDesign` is not available on config.featureToggles so I have to do this fun hack 😊 Lmk if there is something I'm missing!

PR #1427, opened 2025-10-30
give the user more feedback when client creation fails

antonysouthworth-halter:
Screenshot 2025-10-31 at 11 48 19

Hopefully obvious but, that error message doesn't give me a whole lot to go on.

I think we should instead wrap the error so that the user is given some feedback as to what's gone wrong. Otherwise you basically just have to guess in the absence of any logs (such as when using Grafana Cloud, as we are).

### DIFF
--- a/pkg/plugin/driver.go
+++ b/pkg/plugin/driver.go
@@ -253,7 +253,7 @@ func (h *Clickhouse) Connect(ctx context.Context, config backend.DataSourceInsta
 		}
 
 		backend.Logger.Error("failed to create ClickHouse client", "error", err)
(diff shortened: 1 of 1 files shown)


SpencerTorres (approved):
Looks good, but need to make sure no secrets get leaked through logs 👍

adamyeats:
Hi @antonysouthworth-halter , thanks for the contribution! Unfortunately, we're unable to accept this PR at present. The error message here has been designed to be deliberately opaque, as presenting the error string from that error can leak secrets into logs e.g. in Grafana Cloud. We're currently working on a solution for this internally that should present more information to the user, while ensuring that secrets aren't leaked into logs.

PR #1429, opened 2025-11-06
Fix flaky test TestHTTPConnectWithHeaders (#1397)

### LINKED ISSUE #1397: Flaky or indeterministic unit test: `TestHTTPConnectWithHeaders`
```
--- FAIL: TestHTTPConnectWithHeaders (0.00s)
    --- FAIL: TestHTTPConnectWithHeaders/should_not_forward_http_headers (0.00s)
        driver_test.go:1248: 
            	Error Trace:	/opt/actions-runner/_work/clickhouse-datasource/clickhouse-datasource/pkg/plugin/driver_test.go:1248
            	Error:      	Not equal: 
            	            	expected: <nil>(<nil>)
            	            	actual  : status.ErrorWithSource(status.ErrorWithSource{source:"downstream", err:status.ErrorWithSource{source:"downstream", err:status.ErrorWithSource{source:"downstream", err:(*errors.errorString)(0xc000c227b0)}}})
            	Test:       	TestHTTPConnectWithHeaders/should_not_forward_http_headers
        driver_test.go:1251: 
            	Error Trace:	/opt/actions-runner/_work/clickhouse-datasource/clickhouse-datasource/pkg/plugin/driver_test.go:1251
            	Error:      	Not equal: 
            	            	expected: true
            	            	actual  : false
            	Test:       	TestHTTPConnectWithHeaders/should_not_forward_http_headers
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x101e11b]
```

Intermittent unit test failures in CI on this unit test.

MattHandzel:
The test was failing intermittently due to a race condition where the
datasource connection attempt occurred before the proxy server was ready
to accept connections.

This fix adds synchronization logic that polls the proxy server port
until it's accepting connections (with a 5-second timeout)

Changes:

- Added net.DialTimeout polling loop after starting proxy server
- Added 100ms retry interval with 50 retries (5 second total timeout)
- Test fails explicitly if proxy doesn't start within timeout

Fixes #1397

Creating this PR to test in CI


pkg/plugin/driver_integration_test.go
```diff
@@ -1202,6 +1202,20 @@ func TestHTTPConnectWithHeaders(t *testing.T) {
...
+		conn, err := net.DialTimeout("tcp", proxyAddr, 100*time.Millisecond)
+		if err == nil {
+			conn.Close()
+			break
+		}
+		if i == 49 {
+			t.Fatalf("Proxy server failed to start after 5 seconds")
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
```
adamyeats, line 1227:
```suggestion
	// Wait for proxy server to be ready
	proxyAddr := net.JoinHostPort(proxyHost, proxyPort)

	waitCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	defer cancel()

	d := &net.Dialer{}
	tick := time.NewTicker(100 * time.Millisecond)
	defer tick.Stop()

	for {
		conn, err := d.DialContext(waitCtx, "tcp", proxyAddr)
		if err == nil {
			_ = conn.Close()
			break
		}

		select {
		case <-waitCtx.Done():
			t.Fatalf("proxy server failed to start within 5s: %v", err)
		case <-tick.C:
		}
	}
```

With my suggestion, we avoid magic “50 attempts × 100ms” loop while keeping the same behaviour as your original proposal, by using a `context`. There are couple of points about the differences here:

- This introduces guarantees that we don’t exceed the wall-clock budget even if the scheduler slips.
- `time.Ticker` has a clearer intent and prevents busy-waiting on retries.
- `net.JoinHostPort(proxyHost, proxyPort)` handles IPv6 correctly (might not be necessary in this case but good to have it this way in case there are changes here in future) and avoids manual string building. `fmt.Sprintf("%s:%s", xxx)` doesn't do this.

Is this change acceptable to you?

PR #1433, opened 2025-11-06
Format Referer header on X-Dashboard-Title and X-Panel-Title

velom:
The plugin can forward HTTP headers.
Unfortunately, ClickHouse stores only `Referer` header in the `system.query_log`.
This PR adds additional header `Referer` based on `X-Dashboard-Title` and `X-Panel-Title`.

Examle of how it looks in the `query_log`:
```
    ┌─http_referer────────────────────────────────────────────────┐
 1. │ Explorer                                                    │
 2. │ Dashboard title: "New dashboard", pannel title: "New panel" │
    └─────────────────────────────────────────────────────────────┘
```

### DIFF
--- a/pkg/plugin/driver.go
+++ b/pkg/plugin/driver.go
@@ -462,7 +462,7 @@ func extractForwardedHeadersFromMessage(message json.RawMessage) (map[string]str
 		return nil, errors.New("Couldn't parse message as args")
 	}
 
-	var dashboard, pannel string
+	var dashboard, panel string
 	httpHeaders := make(map[string]string)
 	if grafanaHttpHeaders, ok := messageArgs[sqlds.HeaderKey]; ok {
 		fwdHeaders, ok := grafanaHttpHeaders.(map[string]interface{})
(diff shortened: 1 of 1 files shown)


SpencerTorres (changes requested):
I like this, but is it possible to use the array-like behavior of the headers to add more data? I want to make it easier to fetch this out of the query log. The string formatting works, but I'm wondering if we can make it more HTTP-compliant

What I mean is, is it possible where the header can have multiple entries? I forgot the interface in Go, but there should be a way to do that

SpencerTorres (approved):
I suppose this is better than what we have now, I just wish we had a better way to attach metadata to queries. This doesn't apply to TCP connections does it?

pkg/plugin/driver.go
```diff
@@ -458,9 +465,23 @@ func extractForwardedHeadersFromMessage(message json.RawMessage) (map[string]str
...
+			if k == headerPanelTitle {
+				panel = httpHeaders[k]
+			}
 		}
 	}
 
+	if dashboard != "" || panel != "" {
+		httpHeaders[headerReferer] = fmt.Sprintf("Dashboard title: %q, panel title: %q", dashboard, panel)
+	} else {
+		httpHeaders[headerReferer] = "Explorer"
```
aangelisc, line 499: This is technically untrue; the query could also come from alerting. You should be able to validate this by checking the `uname` from the plugin context for `grafana_scheduler`.

aangelisc (commented):
Generally this LGTM. I wonder if we should have this configurable - I'm not sure all users would want this enabled, wdyt?

PR #1454, opened 2025-11-22
Fix docker image name in DEV_GUIDE

taro-yanagiya:
Update the Clickhouse image in DEV_GUIDE, since [`clickhouse/clickhouse-client` image](https://hub.docker.com/r/clickhouse/clickhouse-client/) is no longer maintained.


adamyeats:
Hey @taro-yanagiya thank you for the contribution! Having given this some thought, I think we'll merge this, with thanks - however, `DEV_GUIDE.md` needs a complete overhaul to be in line with how we work on Clickhouse today, e.g. using the `compose.yaml` file, so we'll add a ticket to ensure this is done.

PR #1456, opened 2025-11-25
Fix filter for Map type LogLabels

waterdrag0n:
Fix filter not working when adding via + button in Explore logs

https://github.com/grafana/clickhouse-datasource/pull/1369 introduced OTel attribute support but broke existing Map/JSON type filter logic.

## Problem 1: Empty mapKey
When clicking + button on a log field from a Map type column (e.g., LogAttributes as labels) in Explore, the filter was created with an empty key.

- Expected: `labels['key'] = 'value'`
- Actual: `labels = 'value'`

Fix: Fallback to columnName when mapKey is not set by OTel prefix parsing.

## Problem 2: Empty columnType
When OTel mode is enabled, columns were set without type information, causing Map columns to be incorrectly treated as JSON.

- Expected: `labels['key'] = 'value'` (Map syntax)
- Actual: `` labels.`key` = 'value' `` (JSON syntax)

Fix: Fetch column type from table schema in `useOtelColumns`.

Related: https://github.com/grafana/clickhouse-datasource/pull/1369#issuecomment-3437260770

### DIFF
--- a/src/components/queryBuilder/views/LogsQueryBuilder.tsx
+++ b/src/components/queryBuilder/views/LogsQueryBuilder.tsx
@@ -108,7 +108,7 @@ export const LogsQueryBuilder = (props: LogsQueryBuilderProps) => {
   }, builderState);
 
   useLogDefaultsOnMount(datasource, isNewQuery, builderOptions, builderOptionsDispatch);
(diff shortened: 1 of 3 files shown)


SpencerTorres (approved):
Thanks!

adamyeats (approved):
LGTM, thanks @waterdrag0n ! 🙌

PR #1462, opened 2025-12-01
Fix "View logs" link if you aren't using otel schema in Explore and Dashboards view

aiakovlev-twilio:
Context:
We investigated why links broken in our configuration and looking at codebase we found that clickhouse grafana plugin doesn't handle Links generation in case if "Use OTel" Data Source setting set to "False".
When it's set to "True", plugin look at default column hints and takes "ColumnHint.TraceId" because it's defined. 
If it's not set to "True". it takes {}.

While worked on fix I tried different options:
1. hardcode TraceId for logs to logs correlation (didn't continue because of hardcoding)
2. implement UI for TraceId column in logs configuration only for correlations (didn't continue because resulted codebase looks identical to existing tracing configuration)
3. Implement TraceId column ID discovery from tracing configuration, but re-use this for logging. Maybe it breaks some internal guidelines (please let me know if I can make it better), but my solution works in following cases:
  - "Use OTel" set to True for logging 
  - "Use OTel" set to False for logging and "Use OTel" set to True for tracing
  - "Use OTel" set to False for logging and "Use OTel" set to False for tracing with override for TraceId column ID

Current behaviour:
When you go to explore or any dashboard, write any query with "TraceId" field in column list, click "View Logs".
For explorer - side panel opens and query appears, but without any "WHERE" statements
For dashboard - side panel opens and query appears, but without any "WHERE" statements
Screenshot 2025-12-01 at 19 33 52

Fixed behaviour:
When you go to explore or any dashboard, write any query with "TraceId" field in column list, click "View Logs".
For explorer - side panel opens and proper query appears
For dashboard - new window opens and proper query appears

Screenshot 2025-12-01 at 19 18 51
Screenshot 2025-12-01 at 19 27 10
Screenshot 2025-12-01 at 19 24 32

### DIFF
--- a/src/data/utils.test.ts
+++ b/src/data/utils.test.ts
@@ -233,7 +233,7 @@ describe('transformQueryResponseWithTraceAndLogLinks', () => {
     const out = transformQueryResponseWithTraceAndLogLinks(mockDatasource, request, response);
 
     const links = out?.data[0]?.fields[0]?.config?.links;
-    const viewLogsLink = links?.find((link) => link.title === 'View logs');
+    const viewLogsLink = links?.find((link: any) => link.title === 'View logs');
 
     const logsQuery = viewLogsLink?.internal?.query as CHBuilderQuery;
     expect(logsQuery.builderOptions.columns).toBeDefined();


SpencerTorres (approved):
Thanks!

aangelisc (approved):
LGTM, thank you @aiakovlev-twilio!

src/data/utils.test.ts
```diff
@@ -211,10 +211,47 @@ describe('transformQueryResponseWithTraceAndLogLinks', () => {
...
+    const builderOptions: Partial<QueryBuilderOptions> = {
+      queryType: QueryType.Traces,
+      columns: [{ name: 'a' }],
+    };
+
+    const [request, response] = buildTestRequestResponse(builderOptions);
+    const out = transformQueryResponseWithTraceAndLogLinks(mockDatasource, request, response);
+
+    const links = out?.data[0]?.fields[0]?.config?.links;
+    const viewLogsLink = links?.find((link) => link.title === 'View logs');
```
bossinc: lint is complaining that `link` is implicitly type any.
```suggestion
    const viewLogsLink = links?.find((link: any) => link.title === 'View logs');
```

aiakovlev-twilio:
thank you for your help @bossinc !
if someone could update contibuting.md file with more steps how I can check linter and other CI\CD checks locally it will definitely help in future to avoid this issue.

Issue #1434, opened 2025-11-08
Map type broken for Ad-Hoc filters


SpencerTorres:
There have been multiple reports of the Map type being broken for Ad-Hoc filters for OTel data. Trouble with the filters being applied as well as errors in showing the filter values as seen below:

<img width="1092" height="640" alt="object showing in ad hoc filter suggestions" src="https://github.com/user-attachments/assets/1801d4df-7dfd-48cd-92cd-dd15d1911a1f" />

PR #1464, opened 2025-12-03
Add dashboard variable to control AdHoc filter syntax

SpencerTorres:
Fixes a regression from #1369 (reported in #1434) by allowing users to configure per-dashboard whether their OTel labels are Map or JSON type. Basically reverting the AdHoc filter changes and making them opt-in.

TL;DR: To "opt in" to JSON AdHoc filters, add `clickhouse_adhoc_use_json` as a constant variable (the `value` doesn't matter) to your dashboard.

## Details

For OTel data, we merge and flatten `ResourceAttributes`, `LogAttributes`, and `ScopeAttributes` into a single `labels` field. The `labels` field then has values such as `ResourceAttributes.host.name`.

This means for JSON type, we can use the `ResourceAttributes.host.name` syntax directly, while for the Map type we must change the syntax to `ResourceAttributes['host.name']`.

example adhoc filter

Because the above Ad-Hoc filter doesn't specify type, we must add a way to configure this syntax to interpret as JSON or Map. We can't add it at the top level config, since it may be different per dashboard. So instead a user can simply add `clickhouse_adhoc_use_json` to enable the JSON syntax. If this is not present then it will default to Map.

Let me know if you have any thoughts on how we might improve this solution.

### DIFF
--- a/src/data/CHDatasource.test.ts
+++ b/src/data/CHDatasource.test.ts
@@ -123,7 +123,10 @@ describe('ClickHouseDatasource', () => {
       const spyOnGetVars = jest.spyOn(templateSrvMock, 'getVariables').mockImplementation(() => []);
 
       // Setup ad-hoc filters
-      const adHocFilters = [{ key: 'column', operator: '=', value: 'value' }];
+      const adHocFilters = [
+        { key: 'column', operator: '=', value: 'value' },
+        { key: 'column.nested', operator: '=', value: 'value2' }
+      ];
 
       // Mock getAdhocFilters to return our test filters
       jest.spyOn(templateSrvMock, 'getAdhocFilters').mockImplementation(() => adHocFilters);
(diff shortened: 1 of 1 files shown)


aangelisc (commented):
The changes LGTM but I'm uncertain on the enablement mechanism. I'd prefer updating the `AdHocVariableForm` in Grafana core to allow plugins to provide additional config 🤔

vstariradev:
Bump. Our team would really appreciate this PR getting some attention in the coming weeks. 🤞

src/data/CHDatasource.ts
```diff
@@ -245,7 +246,9 @@ export class Datasource
           `Unable to apply ad hoc filters. Upgrade ClickHouse to >=${this.adHocCHVerReq.major}.${this.adHocCHVerReq.minor} or remove ad hoc filters for the dashboard.`
         );
       }
-      rawQuery = this.adHocFilter.apply(rawQuery, filters);
+
+      const useJSON = Boolean(templateSrvVariables.find(v => v.name === 'clickhouse_adhoc_use_json'));
```
aangelisc, line 254: I think let's add a test for this specifically.

PR #1481, opened 2025-12-18
Support columns with `.` in ad-hoc filters

### LINKED ISSUE #1251: Ad hoc filters: Column names with dots in them get truncated when a query is constructed
### What happened?

I'm using "Ad hoc filters" on a dashboard that visualizes data taken from a ClickHouse database. 

Our table (`runtime_resource_usage`) may look like a one having nested columns, but it is flat - there are dots in some column names, which are there historically.
Column name examples: `master.max_cpu.real_time_total`, `master.total_size.external`, ...

Grafana allows creating an adhoc filter on these columns correctly, e.g. `runtime_resource_usage.master.max_cpu.real_time_total > 90`. However, these filters don't work: 400 (Bad Request).
Query inspector shows that everything after `master` is omitted (see the last line, where `' master > \'90\' '`):

```
SELECT
  toStartOfInterval(toDateTime(time), INTERVAL 300 second) as time,
  max("master.max_cpu.real_time_total") as "master.real_time_total",
  "device"
FROM default.runtime_resource_usage
WHERE
  (time >= toDateTime(1743116629) AND time <= toDateTime(1743721429))
GROUP BY time, "device"
ORDER BY time ASC settings additional_table_filters={'default.runtime_resource_usage' : ' master > \'90\' '}
```

This makes me think that column names get mistakenly truncated when constructing filter expressions.

### What did you expect to happen?

I would expect the last line of the query to look like this:
```
ORDER BY time ASC settings additional_table_filters={'default.runtime_resource_usage' : ' master.max_cpu.real_time_total > \'90\' '}
```

### Did this work before?

I don't know

### How do we reproduce it?

1. Create a small table in ClickHouse
```
DROP TABLE IF EXISTS test_dots;
CREATE TABLE test_dots (
	build_number UInt32,
	`master.max_cpu.real_time_total` Float32
)
ORDER BY build_number
INSERT INTO test_dots(
	build_number,
	`master.max_cpu.real_time_total`
)
VALUES
	(1000, 70),
	(1000, 71),
	(1000, 72),
	(1000, 73),
	(1000, 74)
```

2. Add a dashboard
3. Add variables to that dashboard:
	* Ad hoc filters, `filter`
	* Constant, Name: `clickhouse_adhoc_query`, Value: `default.test_dots`
4. Add a visualization to the dashboard:
```
SELECT
build_number,
`master.max_cpu.real_time_total`
FROM default.test_dots
```
5. Add a filter: `test_dots.master.max_cpu.real_time_total = 72` and observe errors in developer's panel

### Is the bug inside a dashboard panel?

_No response_

### Environment (with versions)?

Grafana: v11.6.0 (d2fdff9ee4)
OS: Ubuntu 24
Browser: Chrome

### Grafana platform?

I don't know

### Datasource(s)?

ClickHouse 4.8.0 plugin

aangelisc:
Currently, columns that contain a `.` will have their name truncated when used with ad-hoc filters. This will lead to a malformed query.

I've updated the logic to appropriately rejoin the parts when splitting the table and column names. I've also updated the logic that retrieves tag values to ensure we retrieve the values for the correct column.

It's worth noting that `.` values are also valid in table names. At this point in time we are *not* supporting tables with `.`'s as well as columns with `.`'s.

I've also added tests for this case.

Fixes #1251

### DIFF
--- a/src/data/adHocFilter.test.ts
+++ b/src/data/adHocFilter.test.ts
@@ -253,7 +253,7 @@ describe('AdHocManager', () => {
     expect(result).toContain('ResourceAttributes.cloud.region');
   });
 
-  it('apply ad hoc filter with . in column name', () => {
+  it('should apply ad hoc filter with . in column name', () => {
     const ahm = new AdHocFilter();
     const val = ahm.apply('SELECT stuff FROM foo', [
       { key: 'TABLE.key.key2', operator: '=', value: 'val' },


aangelisc:
@SpencerTorres do you have a link to any documentation detailing what the permitted characters are for databases/table/columns. This fix is okay for now but I think if we want to support this correctly we should be looking up the table and columns from the schema (though this would be inefficient for larger schemas).

src/data/adHocFilter.test.ts
```diff
@@ -252,4 +252,12 @@ describe('AdHocManager', () => {
     ] as AdHocVariableFilter[]);
     expect(result).toContain('ResourceAttributes.cloud.region');
   });
+
+  it('apply ad hoc filter with . in column name', () => {
```
adamyeats:
```suggestion
  it('should apply ad hoc filter with . in column name', () => {
```

PR #1488, opened 2025-12-20
Allow manual placement of ad-hoc filters

jkeljo:
The postgres SQL parser that this plugin uses does not support ClickHouse specific syntax (e.g. `quantile(.50)(foo)` or `INTERVAL 1 day`). Further, the plugin's logic for detecting the target table from the query can't handle CTEs. As a result, some queries will just not apply ad-hoc filters. In the former case, there's at least a SQL error in the console, but the latter case is completely silent.

This change allows users to manually place the ad-hoc filters in the SETTINGS clause of their queries using a new `$__adHocFilters` macro. This bypasses all of the automatic table detection logic, including the SQL parsing.


src/data/CHDatasource.ts
```diff
@@ -284,6 +293,23 @@ export class Datasource
...
+      return rawQuery;
+    }
+
+    // Match $__adHocFilters('table_name') or $__adHocFilters("table_name")
+    const regex = /\$__adHocFilters\s*\(\s*['"](.+?)['"]\s*\)/g;
+
+    return rawQuery.replace(regex, (match, tableName) => {
+      const filterStr = this.adHocFilter.buildFilterString(filters);
+      if (filterStr === '') {
+        return 'additional_table_filters={}';
```
bossinc, line 310: Should we still replace the macro with `'additional_table_filters={}'` if there are no adhoc filters? Would it make more sense to return `''`?

jkeljo: Typical usage of this is `SETTINGS $__adHocFilters('my_complex_table')`, so returning empty would result in illegal SQL.

jkeljo:
Conflicts fixed. Thanks for the feedback! Looking forward to having this functionality.

PR #1491, opened 2025-12-26
Add macros $__adhoc_column

### LINKED ISSUE #1490: Adhoc long query when selecting a value
**What happened**:
If a table has many rows, adding an adhoc query will execute the query
Select DISTINCT column FROM TABLE LIMIT 1000
This query can scan the entire table if there are fewer than 1000 values.
This issue has already been discussed in
https://github.com/grafana/clickhouse-datasource/issues/120
and even added the __adhocColumns macro
but it was later removed.

1. add adhoc filter by big table for column.

**Screenshots**

Image

**Environment**:

- Grafana version: 12.3.1
- Plugin version: 4.11.3
- OS Grafana is installed on: Ubuntu
- User OS & Browser: Chrome
- Others:

Smelentyev:
### Summary  
Fixes an issue where ad hoc filters fail when executing long or complex ClickHouse queries.  
Previously, ad hoc filtering could time out or mis-handle large SQL statements, especially when the filter variable contained multiple conditions.  
This change adjusts query handling to properly support long query strings by improving SQL parsing and execution logic.

### What this PR does  
- Fixes handling of ad hoc queries with long SQL expressions.  
- Includes tests ensuring correct behavior for complex filters.  

### How to test  
1. Set clickhouse_adhoc_query = 
```sql
SELECT DISTINCT $__adhoc_column
FROM database.table
WHERE $__timeFilter(timestamp)
LIMIT 1000
```
3. Create an ad hoc variable with a long `SELECT DISTINCT` query (e.g., filtering over multiple joined tables).  
4. Use that variable in dashboard panels.  
5. Verify that filters are correctly applied and queries execute without timeout or truncation.  

### Related issues  
Fixes #1490 

### Notes  
This change should enhance stability and user experience for dashboards using ad hoc filters on large tables.


Smelentyev:
The problem is described in https://github.com/grafana/clickhouse-datasource/issues/1490. 
We couldn't solve it using variables because when the list of possible options is generated, we don't know which column was selected in the ad hoc filter.

If you could suggest how to implement this using existing tools, that would be great.

PR #1513, opened 2026-01-07
Log HTTP headers as query_log.log_comment

borismattijssen:
## Summary

  Add support for logging HTTP headers as JSON in ClickHouse `query_log` via the `log_comment` field. This enables better observability by correlating queries with their originating Grafana context (dashboard, panel, alert rule, etc.).

  ## Changes

  ### Frontend
  - Added "Log Headers as Comment" toggle
  - Added additional "Header Whitelist Regex" input to control which headers are logged (default: `(?i)^(x-dashboard|x-panel|x-rule)`) 

  ### Backend
  - Serialize whitelisted headers as JSON and pass via ClickHouse's `log_comment` setting in `pkg/plugin/driver.go`

  ### Security Considerations
  Headers are filtered by regex to prevent logging sensitive data. For example, `X-Grafana-Id` contains user JWT tokens and should be excluded from the whitelist pattern. The default pattern only captures dashboard/panel/rule context headers.

  ## Implementation Evolution
  1. **Commit 1** (9a565f4a24a121cd25a64822bcbbd9efb35d07f8): Added FE&BE logic for logging headers as log_comment (header whitelist is hard-coded)
  2. **Commit 2** (dd7d89102a4d8e80947d035b551eeaa4b0eeca11): Make header whitelist user configurable from the FE

  ## Test Plan
  1. Create a new data source with the setting enabled
  2. Create a dashboard with the following query:
  ```sql
  SELECT
      log_comment AS lc,
      query AS q,
      *
  FROM system.query_log
  ORDER BY event_time DESC
  LIMIT 20
```
  3. Observe the log comment contains the filtered headers as JSON

### DIFF
--- a/src/components/configEditor/HttpHeadersConfig.tsx
+++ b/src/components/configEditor/HttpHeadersConfig.tsx
@@ -103,10 +103,7 @@ export const HttpHeadersConfig = (props: HttpHeadersConfigProps) => {
         />
       </Field>
       {logHeadersAsComment && (
-        <Field
-          label={labels.logHeadersAsCommentRegex.label}
-          description={labels.logHeadersAsCommentRegex.tooltip}
-        >
+        <Field label={labels.logHeadersAsCommentRegex.label} description={labels.logHeadersAsCommentRegex.tooltip}>
           <Input
             data-testid={selectors.logHeadersAsCommentRegexInput}
             value={logHeadersAsCommentRegex}
--- a/src/views/CHConfigEditor.tsx
+++ b/src/views/CHConfigEditor.tsx
@@ -71,7 +71,12 @@ export const ConfigEditor: React.FC<ConfigEditorProps> = (props) => {
   const onSwitchToggle = (
     key: keyof Pick<
       CHConfig,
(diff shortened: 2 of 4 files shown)


aangelisc:
Hi @borismattijssen - thank you for the contribution here 😊 

> Maybe it's time the plugin creates its own table as a complimentary addition to the system.query_log. 

I'm not entirely anti this approach but we strongly discourage users from configuring data sources with write access to their databases. The potential risk of a malicious user modifying data is unavoidable if the data source has write access. That being said, if there is no appetite to expand the `query_log` table with a `metadata` column then perhaps we can explore this option instead. However, we'd have to be considerate of any performance impacts this may have. If we have a substantial number of users issuing queries in parallel, then we'd effectively double the number of queries we're issuing to a downstream DB (reads and writes).

Alternatively, we could extend @SpencerTorres's approach with the user information to also include dashboard & panel in the client info section. Thoughts on this?

borismattijssen:
> Alternatively, we could extend @SpencerTorres's approach with the user information to also include dashboard & panel in the client info section. Thoughts on this?

This would also work for us! Most important for us is to include the IDs I mentioned earlier (and preferably the titles, but this is a nice to have).

borismattijssen:
Hi @SpencerTorres , I was wondering what your current thoughts are on this topic? Do you think we can work towards an implementation either through this PR, the client_info field, or with #1433?

PR #1521, opened 2026-01-12
fix: set FrameTypeLogLines when body field exists in logs

ku524:
## Summary

Set `FrameTypeLogLines` metadata when a `body` field exists in log query results, ensuring Grafana correctly identifies the log message column.

## Motivation

Grafana's logs panel relies on frame metadata to determine which field contains the log message. Without `FrameTypeLogLines`, Grafana falls back to using the second string field as the log line, which can lead to incorrect column selection when the `body` field is present but not explicitly marked.

This issue particularly affects sql editor users, where:
- The **Log Message column** setting (from datasource config) is aliased as `body` in SQL
- Without proper metadata, Grafana may ignore the `body` field and use another string column
- Users see incorrect log messages in the logs panel

### Backend (`pkg/plugin/driver.go`)
- Modified `MutateResponse()` to detect `body` field in logs frames
- Set `frame.Meta.Type = data.FrameTypeLogLines` when `body` field exists
- Added `frameHasField()` helper function to check field existence

### Tests (`pkg/plugin/driver_test.go`)
- Added `TestFrameHasField()` for field detection logic
- Added `TestMutateResponseLogsFrameType()` covering:
  - ✅ Sets `FrameTypeLogLines` when body field exists
  - ✅ Does not set when body field is missing
  - ✅ Does not affect non-logs visualizations

### Code Quality
- Applied prettier formatting to frontend files

## Behavior

| Scenario | Result |
|----------|--------|
| Query Builder + Log Message column configured | `body` field → `FrameTypeLogLines` set ✅ |
| SQL Editor with `SELECT body` | `body` field → `FrameTypeLogLines` set ✅ |
| SQL Editor with `SELECT col AS body` | `body` field → `FrameTypeLogLines` set ✅ |
| SQL Editor with `SELECT message` (no alias) | No `body` field → Falls back to Grafana default behavior |
| Table/Traces visualization | No change (only affects logs) |

**Note**: SQL Editor users should either select a column named `body` or use `AS body` alias for the log message column to benefit from this feature. Query Builder handles this automatically.

## Test Plan

- [x] `go test ./pkg/plugin/...` passes (all tests including new ones)
- [x] `npm run typecheck` passes
- [x] `npm run lint` passes

## Related

This complements the existing `ColumnHint.LogMessage` system:
- **Frontend**: Maps configured column to `body` alias in SQL
- **Backend**: Detects `body` field and marks frame appropriately
- **Grafana**: Uses frame metadata to display log lines correctly

### DIFF
--- a/pkg/plugin/driver.go
+++ b/pkg/plugin/driver.go
@@ -371,9 +371,16 @@ func (h *Clickhouse) MutateQuery(ctx context.Context, req backend.DataQuery) (co
 
 // MutateResponse converts fields of type FieldTypeNullableJSON to string,
 // except for specific visualizations (traces, tables, and logs).
+// For logs visualization, it also sets FrameTypeLogLines when a "body" field is present.
 func (h *Clickhouse) MutateResponse(ctx context.Context, res data.Frames) (data.Frames, error) {
 	for _, frame := range res {
 		if frame.Meta.PreferredVisualization == data.VisTypeLogs {
+			// Set FrameTypeLogLines if body field exists, so Grafana uses it as log line
+			// instead of falling back to the second string field
+			if frameHasField(frame, "body") {
+				frame.Meta.Type = data.FrameTypeLogLines
+			}
+
 			err := mergeOpenTelemetryLabels(frame)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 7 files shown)


ku524:
Hmm,,,I already apply this change into my local env, so it's hard to reproduce. I will reopen later when I reproduce. Thank you for your review!

Issue #807, opened 2024-04-19
Scan error on Nullable( Enum(..) ): converting NULL to string is unsupported


sergey-v9:
For the nullable enum column (in this case, 'status') Query Builder fails to select the column as is and displays the error message:
`sql: Scan error on column index 2, name "status": converting NULL to string is unsupported: Could not process SQL results`

As a workaround, this can be fixed by manually wrapping the enum column with `toString()` function in SQL Editor.

Sample table to reproduce: 
```sql
CREATE TABLE test_nullable_enum
(
    timestamp DateTime64, 
    id String(64),
    status Nullable(
        Enum(
            'Success' = 1,
            'Fail' = 2
        )
    )
) ENGINE = MergeTree
ORDER BY id;

INSERT INTO test_nullable_enum (timestamp, id, status) VALUES ('2024-04-18 00:00:00.000', 'i0', NULL);
INSERT INTO test_nullable_enum (timestamp, id, status) VALUES ('2024-04-18 11:11:11.000', 'i1', 'Success');
INSERT INTO test_nullable_enum (timestamp, id, status) VALUES ('2024-04-18 22:22:22.000', 'i2', 'Fail');

SELECT * FROM test_nullable_enum;

```

**Screenshots**

![image](https://github.com/grafana/clickhouse-datasource/assets/4435150/da7662a4-aea3-4337-bc94-451a3ab8a64a)

![image](https://github.com/grafana/clickhouse-datasource/assets/4435150/dc823996-45b6-45d6-bbfc-3f1bd02458e4)

![image](https://github.com/grafana/clickhouse-datasource/assets/4435150/3833167e-9d22-47d1-a2c9-4757b05bf075)

**Environment**:

- Grafana version: v10.4.2 (701c851be7) Docker image grafana/grafana-oss:10.4.2
- Plugin version: 4.0.6
- OS Grafana is installed on: linux/amd64 (Docker for Windows)
- User OS & Browser: Windows 11, Google Chrome 124

PR #1523, opened 2026-01-12
fix(converters): add support for Nullable(Enum8/16) column types

sanjams2:
== Motivation ==

Fix scan errors when querying columns with Nullable(Enum8) or Nullable(Enum16) types containing NULL values

== Details ==

Fixes: https://github.com/grafana/clickhouse-datasource/issues/807

The plugin was missing converter definitions for Enum types, causing queries on Nullable(Enum8/16) columns to fail with "converting NULL to string is unsupported" when NULL values were present. This happened because without a matching converter, the framework fell back to default handling which couldn't properly scan NULL into a non-pointer string destination.

The fix adds regex-based converters for both Enum8 and Enum16 (and their Nullable variants) following the same pattern as other types like Nullable(String). Non-nullable enums use *string scan type, nullable enums use **string to allow the outer pointer to be nil for NULL values.


bossinc (approved):
Thank you for this PR! It looks great

SpencerTorres (approved):

PR #1530, opened 2026-01-16
Bump Go version because of CVEs.

JiriCtvrtka:
Bump Go version because of CVEs.


aangelisc:
Thank you for the contribution @JiriCtvrtka but we have now updated to 1.26.0 in #1670

PR #1545, opened 2026-01-24
improvements to query overview panel in query analysis dashboard

janeklb:
## Type of Change

- [x] 🐛 Bug Fix

---

### What is the bug?

- Clicking the header of the "Memory usage" column sorts the rows lexically (rather than by actual bytes quantity)
- The "Duration" column's unit is not using abbreviations, causing the actual valuable information (the amount) to overflow out of the cell width

image

### Changes

- update query to return `memory_usage` value directly; delegate formatting to Grafana via `unit` field property override
- update `unit` property override for `"Duration"` to be `ms`

## Screenshots / Videos

(After)

image

### DIFF
--- a/src/dashboards/query-analysis.json
+++ b/src/dashboards/query-analysis.json
@@ -1081,12 +1081,16 @@
           {
             "matcher": {
               "id": "byName",
-              "options": "memory usage"
+              "options": "memory_usage"
             },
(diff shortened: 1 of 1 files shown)


bossinc (approved):
Thank you for fixing this! Looks great

PR #1560, opened 2026-02-04
Add separate columns for Resource/Scope/Log Attributes

SpencerTorres:
This PR removes the previous `LogLabels` hint in favor of separate `ResourceAttributes`, `ScopeAttributes`, and `LogAttributes` hints. This makes it easier to find these columns in our datasource logic. It also helps prevent bugs caused by label merging.

These 3 columns are merged under one `labels` field, which has previously caused issues for quick filtering from the logs panel, as well as breaking logs context. With these changes these issues are fixed.

Note that this means the log labels field is removed from the query builder. Most users are using an OTel-like schema, which will benefit from the labels merging. For those who have a non-standard schema, they can still select a field and use `labels` as an alias to get the previous behavior.

removed Log Labels dropdown

Considering we're removing a field from the builder, let me know if you have any concerns about backwards compatibility or test coverage.


src/data/CHDatasource.ts
```diff
@@ -325,7 +325,7 @@ export class Datasource
     }
 
     let columnName = action.options.key || '';
-    const actionFrame: DataFrame | undefined = (action as any).frame;
+    // const actionFrame: DataFrame | undefined = (action as any).frame;
```
bossinc:
```suggestion
```

bossinc (approved):
Nice! works great 🎉

PR #1630, opened 2026-02-12
Fix additional settings config bug

### LINKED ISSUE #1555: Ensure additional settings section is collapsed when no changes have been made
Currently the `Additional Settings` section on the config page is expanded even when no changes have been made. The section should be collapsed on data source creation and if no changes have been made.

alyssajoyner:
The Additional settings section should be collapsed by default when creating a new data source or viewing an existing one, unless the user has actually modified additional settings. This PR fixes a bug where the section was incorrectly opening by default even when no changes had been made.

Fixes #1555

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
     "@grafana/runtime": "12.4.1",
     "@grafana/schema": "12.4.1",
     "@grafana/ui": "12.4.1",
+    "@openfeature/web-sdk": "^1.7.3",
     "es-toolkit": "^1.45.1",
     "js-sql-parser": "^1.6.0",
     "pgsql-ast-parser": "^12.0.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1603,6 +1603,11 @@
   resolved "https://registry.yarnpkg.com/@openfeature/react-sdk/-/react-sdk-1.2.1.tgz#76691248eaee4e8660282852571591e47e5c1313"
   integrity sha512-W4vRe76HVB/PkCJQGycllIcHCT4q3C++8wFXxL3XZC3VWm7NEnFAzmboXfsMg/6N9KNWh0KRJYQAj0XTWAiyww==
 
+"@openfeature/web-sdk@^1.7.3":
+  version "1.7.3"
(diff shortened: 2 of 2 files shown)


src/views/config-v2/AdditionalSettingsSection.tsx
```diff
@@ -153,8 +158,8 @@ export const AdditionalSettingsSection = (props: Props) => {
       jsonData.maxOpenConns ||
       jsonData.queryTimeout ||
       jsonData.validateSql ||
-      !shallowSettingsCompare(jsonData.logs, defaultCHAdditionalSettingsConfig.logs) ||
-      !shallowSettingsCompare(jsonData.traces, defaultCHAdditionalSettingsConfig.traces) ||
+      areDifferent(jsonData.logs, defaultCHAdditionalSettingsConfig.logs) ||
+      areDifferent(jsonData.traces, defaultCHAdditionalSettingsConfig.traces) ||
```
adamyeats: I would replace this by removing the `areDifferent` function, doing `import { isEqual } from 'es-toolkit/compat';` (after running `npm install es-toolkit --save`), and then replacing these lines like so:

```suggestion
      !isEqual(jsonData.logs, defaultCHAdditionalSettingsConfig.logs ?? {}) ||
      !isEqual(jsonData.traces, defaultCHAdditionalSettingsConfig.traces ?? {}) ||
```

You've basically doing a shallow equality comparison here. Re-implementing this algorithm introduces the possibility of some obscure bugs - I'd rely on some well-known library code here.  Wdyt?

alyssajoyner: I did not know this existed - thank you @adamyeats!

package.json
```diff
@@ -90,13 +90,16 @@
...
-    "semver": "^7.7.3",
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "react-router-dom": "5.3.4",
+    "semver": "^7.7.3",
     "sql-formatter": "^15.6.12",
     "tslib": "2.8.1"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
adamyeats, line 104:
```suggestion
  }
```

We shouldn't need this, right? We use `npm` in this repo.

PR #1663, opened 2026-02-17
Skip OTel trace time range optimization when trace timestamp table does not exist

karl-power:
## Changes

Fixes an issue where trace ID / trace details lookups were failing because a `{traces table name}_trace_id_ts` table did not always exist.

- Adds a new field to the `QueryBuilderOptions` meta called `hasTraceTimestampTable` (optional boolean). When this value is `true` (and the existing conditions are met), then the trace ID lookup will use the trace ID time range optimization.
- Adds a check to see if this `{traces table name}_trace_id_ts` table exists inside the `TraceQueryBuilder` component, and passes the value into the SQL generator via the query options.
 
### Before
Screenshot 2026-02-17 at 17 27 52

#### When optimization table exists (extra query parts in red)
Screenshot 2026-02-17 at 17 17 49

#### Without optimization table (above highlighted query parts are not added)
Screenshot 2026-02-17 at 17 18 15

### DIFF
--- a/src/components/configEditor/LogsConfig.test.tsx
+++ b/src/components/configEditor/LogsConfig.test.tsx
@@ -14,6 +14,7 @@ describe('LogsConfig', () => {
         onDefaultTableChange={() => {}}
         onOtelEnabledChange={() => {}}
         onOtelVersionChange={() => {}}
+        onFilterTimeColumnChange={() => {}}
         onTimeColumnChange={() => {}}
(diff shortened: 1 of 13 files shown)


SpencerTorres (approved):
Flawless. Exactly what we needed 🚀

aangelisc (approved):
Thank you for the contribution @karl-power!

Issue #1687, opened 2026-02-21
Optimize multi-timestamp `ORDER BY`


SpencerTorres:
The `ORDER BY` for `TimestampTime` and `Timestamp` must be wrapped in parenthesis to make the best use of the table's primary key. This is based on changes from https://github.com/grafana/clickhouse-datasource/pull/1642

We need to modify the logs SQL generator to look for both columns and override the ORDER BY. We don't have any code to pair order by columns like this, so it will likely need to be a special case.

PR #1695, opened 2026-02-25
Support grouping orderBy columns by their column hint values

karl-power:
Fixes [#1687](https://github.com/grafana/clickhouse-datasource/issues/1687)

Follow-up to https://github.com/grafana/clickhouse-datasource/pull/1642

## Changes
- Adds an optional `hintsToGroup?: Set<ColumnHint>` argument to the `getOrderBy` of the SQL generator. When set, any ORDER BY entries whose hint is in the set are combined into a single ORDER BY term instead of separate ones.

- Columns whose hint is in the set are emitted as one grouped expression, e.g. `(TimestampTime, Timestamp) DESC`, using the direction of the **first** grouped column.

- The logs query now calls getOrderBy with `hintsToGroup = new Set([ColumnHint.FilterTime, ColumnHint.Time])` so time columns are ordered together.

- The order that the grouped columns appear in the options.orderBy items is what determines their order in the resulting group. The order of column hints in the `hintsToGroup` Set has no impact on grouping order.


SpencerTorres (approved):
Awesome, I like the flexibility of this solution 👌

PR #1769, opened 2026-04-10
Add config validation

alyssajoyner:
- [x] Merge after: https://github.com/grafana/grafana/pull/122733 

Adds inline field validation to the ClickHouse v2 config editor behind the `clickHouseConfigValidation` feature toggle. When enabled, the following fields are validated on save and display inline errors:

- **Server address** (`ServerAndEncryptionSection`)
- **Port** (`ServerAndEncryptionSection`)
- **Username** (`DatabaseCredentialsSection`)

When validation fails, Grafana's native datasource testing status banner surfaces the errors — no save reaches the backend until all required fields are filled. 
Without this, an empty host/port would reach the ClickHouse backend and return a generic connection error. Inline errors give users immediate, field-level feedback before any network request is made.

### Notes

- Test this feature by setting `clickHouseConfigValidation`= true and `newClickhouseConfigPageDesign` = true
- When `clickHouseConfigValidation = false` (default), behaviour is unchanged — `validation` is `undefined`, no validators register, save proceeds as before.
- Validation only applies to the v2 config editor (`newClickhouseConfigPageDesign = true`).
- Requires the companion grafana/grafana PR that wires `DataSourceConfigValidationAPI` into `EditDataSource` and passes the `validation` prop to plugin config editors.


src/views/config-v2/DatabaseCredentialsSection.tsx
```diff
@@ -66,6 +90,8 @@ export const DatabaseCredentialsSection = (props: Props) => {
               </>
             }
             required
+            invalid={!!fieldErrors.username}
+            error={fieldErrors.username}
```
itsgareth, line 94: should this also have `onBlur={}` similar to server address and server port?

as it stands, the username error is only displayed on save. where the other errors are populated on save and on blur.

itsgareth (changes requested):
LGTM, just one suggested change.

src/views/config-v2/DatabaseCredentialsSection.tsx
alyssajoyner: good catch - thank you!

itsgareth (approved):
LGTM

PR #1775, opened 2026-04-11
Fix SQL mode showing each table entry twice

alex-fedotyev:
## Type of Change

- [ ] 🚀 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧹 Refactor / Chore

---

### What is the bug?

In SQL mode, each table entry appears twice (or more) in the autocomplete suggestions. `registerSQL()` in `sqlProvider.ts` calls `monaco.languages.registerCompletionItemProvider()` and `registerDocumentFormattingEditProvider()` every time the SQL editor mounts, but never disposes of the previous registrations. Since Monaco stacks providers rather than replacing them, each editor remount (e.g. switching between Builder and SQL mode, or navigating between panels) adds another provider returning the same suggestions, causing duplicate table entries that multiply with each remount.

### How to reproduce

1. Open a ClickHouse datasource query editor and switch to SQL mode
2. Start typing a `FROM` clause to trigger table autocomplete — observe suggestions
3. Switch to Builder mode, then back to SQL mode
4. Trigger table autocomplete again — each table now appears twice
5. Repeat switching — duplicates multiply with each remount

### Related Issues

Closes HDX-3957

---

## Please check that:

This is a minimal, targeted fix. The root cause is that Monaco's `registerCompletionItemProvider` and `registerDocumentFormattingEditProvider` return `IDisposable` objects that must be disposed to remove the provider. Without disposal, providers accumulate and each one independently returns the full set of suggestions, leading to duplicates.

`registerSQL()` now returns an `SQLRegistration` object that captures both disposables and exposes a `dispose()` method. `SqlEditor` stores this in a ref and calls `dispose()` in `onEditorWillUnmount`.

All 64 existing test suites (592 tests) continue to pass.

### DIFF
--- a/src/components/SqlEditor.test.tsx
+++ b/src/components/SqlEditor.test.tsx
@@ -19,8 +19,8 @@ const mockMonaco = {
     getLanguages: () => [],
     register: jest.fn(),
     setMonarchTokensProvider: jest.fn(),
-    registerCompletionItemProvider: jest.fn(),
-    registerDocumentFormattingEditProvider: jest.fn(),
+    registerCompletionItemProvider: jest.fn(() => ({ dispose: jest.fn() })),
+    registerDocumentFormattingEditProvider: jest.fn(() => ({ dispose: jest.fn() })),
   },
 };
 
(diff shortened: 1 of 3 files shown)


adamyeats (changes requested):
@alex-fedotyev Thanks for this! As these are UI changes, could we have some E2E tests for these?

alex-fedotyev:
Closing in favor of #1820, which landed last week with the same fix plus a defensive dedupe in suggestions.ts and full unit + E2E coverage. Thanks @adamyeats for the more comprehensive take.

PR #1776, opened 2026-04-13
Add error categorization to config

alyssajoyner:
Previously, all connection failures showed the message `"failed to create ClickHouse client"`, giving users no actionable information about what went wrong.

This PR:

- Adds `connection_error.go` which classifies connection errors into `auth`, `network`, `tls`, `timeout`, `config`, `server`, and `unknown` using typed error inspection (`clickhouse.Exception` codes, `net.OpError`, x509 types) with HTTP status code parsing as a fallback for errors the clickhouse-go HTTP transport surfaces as plain strings. Covers both native and HTTP protocols.
- Returns the original error message prefixed with its category (e.g. `[network] dial tcp: lookup clickhouse-eks: no such host`) instead of the opaque string. The full error is never logged to avoid PII exposure — only the category is logged server-side.
- Shows a user-friendly hint alongside the technical error. Includes a special case for the common TLS misconfiguration where the server doesn't require a secure connection.
- Fires a `grafana_ds_clickhouse_healthcheck_failed` RudderStack event with `error_category` and `protocol` on each health check failure.


src/tracking.ts
```diff
@@ -6,6 +6,10 @@ export const trackClickhouseDashboardLoaded = (props: ClickhouseDashboardLoadedP
   reportInteraction('grafana_ds_clickhouse_dashboard_loaded', props);
 };
 
+export const trackClickhouseHealthCheckFailed = (props: { error_category: string; protocol: string }) => {
+  reportInteraction('grafana_ds_clickhouse_healthcheck_failed', props);
```
kevinwcyu, line 10: nit:

```suggestion
  reportInteraction('grafana_ds_clickhouse_healthcheck_failed', { error_category, protocol });
```

explicitly construct the payload to prevent accidentally exposing more information than required.

pkg/plugin/driver.go
kevinwcyu, line 168: nit, this could be extracted into a separate function e.g. `wrapCategorizedConnectionError` since it's also used below in the `PingContext` check.

also in the tls config block i believe it could be used as well:

```go
		tlsConfig, err = getTLSConfig(settings)
		if err != nil {
			return nil, wrapCategorizedConnectionError(err)
		}
```

right now i get a uncategorized tls error when i put in an invalid value in the tls settings:

Image

Image

so the tracking would return this as an `unknown` error. arguably, the error i triggered could be a config error as well since i put a value that isn't in the correct format.

kevinwcyu (approved):
lgtm, just a couple of nits.

pkg/plugin/driver.go
alyssajoyner: i like this feedback a lot - applied!

kevinwcyu (approved):

PR #1778, opened 2026-04-14
fix(validation): replace `js-sql-parser` with existing ClickHouse lexer

adamyeats:
`js-sql-parser` flagged valid ClickHouse-specific syntax (`FINAL`, `PREWHERE`, `ARRAY JOIN`, `SETTINGS`, `ASOF JOIN`, `::` cast, etc.) as errors, producing false red squiggles in the SQL editor for users who had SQL validation enabled.

The codebase already contains a TypeScript port of ClickHouse's own C++ lexer (`src/ch-parser/lexer.ts`), originally added for autocomplete and covering the full ClickHouse token set including heredocs, nested block comments, hex/binary literals, and ClickHouse-specific operators. This lexer emits typed error tokens (`ErrorSingleQuoteIsNotClosed`, `ErrorMultilineCommentIsNotClosed`, etc.) and handles all ClickHouse syntax correctly by design.

### Validation switch-over

- Replaced `js-sql-parser` in `validate.ts` with a pass over the existing lexer
- Rewrote the test suite with real behavioural tests covering both valid ClickHouse syntax and genuine error cases
- Removed `js-sql-parser` from `package.json` entirely

### Lexer parity fixes

While verifying the port against ClickHouse's `src/Parsers/Lexer.cpp`, three deviations came to light and are fixed in this PR:

- **Unicode handling matches JS codepoints, not UTF-8 bytes.** The original port translated C++ byte-level checks (`E2 80 98`, `E2 88 92`) literally, so U+2018/U+2019 smart quotes and U+2212 minus only fired on mojibake input. Now matched directly as JS codepoints.
- **Heredoc tag validation.** `$tag$...$tag$` now requires the tag to consist solely of ASCII word characters, matching `Lexer.cpp`. Empty tags (`$$...$$`) remain valid.
- **Expanded keyword set.** Added ClickHouse dialect keywords to the autocomplete/parser keyword list: `PREWHERE`, `FINAL`, `SAMPLE`, `SETTINGS`, `FORMAT`, `INTERVAL`, JOIN modifiers (`ASOF`, `ANY`, `ANTI`, `SEMI`, `GLOBAL`, `LATERAL`, `ARRAY`), CTE (`RECURSIVE`), ORDER/LIMIT modifiers (`NULLS`, `FIRST`, `LAST`, `TIES`), DML verbs, common DDL, and window-function clauses.

### Performance

The validator runs on every keyup, so allocations in the lexer hot path matter. Three changes:

- **Heredoc scan no longer allocates the rest of the input per `$`.** The old code did `tokenStream.indexOf('$', 1)` on a fresh `.substring()` of the remaining input, then scanned again for the closing tag. Now it works in absolute positions against `this.text`, eliminating the O(n) allocation per heredoc attempt.
- **Lazy uppercase cache on `Token`.** `matchKeyword` and `isKeyword` were re-uppercasing the token text on every call from the parser's hot loops. The uppercase form is now lazily cached on the token.
- **Removed `.toLowerCase()` on single-char comparisons.** Number-literal parsing compared `c.toLowerCase() === 'e'` and `'p'`, allocating a new string per character. Replaced with direct case-insensitive char-code checks.

### Tests

- Unit tests: added coverage for Unicode smart quotes (both valid and unterminated), the U+2212 minus operator, and heredocs with word-char and empty tags. Full suite passes (670 tests).
- E2E regression guard (`tests/e2e/sqlValidation.spec.ts`): drives the real Monaco editor through the full pipeline (onKeyUp → validate → `setModelMarkers` → DOM) and asserts no `.squiggly-error` decorations appear for each ClickHouse-specific construct the old parser choked on. A negative control (unclosed block comment) confirms the validator is wired up, so the positive assertions can't pass vacuously. Requires `validateSql: true` in the provisioned e2e datasource.

### DIFF
--- a/cspell.config.json
+++ b/cspell.config.json
@@ -16,6 +16,7 @@
   "words": [
     "aggregatable",
     "aheads",
+    "ASOF",
     "apikey",
     "buildinfo",
     "CHSQL",
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,6 @@
         "@grafana/ui": "12.4.2",
         "@openfeature/web-sdk": "^1.7.3",
         "es-toolkit": "^1.45.1",
-        "js-sql-parser": "^1.6.0",
         "pgsql-ast-parser": "^12.0.1",
         "react": "18.3.1",
         "react-dom": "18.3.1",
@@ -12359,12 +12358,6 @@
       "resolved": "https://registry.npmjs.org/js-cookie/-/js-cookie-2.2.1.tgz",
       "integrity": "sha512-HvdH2LzI/EAZcUwA8+0nKNtWHqS+ZmijLA30RwZA0bo7ToCckjK5MkGhjED9KoRcXO6BaGI3I9UIzSA1FKFPOQ==",
       "license": "MIT"
-    },
-    "node_modules/js-sql-parser": {
-      "version": "1.6.0",
-      "resolved": "https://registry.npmjs.org/js-sql-parser/-/js-sql-parser-1.6.0.tgz",
-      "integrity": "sha512-ezHNl95OKis5TzAY180j9fHzIaMr3KDKw1Syf+ABrercb+qADqBdGbQjOb0QtdqdfE97TdF91d3nKbXgtCUwDg==",
-      "license": "ISC"
     },
     "node_modules/js-tokens": {
       "version": "4.0.0",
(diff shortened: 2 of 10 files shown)


adamyeats:
@SpencerTorres Wondering if you think this would be a viable stopgap before, or even fallback for, #1780?

bossinc (approved):
Great work! no notes. I love all the small efficiency gains.

PR #1779, opened 2026-04-14
fix(autocomplete): match column suggestions case-insensitively

adamyeats:
Column suggestion filtering used a case-sensitive `startsWith`, causing typed prefixes like `codefile` to never match columns named `CodeFile`/ Fixed by lowercasing both sides of the comparison in `fetchFieldSuggestions`.


itsgareth (approved):
LGTM

PR #1787, opened 2026-04-18
perf(schema): cache system.tables/columns introspection with singleflight

adamyeats:
## Type of Change

- [x] 🚀 Feature

---

### What is this feature?

Adds a per-datasource TTL cache (with `singleflight`-backed miss deduplication) in front of the three schema introspection resource handlers used by the query builder:

- `Tables` — `SELECT database, name FROM system.tables`
- `Columns` — `SELECT ... FROM system.columns WHERE (database, table) IN (...)`
- `ColumnValues` — `SELECT DISTINCT ... UNION ALL ...` (the heaviest of the three)

The cache lives on `SchemaProvider`, so each datasource instance has its own namespace. New JSON settings:

- `enableSchemaCache` (default **true**) — feature toggle so OSS/self-managed users can opt out if they hit something unexpected.
- `schemaCacheTTLSeconds` (default **60**) — TTL applied with ±5% jitter to avoid stampedes on first dashboard load.

The query builder fires schema resource calls on nearly every keystroke. When a dashboard with many panels opens, N panels independently fan out N identical `system.tables` / `system.columns` round-trips, hammering ClickHouse and the Grafana data source API server. `"Schema introspection on every query"` is called out as the most common justified cache use case in the Grafana MT infrastructure performance docs.

`singleflight` is the load-bearing piece: without it, a 20-panel dashboard sees 20 concurrent cold callers each paying the full round-trip. With it, exactly one caller does the work and the other 19 block on the shared result.

### Who is this feature for?

All ClickHouse datasource users — both Grafana Cloud MT and OSS/self-managed. The toggle defaults to on so the win is automatic; the toggle exists because this caches schema which some users may want disabled (e.g. during active schema migration work).

### How to test this feature

1. Build the plugin locally (`mage -v build:linux` or equivalent).
2. Open a ClickHouse datasource in Grafana with several panels using the query builder.
3. Confirm autocomplete and column pickers still work as expected.
4. In ClickHouse's `system.query_log`, confirm repeated dashboard loads within 60s no longer trigger duplicate `system.tables` / `system.columns` queries.
5. Toggle `enableSchemaCache` off in the datasource JSON settings and confirm the cache is bypassed (queries fire on every request again).

---

## Benchmarks

Apple M2 Pro, simulated 2ms ClickHouse round-trip (see `pkg/plugin/schemacache/cache_bench_test.go`):

```
BenchmarkTablesFetch_Uncached-10             2,331,574 ns/op     48 B/op    1 allocs/op
BenchmarkTablesFetch_Cached-10                      54 ns/op      0 B/op    0 allocs/op
BenchmarkTablesFetch_CachedConcurrent-10     2,300,177 ns/op  3,028 B/op   63 allocs/op
```

- **Cached hot hit**: ~43,000× faster than a ClickHouse round-trip (55 ns vs 2.3 ms), zero allocations.
- **20-panel concurrent cold load**: finishes in the cost of **one** upstream call. Without singleflight this would be ~20× slower.

---

## Implementation notes

- `pkg/plugin/schemacache/` — new generic (`Cache[V any]`) TTL cache, safe for concurrent use, with a `Do(ctx, key, fn)` method that combines cache lookup + singleflight dedup + lazy expiry. 11 unit tests (including a 50-goroutine singleflight dedup test) all pass with `-race`.
- Cache keys:
    - `Tables` uses a constant key (`"all"`) — one entry ever.
    - `Columns` uses the **sorted** joined table list, so callers asking for `[A, B]` and `[B, A]` hit the same entry.
    - `ColumnValues` uses `table + "::" + sorted(columns)` for the same reason.
- `nil` cache is the feature-disabled state; each handler has a fast path that skips the cache entirely, so the toggle is a zero-overhead bypass rather than a separate code path.
- Jitter is applied as a uniform ±5% band around the TTL on every `Set`, per the Grafana MT caching guidance.
- Headers intentionally do not participate in the cache key — they are per-request forwarded auth tokens that do not change the schema ClickHouse sees for the same datasource. If per-user schema filtering is ever needed, this assumption must be revisited.

## Please check that:

- TTL of 60s was chosen deliberately: `ColumnValues` caches DISTINCT user data (potentially sensitive row values), and 60s staleness is the longest window users would tolerate for filter-dropdown freshness. Tunable via `schemaCacheTTLSeconds` per-datasource if needed.


pkg/plugin/schemacache/cache.go
```diff
@@ -0,0 +1,191 @@
...
+	if v, ok := c.Get(key); ok {
+		return v, nil
+	}
+	result, err, _ := c.sf.Do(key, func() (any, error) {
+		// Re-check after acquiring the singleflight slot: another caller
+		// may have populated the entry while we waited.
+		if v, ok := c.Get(key); ok {
+			return v, nil
+		}
+		v, err := fn(ctx)
```
alyssajoyner:
```suggestion
		v, err := fn(context.WithoutCancel(ctx)
```

claude told me it might be worth protecting against mid-flight cancellation — if the first caller's context is cancelled while others are waiting on the singleflight, they'll all get that cancellation error even if their own contexts are still valid.

alyssajoyner (approved):
LGTM - one minor nit 😊

pkg/plugin/schemacache/cache.go
adamyeats: Good catch, resolved!

PR #1790, opened 2026-04-18
feat(queryBuilder): explain column roles in logs/traces/timeseries builders

adamyeats:
## Type of Change

- [x] 🚀 Feature
- [ ] 🐛 Bug Fix
- [x] 📝 Documentation
- [ ] 🧹 Refactor / Chore

---

### What is this feature?

In-editor guidance for the column-role slots in the Logs, Traces, and Time Series query builders. Specifically:

- Each column selector now has an enriched tooltip that explains the column's role, the SQL alias it's rewritten to (e.g. `Body → body`, `TraceId → traceID`, `Duration → duration`), a list of conventional non-OTel column names, and the OTel column name for reference.
- A new `ColumnRolesHelp` note sits at the top of each builder's column section with a one-line summary and a "Learn about column roles" link that deep-links to a new **Column roles** section in the query-editor docs.
- The docs page gains a "Column roles" section documenting each role → SQL alias → OTel / common name mapping for Logs, Time Series, and Traces.

The Query Builder requires specific column semantics (which column becomes `timestamp`, `body`, `traceID`, etc.) but gave almost no in-editor explanation of what those slots meant or what SQL aliases they produced. Non-OTel users — the majority outside vendor-preset schemas — had to guess or read the source.

### Who is this feature for?

Anyone using the Query Builder against a non-OTel ClickHouse schema — which includes most bring-your-own-schema users. OTel users also benefit from the explicit alias/convention documentation.

### How to test this feature

1. `docker compose up` from the repo root to bring up Grafana + ClickHouse with the provisioned datasource.
2. Open **Explore**, pick the ClickHouse data source, switch to **Query Builder**.
3. For each of **Logs**, **Traces**, and **Time Series** query types:
   - Confirm the "column roles" help note appears above the column selectors with a **Learn about column roles** link that opens the query-editor docs at `#column-roles`.
   - Hover each column label (Time / Message / Log Level for Logs; Trace ID / Span ID / Duration Time / etc. for Traces; Time for Time Series) and verify the tooltip mentions the SQL alias and conventional names.
4. Confirm the **Table** query type shows no help note (the roles only apply to the three typed builders).
5. Unit + E2E coverage: `npm test -- labels queryBuilder/views ColumnRolesHelp` and `npm run e2e -- columnRoles.spec.ts`.

---

### Logs query builder

**1. Column-roles help note above the column selectors** — new `ColumnRolesHelp` banner explaining that the selectors map table columns onto the roles the Logs panel expects, with a deep-link to the new "Column roles" docs section.

Logs query builder showing the new column-roles help note with a 'Learn about column roles' docs link above the Time / Message / Log Level selectors

**2. Enriched Message-column tooltip** — hovering the Message label surfaces the role description, the SQL alias it produces (`body`), common non-OTel column names, and the OTel column it maps to.

Tooltip on the Message column selector explaining the role, showing the generated SQL alias `body`, common names (message, msg, log_message), and the OTel column Body

### Traces query builder

**3. Column-roles help note in the Traces builder** — same `ColumnRolesHelp` banner rendered under the Traces builder, wired to the same docs anchor so Trace ID / Span ID / Duration Time / etc. slots are explained in one place.

Traces query builder showing the column-roles help note above the Trace ID, Span ID, Service Name, Duration and related selectors, with a 'Learn about column roles' link

**4. Enriched Duration Time tooltip** — hovering the Duration Time label explains how the selected column gets aliased to `duration` and that the **Duration Unit** setting controls the conversion to milliseconds for the trace panel (the one trace role whose semantics aren't obvious from the alias alone).

Tooltip on the Duration Time column selector explaining the role, the generated `duration` alias, and the Duration Unit setting used to normalise values to milliseconds

### Time Series query builder

**5. Column-roles help note in the Time Series builder** — shorter variant of the help note: only the Time role is required for time series, so the note makes that explicit and still links to the same docs section.

Time Series query builder showing the column-roles help note stating 'Time column is required' above the Time selector, with a 'Learn about column roles' link

**6. Enriched Time tooltip** — hovering the Time label explains the `DateTime` / `DateTime64` requirement and that the panel uses this column to order and bucket the series.

Tooltip on the Time column selector for Time Series, explaining that it must be a DateTime or DateTime64 column and that the panel uses it to order and bucket the series

---

## Please check that:

New coverage:

- `tests/e2e/columnRoles.spec.ts` — Playwright coverage for the help note, its docs link, and the per-role tooltip contents for Logs / Traces / Time Series, plus a negative assertion for Table.
- Hook/view unit tests updated where tooltip strings are asserted (`labels`, `queryBuilder/views`, `ColumnRolesHelp`).

- The Playwright helper `switchToBuilderMode` accepts an explicit `queryType` and re-clicks the radio after switching editors. This is deliberate: Grafana Explore does **not** restore `queryType` from the URL pane state for this plugin, and switching editor types resets it to **Table**. Without the re-click the help note assertions race against the default query type. Worth a glance in case we prefer a different workaround.
- The help note uses a single shared `ColumnRolesHelp` component so the three builders stay in lockstep; per-builder testids route through `Components.QueryBuilder.*.columnRolesHelp` / `...columnRolesHelpLink` so the E2E tests can target each builder independently.
- Tooltip strings pull their SQL alias from the same authoritative source the generator uses (`logColumnHintsToAlias` in `sqlGenerator.ts` for logs; hardcoded trace aliases in `generateTraceIdQuery`). If a future refactor centralises the trace aliases into a map, the tooltip copy should be re-derived from it rather than drifting.
- Scope is intentionally frontend-only. No backend / macro engine changes.

### DIFF
--- a/src/components/queryBuilder/ColumnRolesHelp.tsx
+++ b/src/components/queryBuilder/ColumnRolesHelp.tsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import { Icon } from '@grafana/ui';
+import { css } from '@emotion/css';
+
+interface ColumnRolesHelpProps {
+  text: string;
+  linkText: string;
+  href: string;
+  testIdWrapper: string;
+  testIdLink: string;
+}
+
+const styles = {
+  wrapper: css`
+    display: flex;
+    align-items: center;
+    gap: 6px;
+    margin: 4px 0 8px 0;
+    font-size: 12px;
+    opacity: 0.85;
+  `,
+  link: css`
+    text-decoration: underline;
+  `,
+};
+
+/**
+ * Small inline note rendered above the Columns section of the query builder
+ * views, giving a one-line description of the column-role concept and a link
+ * to the in-repo documentation page that enumerates the role → SQL alias
+ * mapping.
+ */
+export const ColumnRolesHelp = (props: ColumnRolesHelpProps) => {
+  const { text, linkText, href, testIdWrapper, testIdLink } = props;
+  return (
+    <div className={styles.wrapper} data-testid={testIdWrapper}>
+      <Icon name="info-circle" size="sm" />
+      <span>
+        {text}{' '}
+        <a className={styles.link} href={href} target="_blank" rel="noreferrer" data-testid={testIdLink}>
+          {linkText}
+        </a>
+      </span>
+    </div>
+  );
+};
--- a/src/components/queryBuilder/ColumnRolesHelp.tsx
+++ b/src/components/queryBuilder/ColumnRolesHelp.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
-import { Icon } from '@grafana/ui';
-import { css } from '@emotion/css';
+import { Box, Icon, Text, TextLink } from '@grafana/ui';
 
 interface ColumnRolesHelpProps {
   text: string;
@@ -9,20 +8,6 @@ interface ColumnRolesHelpProps {
   testIdWrapper: string;
   testIdLink: string;
 }
-
-const styles = {
-  wrapper: css`
-    display: flex;
-    align-items: center;
-    gap: 6px;
-    margin: 4px 0 8px 0;
-    font-size: 12px;
-    opacity: 0.85;
-  `,
-  link: css`
-    text-decoration: underline;
-  `,
-};
 
 /**
  * Small inline note rendered above the Columns section of the query builder
@@ -33,14 +18,16 @@ const styles = {
 export const ColumnRolesHelp = (props: ColumnRolesHelpProps) => {
   const { text, linkText, href, testIdWrapper, testIdLink } = props;
   return (
-    <div className={styles.wrapper} data-testid={testIdWrapper}>
-      <Icon name="info-circle" size="sm" />
-      <span>
+    <Box display="flex" alignItems="center" gap={0.5} marginTop={0.5} marginBottom={1} data-testid={testIdWrapper}>
+      <Text variant="bodySmall" color="secondary">
+        <Icon name="info-circle" size="sm" />
+      </Text>
+      <Text variant="bodySmall" color="secondary">
         {text}{' '}
-        <a className={styles.link} href={href} target="_blank" rel="noreferrer" data-testid={testIdLink}>
+        <TextLink href={href} external inline variant="bodySmall" color="secondary" data-testid={testIdLink}>
           {linkText}
-        </a>
-      </span>
-    </div>
+        </TextLink>
+      </Text>
+    </Box>
   );
 };
--- a/src/labels.ts
+++ b/src/labels.ts
@@ -404,7 +404,7 @@ export default {
       columnsHelp: {
         text: 'Map your table columns to the roles the logs panel expects. Each column is aliased in the generated SQL.',
         linkText: 'Learn about column roles',
-        href: '/docs/plugins/grafana-clickhouse-datasource/latest/query-editor/#column-roles',
+        href: 'https://grafana.com/docs/plugins/grafana-clickhouse-datasource/latest/query-editor/#column-roles',
       },
       logTimeColumn: {
         label: 'Time',
--- a/src/labels.ts
+++ b/src/labels.ts
@@ -467,7 +467,7 @@ export default {
       columnsHelp: {
         text: 'Map your table columns to the roles the traces panel expects. Each column is aliased in the generated SQL.',
         linkText: 'Learn about column roles',
-        href: '/docs/plugins/grafana-clickhouse-datasource/latest/query-editor/#column-roles',
+        href: 'https://grafana.com/docs/plugins/grafana-clickhouse-datasource/latest/query-editor/#column-roles',
       },
 
       columns: {
--- a/src/labels.ts
+++ b/src/labels.ts
@@ -401,17 +401,25 @@ export default {
       tooltip: 'Group the results by specific column',
     },
     LogsQueryBuilder: {
+      columnsHelp: {
+        text: 'Map your table columns to the roles the logs panel expects. Each column is aliased in the generated SQL.',
+        linkText: 'Learn about column roles',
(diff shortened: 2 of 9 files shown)


src/labels.ts
```diff
@@ -401,17 +401,25 @@ export default {
       tooltip: 'Group the results by specific column',
     },
     LogsQueryBuilder: {
+      columnsHelp: {
+        text: 'Map your table columns to the roles the logs panel expects. Each column is aliased in the generated SQL.',
+        linkText: 'Learn about column roles',
+        href: '/docs/plugins/grafana-clickhouse-datasource/latest/query-editor/#column-roles',
```
kevinwcyu:
```suggestion
        href: 'https://grafana.com/docs/plugins/grafana-clickhouse-datasource/latest/query-editor/#column-roles',
```

src/labels.ts
```diff
@@ -432,9 +440,15 @@ export default {
       aggregateQueryModeLabel: 'Aggregate',
       builderModeLabel: 'Builder Mode',
       builderModeTooltip: 'Switches the query builder between the simple and aggregate modes',
+      columnsHelp: {
+        text: 'The Time column is required — it anchors the series to the panel time range. Other selected columns become value series.',
+        linkText: 'Learn about column roles',
+        href: '/docs/plugins/grafana-clickhouse-datasource/latest/query-editor/#column-roles',
```
kevinwcyu:
```suggestion
        href: 'https://grafana.com/docs/plugins/grafana-clickhouse-datasource/latest/query-editor/#column-roles',
```

src/components/queryBuilder/ColumnRolesHelp.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <Icon name="info-circle" size="sm" />
+      <span>
+        {text}{' '}
+        <a className={styles.link} href={href} target="_blank" rel="noreferrer" data-testid={testIdLink}>
+          {linkText}
+        </a>
+      </span>
+    </div>
+  );
+};
```
kevinwcyu, line 33:
```suggestion
import React from 'react';
import { Box, Icon, Text, TextLink } from '@grafana/ui';

interface ColumnRolesHelpProps {
  text: string;
  linkText: string;
  href: string;
  testIdWrapper: string;
  testIdLink: string;
}

/**
 * Small inline note rendered above the Columns section of the query builder
 * views, giving a one-line description of the column-role concept and a link
 * to the in-repo documentation page that enumerates the role → SQL alias
 * mapping.
 */
export const ColumnRolesHelp = (props: ColumnRolesHelpProps) => {
  const { text, linkText, href, testIdWrapper, testIdLink } = props;
  return (
    <Box display="flex" alignItems="center" gap={0.5} marginTop={0.5} marginBottom={1} data-testid={testIdWrapper}>
      <Text variant="bodySmall" color="secondary">
        <Icon name="info-circle" size="sm" />
      </Text>
      <Text variant="bodySmall" color="secondary">
        {text}{' '}
        <TextLink href={href} external inline variant="bodySmall" color="secondary" data-testid={testIdLink}>
          {linkText}
        </TextLink>
      </Text>
    </Box>
  );
};
```

nit: remove the custom styles in favour of using the [Text](https://developers.grafana.com/ui/latest/index.html?path=/story/foundations-text--basic&args=element:p;variant:bodySmall;color:secondary) and [TextLink](https://developers.grafana.com/ui/latest/index.html?path=/story/foundations-textlink--inline&args=color:secondary;variant:bodySmall) components. The `color` prop has an option to set it to `secondary` which sets an opacity on the text color instead of the entire wrapping div.

also uses the [Box](https://developers.grafana.com/ui/latest/index.html?path=/docs/layout-box--docs) component for the layout so the spacing can be defined using Grafana spacing tokens instead of hard coding pixel values.

looks like this

Image

Image

tests/e2e/columnRoles.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+const PLUGIN_TYPE = 'grafana-clickhouse-datasource';
+
+const isCloudRun = !!process.env.GRAFANA_URL;
+
+const CLOUD_DEFAULT_UID = 'clickhouse-native-ds-m';
+const LOCAL_DEFAULT_UID = 'clickhouse-e2e';
+const DATASOURCE_UID = process.env.DS_E2E_UID || (isCloudRun ? CLOUD_DEFAULT_UID : LOCAL_DEFAULT_UID);
+
+// Doc anchor the help link points to; kept in sync with labels.ts.
+const COLUMN_ROLES_DOCS_PATH = '/docs/plugins/grafana-clickhouse-datasource/latest/query-editor/#column-roles';
```
kevinwcyu:
```suggestion
const COLUMN_ROLES_DOCS_PATH = 'https://grafana.com/docs/plugins/grafana-clickhouse-datasource/latest/query-editor/#column-roles';
```

tests/e2e/columnRoles.spec.ts
```diff
@@ -0,0 +1,232 @@
+import { expect, test } from '@grafana/plugin-e2e';
+import { Locator, Page } from '@playwright/test';
+import { QueryType } from '../../src/types/queryBuilder';
+import { Components as Selectors } from '../../src/selectors';
+
+const PLUGIN_TYPE = 'grafana-clickhouse-datasource';
+
+const isCloudRun = !!process.env.GRAFANA_URL;
+
+const CLOUD_DEFAULT_UID = 'clickhouse-native-ds-m';
```
kevinwcyu, line 10: question: in the `queryEditor.spec` it has this set as `const CLOUD_DEFAULT_UID = 'cf1uvcf1yrz0ge';` and the note there says
> // CLOUD_DEFAULT_UID points at `[managed_data_source] - ClickHouse Native (PDC)`

should both of them point to the same uid?

kevinwcyu (approved):
lgtm, just a couple nits.

tests/e2e/columnRoles.spec.ts
adamyeats: You're right! And this is correct on `main`, the branch needed a rebase so both should now be `clickhouse-native-ds-m`. Good catch!

Issue #1424, opened 2025-10-27
JSON type support for Logs and Traces


arnitolog:
**What happened**:
o11y data is collected and pushed to the Clickhouse instance through otel collector with JSON type support:`enable_json_type=1`. Clickhouse plugin in Grafana is configured with custom settings:

<img width="545" height="115" alt="Image" src="https://github.com/user-attachments/assets/56c09e4d-cbbd-47fc-b640-e086dc65da11" />

Logs work good, but for traces, when I'm trying to open any trace, I'm getting an error:
```
[...]

PR #1792, opened 2026-04-19
fix(queryBuilder): support JSON column filters and JSON-typed trace tags

### LINKED ISSUE #1281: "filter for value" do not support json type
"filter for value" do not support json type

see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40789#issuecomment-2983171103

### LINKED ISSUE #1321: Cannot query single trace when using JSON column for ResourceAttributes and ScopeAttributes
**What happened**:

Error occurred when query single trace

Image

Should be able to query and display the single trace

1.  Set query type as Traces
2. Query for recent traces
3. Pick one trace, click on it

The `otel_traces` table schema is the same as https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/clickhouseexporter/internal/sqltemplates/traces_json_table.sql

**Environment**:

- Grafana version:  12.0.2
- Plugin version: 4.10.1
- OS Grafana is installed on: K8s
- User OS & Browser: MacOs, Chrome
- Others:

adamyeats:
## Summary

- Adds `isJsonType()` type guard so JSON columns can be detected alongside Map/String/etc. rather than falling through to the string catch-all.
- Fixes filter SQL generation for JSON sub-columns (closes #1281): after JSON path navigation in `getFilters()`, the column is now cast to `Nullable(String)` and `type` set to `'String'`, so `IN` / `NOT IN` / `IsEmpty` / `IsNotEmpty` route through the correct filter-part builders instead of stringifying arrays or skipping the value branch.
- Restricts the operator list for JSON columns in `FilterEditor` to the set that actually works on JSON paths (excludes `IsEmpty` / `IsNotEmpty`, which are unreliable on JSON sub-columns).
- Handles JSON-typed `traceTags` / `traceServiceTags` in `generateTraceIdQuery()` — previously the generator unconditionally emitted Map `mapKeys(...)` syntax, which crashes on `JSON` columns (closes #1321). Partial fix for #1424.
- **New E2E regression guard** validates the generated SQL shape against a real ClickHouse instance (fixture table with a `JSON` column + `jsonFilter.spec.ts` covering Equals / IN / nested path / LIKE).

## Why `::Nullable(String)` and not plain `::String`?

Writing the E2E tests surfaced a runtime bug that the unit tests couldn't have caught: ClickHouse JSON path extraction returns `Dynamic`, and `IN` / `NOT IN` reject `Dynamic` with `ILLEGAL_TYPE_OF_ARGUMENT`. The obvious fix is a `::String` cast — but that quietly breaks `IS NULL`, because `::String` maps a missing JSON key to an empty string rather than null. `::Nullable(String)` satisfies both constraints and is a no-op for Equals / NotEquals / LIKE.

## Scope

- Does NOT enable `fetchPathsForJSONColumns()` (still disabled for perf reasons).
- Does NOT broaden the catch-all `isStringType` in `utils.ts`.
- No new filter types — JSON piggybacks on the existing `StringFilter` / `MultiFilter` shapes since path values are strings.

### DIFF
--- a/src/components/queryBuilder/FilterEditor.tsx
+++ b/src/components/queryBuilder/FilterEditor.tsx
@@ -262,6 +262,23 @@ export const FilterEditor = (props: {
           FilterOperator.OutsideGrafanaTimeRange,
         ].includes(f.value!)
       );
+    } else if (isJSONType) {
+      // JSON sub-column values are strings; exclude IsEmpty/IsNotEmpty which are unreliable on JSON paths
+      return filterOperators.filter((f) =>
+        [
+          FilterOperator.IsAnything,
+          FilterOperator.Equals,
+          FilterOperator.NotEquals,
+          FilterOperator.Like,
+          FilterOperator.NotLike,
+          FilterOperator.ILike,
+          FilterOperator.NotILike,
+          FilterOperator.In,
+          FilterOperator.NotIn,
+          FilterOperator.IsNull,
+          FilterOperator.IsNotNull,
+        ].includes(f.value!)
+      );
     } else {
       return filterOperators.filter((f) =>
         [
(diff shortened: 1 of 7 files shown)


akkikumar72 (approved):
LGTM

Issue #1461, opened 2025-11-28
Plugin scans entire table to retrieve JSON paths, causing high ClickHouse CPU usage


Rajin9601:
**What happened**:

I connected clickhouse that contains Terrabytes of logs which has JSON column. 
When i used grafana, clickhouse cpu was heavily used. 

So when i looked at the clickhouse's slow query log, i noticed grafana is requesting this query.

```
SELECT arrayJoin(distinctJSONPathsAndTypes(LogAttributes)) FROM "otel"."otel_logs" WHERE SETTINGS max_execution_time=10
```

This query scans all rows of otel_logs table. (for my case, it timed out because of 10s of max_execution_time)

This is problem because it uses cpu of clickhouse heavily.

**What you expected to happen**:

The following code determines the available JSON paths:

https://github.com/grafana/clickhouse-datasource/blob/3e357e9d905107423397ec2a7199a64bf4cb3960/src/data/CHDatasource.ts#L591-L600

I think the plugin should provide a setting that allows users to configure how JSON paths are retrieved. We need a discussion about the best way to make this behavior configurable.

Here are some possible options I’ve considered:

1. Add a configurable LIMIT

```
SELECT arrayJoin(distinctJSONPathsAndTypes(LogAttributes)) FROM (
    SELECT LogAttributes from "otel"."otel_logs" limit 
        1000 -- configurable limit
) SETTINGS max_execution_time=10
```

2. Add a configurable WHERE condition

```
SELECT arrayJoin(distinctJSONPathsAndTypes(LogAttributes)) FROM (
    SELECT LogAttributes from "otel"."otel_logs" where 
        Timestamp > now() - INTERVAL 1 hour -- configurable where statement
) SETTINGS max_execution_time=10
```

This is more flexible, but more complicated since the timestamp column name may differ between tables.

3. Simple on/off

If disabled, the plugin won’t query for JSON paths at all.

**Environment**:

- Grafana version: 12.2.1
- Plugin version: 4.11.2
- OS Grafana is installed on:
- User OS & Browser:
- Others:

andaag:
You should talk to the hyperdx team about this, we reported the same thing to them.

In addition max_execution_time is best effort, we've seen these queries hang for 10 minutes plus on our cluster.

ashraf133:
i suggest to make the query builder optional so we can activate it or no

Rajin9601:
[...]

Issue #1063, opened 2024-11-23
Adhoc filter doesn't not work in log view


zfy0701:
**What happened**:

When use filter value in log view
![Image](https://github.com/user-attachments/assets/8ceea216-034f-475b-9d2e-c08c29504fa3)
it construct wrong query filter key that doesn't work,
e.g.
![Image](https://github.com/user-attachments/assets/9901afb2-5b80-4cf8-9b5a-bfe6b3e38235)

**What you expected to happen**:

Either ad-hoc filter should convert the label filter in query, or the UI should convert label filter into proper adhoc query key.

**How to reproduce it (as minimally and precisely as possible)**:

1. go to log view, with default otel column setup
2. click filter for value

**Screenshots**

**Anything else we need to know?**:

**Environment**:

- Grafana version:  11.3.1
- Plugin version: 4.5.0
- OS Grafana is installed on: k8s
- User OS & Browser: chrome/safari
- Others:

SpencerTorres:
Hey thanks for submitting this issue. Ad-hoc filtering is always a challenge since everyone's tables and data look different. Can you provide the query, some sample data, and your table schema? I see you mention the default otel schema, but I'd like to see how you have the query builder set up, as well as what the actual SQL is. With this information I will be able to see how these filters are added for your use case.

Thanks!

zfy0701:
everything pretty standard
schema:

```
CREATE TABLE default.otel_logs
(
    `Timestamp` DateTime64(9) CODEC(Delta(8), ZSTD(1)),
    `TimestampTime` DateTime DEFAULT toDateTime(Timestamp),
    `TraceId` String CODEC(ZSTD(1)),
    `SpanId` String CODEC(ZSTD(1)),
    `TraceFlags` UInt8,
    `SeverityText` LowCardinality(String) CODEC(ZSTD(1)),
    `SeverityNumber` UInt8,
    `ServiceName` LowCardinality(String) CODEC(ZSTD(1)),
    `Body` String CODEC(ZSTD(1)),
    `ResourceSchemaUrl` LowCardinality(String) CODEC(ZSTD(1)),
    `ResourceAttributes` Map(LowCardinality(String), String) CODEC(ZSTD(1)),
    `ScopeSchemaUrl` LowCardinality(String) CODEC(ZSTD(1)),
    `ScopeName` String CODEC(ZSTD(1)),
    `ScopeVersion` LowCardinality(String) CODEC(ZSTD(1)),
[...]

Issue #798, opened 2024-04-15
Adhoc filters don't support columns with a dot in the name


Hexta:
**What happened**:
Adhoc filters don't work with columns with a dot in the name.

**What you expected to happen**:
Adhoc filters work with columns with a dot in the name.

**How to reproduce it (as minimally and precisely as possible)**:
1. Create a table with `a.x` column.
2. Add an adhoc filter for `a.x` column.

**Screenshots**

**Anything else we need to know?**:

**Environment**:

- Grafana version: 10.4.1
- Plugin version: 
- OS Grafana is installed on:
- User OS & Browser:
- Others:

aangelisc:
Hi @Hexta, I can confirm I've replicated this issue. I'll add it to our backlog to investigate 😊

adamyeats:
Fixed in #1793. Please re-open if the issue still persists.

PR #1793, opened 2026-04-19
fix(adhoc): support Map-typed columns in ad-hoc filters

### LINKED ISSUE #1434: Map type broken for Ad-Hoc filters

adamyeats:
## Summary

Fixes #1434 — ad-hoc filters on `Map(...)` columns previously failed in two ways, both entirely in the plugin (despite the issue's `waiting-on-clickhouse` label, no upstream change is required):

1. **Filter values dropdown showed `[object Object]`.** `fetchTagValuesFromSchema()` issued `SELECT DISTINCT <col>` against a Map column, returning whole Map values that the frontend stringified via `String(value)`.
2. **Filters silently produced invalid SQL for non-OTel Map columns.** `escapeKey()` used a hard-coded allowlist of three OTel names (`ResourceAttributes`, `ScopeAttributes`, `LogAttributes`) and treated every other dotted key as a table prefix, so user-defined Map columns and OTel variants like `SpanAttributes` didn't work.

### 1. Map-aware tag-keys expansion ([src/data/CHDatasource.ts](src/data/CHDatasource.ts))

`getTagKeys()` now inspects the `type` column from `system.columns`, collects Map-typed columns per-table, and — when the adhoc context points at a specific `db.table` — fans out parallel `fetchUniqueMapKeys()` probes and expands each Map column into one tag key per discovered map key. Falls back to the flat entry when no table context is available (avoids N-table fan-out on database-wide adhoc variables).

Detection handles `Map(...)`, `Nullable(Map(...))`, and `LowCardinality(Map(...))`.

### 2. Map-aware tag-values fetch ([src/data/CHDatasource.ts](src/data/CHDatasource.ts))

`fetchTagValuesFromSchema()` recognises dotted keys whose head refers to a Map-typed column (from the cache populated in step 1) and rewrites the SELECT to `SELECT DISTINCT col['key'] FROM …` instead of `SELECT DISTINCT col FROM …`. Fixes the `[object Object]` dropdown.

### 3. Schema-driven `escapeKey` ([src/data/adHocFilter.ts](src/data/adHocFilter.ts))

Replaces the three-name allowlist with a schema-driven set published by the datasource via `AdHocFilter.setMapColumns()`. The OTel names remain as fallback defaults so behavior does not regress for users who hand-construct `AdHocFilter` or whose tables use the canonical names. Now correctly handles:

- `table.MapCol.key1.key2` → `MapCol['key1.key2']` (table-prefixed)
- `MapCol.key` → `MapCol['key']` (no prefix, e.g. hideTableNameInAdhocFilters=true)
- `table.plain_col` → `plain_col` (unchanged, non-Map dotted path)

### Tests

- Unit tests for all three paths in [src/data/adHocFilter.test.ts](src/data/adHocFilter.test.ts) and [src/data/CHDatasource.test.ts](src/data/CHDatasource.test.ts).
- New e2e regression guard [tests/e2e/mapFilter.spec.ts](tests/e2e/mapFilter.spec.ts) runs the exact SQL shapes our adhoc code paths produce (mapKeys discovery, bracket-access values, full `additional_table_filters` apply) against a real ClickHouse fixture added to [tests/e2e/fixtures/seed.sql](tests/e2e/fixtures/seed.sql) (`e2e_test.map_events` with a `Map(String, String)` column).

## Scope notes

- **Probes are gated to specific `db.table` adhoc contexts.** Database-wide or schema-wide adhoc variables still show the flat Map-column entry to avoid a per-table fan-out on every dashboard render. A follow-up can relax this if needed.
- **OTel fallback preserved.** `setMapColumns()` is additive — `DEFAULT_MAP_COLUMNS` (`ResourceAttributes`, `ScopeAttributes`, `LogAttributes`) is always merged in. Tests that construct `AdHocFilter` directly without a datasource still pass.
- **Aligned with [#1461](https://github.com/grafana/clickhouse-datasource/issues/1461)** — both this and the JSON-path discovery bug share a "sample map-like keys from a large table" pattern. This PR reuses the existing `fetchUniqueMapKeys()` implementation; if [#1461](https://github.com/grafana/clickhouse-datasource/issues/1461)'s sampling strategy changes, this will inherit it for free.

## Related

- Touches the same `escapeKey()` / dotted-column-path area as [#798](https://github.com/grafana/clickhouse-datasource/issues/798) (dots in column names). Not fixed here but adjacent.
- [#1063](https://github.com/grafana/clickhouse-datasource/issues/1063) (adhoc filter no-ops in Logs view) may also be resolved by the Bug 2 fix if the underlying cause is SQL rejection — worth a repro after this lands.

Closes #1434

### DIFF
--- a/cspell.config.json
+++ b/cspell.config.json
@@ -91,6 +91,7 @@
     "sqlutil",
     "stagename",
     "stretchr",
+    "subkey",
     "subquery",
     "subqueries",
     "subresource",
--- a/src/data/CHDatasource.test.ts
+++ b/src/data/CHDatasource.test.ts
@@ -12,7 +12,14 @@ import { DataQuery } from '@grafana/schema';
 import { mockDatasource } from '__mocks__/datasource';
 import { cloneDeep } from 'lodash';
 import { of } from 'rxjs';
-import { BuilderMode, ColumnHint, FilterOperator, OrderByDirection, QueryBuilderOptions, QueryType } from 'types/queryBuilder';
+import {
+  BuilderMode,
+  ColumnHint,
+  FilterOperator,
+  OrderByDirection,
+  QueryBuilderOptions,
+  QueryType,
+} from 'types/queryBuilder';
 import { CHBuilderQuery, CHQuery, CHSqlQuery, EditorType } from 'types/sql';
 import { AdHocFilter } from './adHocFilter';
 import { Datasource } from './CHDatasource';
@@ -139,7 +146,7 @@ describe('ClickHouseDatasource', () => {
       // Setup ad-hoc filters
       const adHocFilters = [
         { key: 'column', operator: '=', value: 'value' },
-        { key: 'column.nested', operator: '=', value: 'value2' }
+        { key: 'column.nested', operator: '=', value: 'value2' },
       ];
 
       // Mock getAdhocFilters to return our test filters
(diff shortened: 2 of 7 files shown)


akkikumar72 (commented):
Overall looks good to me, but wanna have a peak at High prior review from Co-pilot?

akkikumar72 (approved):

PR #1794, opened 2026-04-19
fix(adhoc): map regex operators to REGEXP instead of ILIKE

### LINKED ISSUE #1443: Adhoc filter for regex uses ILIKE by default
**What happened**:
This is for Adhoc filters feature and its interpretation in additional_table_filters. When I have to do a "Matches regex" or "Does not match regex" via filters it uses ILIKE in the additional_table_filters. I have indexed a field and when ILIKE is used, the query execution time increases a lot as Clickhouse is no more honoring the indexed data. 

Please change the default to LIKE or provide a way where user can explicitly choose what they want to use

- Create a table in clickhouse with minimal fields as time, message, logLevel, applicationName
- Connect this as a datasource on Grafana Logs panel
- Editor Type: SQL Editor, Query Type: Logs
- Add an Adhoc filter for any field for "Matches regex" or "Does not match regex"
- Check the query created via query inspector and notice content of additional_table_filters

**Screenshots**

Image

This is for an airgapped environment

**Environment**:

- Grafana version: 11.6.2
- Plugin version: 4.11.3 and earlier versions
- OS Grafana is installed on: OEL8
- User OS & Browser: MAC, Google Chrome
- Others:

adamyeats:
Fixes #1443

## Summary

- Grafana's ad-hoc filter operators `=~` ("Matches regex") and `!~` ("Does not match regex") were being translated to `ILIKE` / `NOT ILIKE` in the `additional_table_filters` setting. That is semantically wrong (ILIKE is a LIKE pattern, not a regex) and, per the reporter, prevents primary-key / skip-index pruning on indexed columns, producing very slow queries over large tables.
- Map `=~` / `!~` to ClickHouse's `REGEXP` / `NOT REGEXP` (RE2-backed, alias for `match()`). Anchored patterns can still benefit from prefix pruning, and the semantics now match the operator label shown in the UI.
- Only `src/data/adHocFilter.ts` changes runtime behaviour. The Query Builder's `ILIKE` / `NOT ILIKE` options in `FilterEditor.tsx` are untouched — they're an explicit user choice and unrelated to the regex operators.

### DIFF
--- a/src/data/adHocFilter.test.ts
+++ b/src/data/adHocFilter.test.ts
@@ -131,25 +131,25 @@ describe('AdHocManager', () => {
     expect(val).toEqual(`SELECT foo.stuff FROM foo settings additional_table_filters={'foo' : ' key = \\'val\\' '}`);
   });
 
-  it('apply ad hoc filter converts "=~" to "ILIKE"', () => {
+  it('apply ad hoc filter converts "=~" to "REGEXP"', () => {
     const ahm = new AdHocFilter();
     ahm.setTargetTableFromQuery('SELECT * FROM foo');
     const val = ahm.apply('SELECT stuff FROM foo WHERE col = test', [
       { key: 'key', operator: '=~', value: 'val' },
     ] as AdHocVariableFilter[]);
     expect(val).toEqual(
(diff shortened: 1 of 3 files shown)


mattvella07 (approved):
LGTM

PR #1797, opened 2026-04-19
fix(headers): forward X-Grafana-User to ClickHouse when enabled

### LINKED ISSUE #1451: Plugin does not pass X-Grafana-User http header even when grafana enable it
**What happened**:
Hi
I've tried enabling `[dataproxy] send_user_header = true` and use `Forward Grafana HTTP Headers` toggle in the datasource settings. Connected to a server that dumps all http headers - no X-Grafana-User header, when i try the same with Altinity plugin - it works. If i try to enable/disable send_user_header parameter - it appear/disappear from Altinity plugin requests.

Here is an example of Clickhouse plugin headers:
```
echo-server-1  | 2025-11-21T07:26:40.301Z Headers {
echo-server-1  |   host: '10.1.0.249:8080',
echo-server-1  |   'user-agent': 'grafana/12.2.1 clickhouse-datasource/4.11.2 clickhouse-go/2.40.3 (lv:go/1.24.7; os:linux)',
echo-server-1  |   'content-length': '55',
echo-server-1  |   'accept-encoding': 'gzip',
echo-server-1  |   authorization: 'Basic bmFtZTpwdw=='
echo-server-1  | } /?client_protocol_version=54460&database=analytics&default_format=Native&limit=1000000&max_execution_time=14
```

X-Grafana-User header should present on http request to clickhouse from grafana

Basic grafana config:
```
data:
  grafana.ini: |
    [server]
    protocol = http

    [dataproxy]
    send_user_header = true

    [users]
    allow_sign_up = false

    [auth]
    disable_login_form = true

    [auth.proxy]
    enabled = true
    header_name = X-AUTH-REQUEST-EMAIL
    header_property = email
    headers_encoded = false
    auto_sign_up = true
    sync_ttl = 0
```
Create new datasource, connect to HTTP Clickhouse.
Test data-source, see headers in the proxy

**Screenshots**

Tried both HTTP and HTTPS versions, tried to use the data-source in the dashboard - results are the same, no header in the request

**Environment**:

- Grafana version: Grafana v12.2.1 (563109b696)
- Plugin version: 4.11.2
- OS Grafana is installed on: k8s installation
- User OS & Browser: Windows Chrome
- Others:

Possible duplicate [issue](https://github.com/grafana/clickhouse-datasource/issues/1137)

adamyeats:
## Summary

- Populate `X-Grafana-User` from the plugin's request-user context when "Forward Grafana HTTP Headers" is enabled on the datasource, so operators can attribute ClickHouse queries back to the Grafana user who ran them

## User-facing impact

When `forwardGrafanaHeaders` is enabled on the datasource and a Grafana user runs a query:

- **Before:** ClickHouse saw no `X-Grafana-User` header. `system.query_log.http_user_agent` showed only the generic `grafana/<ver> clickhouse-datasource/<ver>` string, and row policies / quotas could not key on the originating user.
- **After:** ClickHouse sees `X-Grafana-User: <login>` on every query, matching the behavior of the Altinity plugin. Existing ClickHouse policies, quotas, and query-log attribution keyed on that header begin working.

No behavior change when `forwardGrafanaHeaders` is off. No new datasource settings.

## Why

Grafana's `dataproxy.send_user_header = true` only injects `X-Grafana-User` on the Grafana *proxy* path (core HTTP datasources routed through Grafana's HTTP proxy). Plugin-initiated connections — including this one — never hit that middleware, so the header never reaches ClickHouse. Users on the [#1451](https://github.com/grafana/clickhouse-datasource/issues/1451) thread have observed that the Altinity plugin forwards the user header end-to-end while this plugin silently drops it.

## What changed

[pkg/plugin/driver.go:391](pkg/plugin/driver.go#L391) — new `injectGrafanaUserHeader` helper, called from `MutateQueryData`:

1. Gated on `settings.ForwardGrafanaHeaders == true` (the existing "Forward Grafana HTTP Headers" toggle). No new config surface.
2. Only populates the header if Grafana didn't already set it (preserves proxy-path compatibility).
3. Reads the user via `backend.UserFromContext(ctx)`, which is already the canonical source — the existing `MutateQuery` path uses it to emit the `grafana_user:<login>` query comment.
4. Calls `req.SetHTTPHeader("X-Grafana-User", user.Login)`, which sqlds picks up via `req.GetHTTPHeaders()` on the next hop, threads through `ConnectionArgs`, and our existing `extractForwardedHeadersFromMessage` → `opts.HttpHeaders` path forwards to the ClickHouse HTTP client.

Connection caching is unaffected: since `EnableMultipleConnections` is already on when `forwardGrafanaHeaders` is set ([pkg/plugin/datasource.go:17](pkg/plugin/datasource.go#L17)), sqlds keys the connection pool on `ConnectionArgs`, so different Grafana users get separate connections with their own baked-in headers.

## Docs

[docs/sources/configure.md](docs/sources/configure.md) — new **Forward Grafana HTTP headers** section covering:

- When the toggle applies (HTTP protocol only) and how to enable it from **Optional HTTP settings**
- The full header set that gets forwarded, including the note that `X-Grafana-User` is now populated from the plugin request context — operators no longer need `[dataproxy] send_user_header = true` in the Grafana configuration for this plugin
- Use cases: query-log attribution, ClickHouse row policies, and quotas keyed on `X-Grafana-User`
- Connection-pool sizing implications (per-user ClickHouse connections when forwarding is on)
- Relationship to **Custom HTTP headers** (always sent, independent of the toggle)

Addresses @alyssajoyner's review comment.

## Testing

- `go test ./pkg/plugin/` — all pass, including six new subtests in `TestMutateQueryData_XGrafanaUserForwarding`:
  - populates `X-Grafana-User` when the toggle is on
  - no-op when the toggle is off
  - does not override a header Grafana already set on the proxy path
  - no-op when there is no user in context (ephemeral system queries)
  - no-op when `DataSourceInstanceSettings` is nil (defensive guard)
  - no-op when `user.Login` is empty

### Manual verification

With an echo-server pointed at ClickHouse's ingress:
- Enable `forwardGrafanaHeaders: true` on the datasource
- Run a query as user `alice`
- Echo-server shows `X-Grafana-User: alice` on the HTTP request
- Toggling the datasource setting off removes the header on next connection

Fixes #1451

### DIFF
--- a/cspell.config.json
+++ b/cspell.config.json
@@ -28,6 +28,7 @@
     "concats",
     "createddate",
     "dataframe",
+    "dataproxy",
     "DataSource",
     "datasourcese",
     "DataSources",
--- a/docs/sources/configure.md
+++ b/docs/sources/configure.md
@@ -225,6 +225,37 @@ Once you have configured your ClickHouse connection settings, click **Save & tes
 
 If the test fails, refer to [Troubleshoot ClickHouse data source issues](/docs/plugins/grafana-clickhouse-datasource/<CLICKHOUSE_PLUGIN_VERSION>/troubleshooting/) for common configuration errors and solutions.
 
+## Forward Grafana HTTP headers
+
+When you use the **HTTP** protocol, you can propagate Grafana's per-request HTTP headers end-to-end to ClickHouse. This attaches context about the originating Grafana user, dashboard, and panel to every query so you can drive query-log attribution, quotas, and row policies from ClickHouse.
+
+To enable it, expand **Optional HTTP settings** on the data source and turn on **Forward Grafana HTTP headers to data source**.
+
+{{< admonition type="note" >}}
+This setting is only available on the HTTP protocol. The native protocol does not carry HTTP headers.
+{{< /admonition >}}
+
+When the toggle is on, the following headers are forwarded on each ClickHouse connection:
+
+| Header | Source |
+| ------ | ------ |
+| `X-Grafana-User` | The logged-in Grafana user's login. Populated from the plugin's request context, so you do not need to enable the Grafana `dataproxy.send_user_header` setting for this plugin. |
+| `X-Dashboard-Uid`, `X-Panel-Id`, `X-Panel-Plugin-Id`, `X-Dashboard-Title`, `X-Panel-Title` | Identifiers set by Grafana when the query originates from a dashboard panel. |
+| `X-Grafana-Org-Id`, `X-Query-Group-Id`, `X-Grafana-From-Expr`, `X-Datasource-Uid` | Request-context headers set by Grafana core. |
+
+### Use cases
+
+- **Query-log attribution** — ClickHouse records the forwarded headers in `system.query_log.http_user_agent` and related fields, so operators can correlate queries back to the Grafana user and dashboard that triggered them.
+- **Row policies and quotas** — ClickHouse [row policies](https://clickhouse.com/docs/en/operations/access-rights/#row-policies) and [quotas](https://clickhouse.com/docs/en/operations/quotas) can key on the `X-Grafana-User` header, so a single shared ClickHouse account can still enforce per-viewer access rules.
+
+### Connection pool implications
+
+With header forwarding enabled, connections are keyed by the forwarded header set, which means each distinct Grafana user opens their own ClickHouse connection. Expect the connection count to scale with concurrent unique users and size `max_connections` on your ClickHouse server accordingly.
+
+### Custom headers
+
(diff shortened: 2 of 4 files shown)


alyssajoyner (commented):
This looks really good. One question - is it worth adding this to the docs? Right now we don't have anything about forwarding headers and it could be useful to add it to `configure.md`. Wdyt?

SpencerTorres (approved):

adamyeats:
@alyssajoyner Updated!

alyssajoyner (approved):

PR #1802, opened 2026-04-22
refactor(macros): adopt grafana/macropro as the macro engine

adamyeats:
## Summary

Swaps the ClickHouse datasource's macro engine from `sqlds`'s built-in `MacroFunc` set to [`github.com/grafana/macropro`](https://github.com/grafana/macropro) v1.0.0. Macros are now expanded once in `MutateQueryData` and `sqlds.Macros()` returns empty so `sqlds`/`sqlutil` do not re-scan already-expanded SQL.

This branch includes the initial refactor plus three follow-up fixes:

- **Downstream-error classification** — argument-count and bad-duration errors from every macro handler are wrapped via `backend.DownstreamError` again, and argument-count errors continue to unwrap to `sqlutil.ErrorBadArgumentCount` so existing `errors.Is` checks at the `sqlds` boundary keep working. Prior to this branch these were plain `fmt.Errorf` and would have been attributed to the plugin rather than user input.
- **Surface expansion failures per-query** — `MutateQueryData` has no error return, so previously a macro handler error was logged and the unexpanded SQL was handed to ClickHouse, which produced a cryptic syntax error on the raw `$__macro` token. On failure the query's `rawSql` is now rewritten to `SELECT throwIf(1, 'macro expansion failed: <message>')`, which raises a ClickHouse Exception carrying the real error. `MutateQueryError` already classifies ClickHouse exceptions as downstream, so the user sees why it failed. Single quotes are doubled so a quoted value in the error can't close the string literal early.

### DIFF
--- a/pkg/plugin/driver.go
+++ b/pkg/plugin/driver.go
@@ -450,40 +450,38 @@ func expandMacros(req *backend.QueryDataRequest) *backend.QueryDataRequest {
 // instead of a cryptic DB syntax error on the unexpanded macro. We cannot
 // return an error from MutateQueryData directly — sqlds's hook signature
 // forbids it — so the DB round-trip is the only available error channel.
+//
+// The query JSON is decoded once into a map so the rewritten rawSql can
+// be written back while preserving plugin-specific fields (meta.timezone,
+// format, etc.). macropro only needs the raw SQL plus TimeRange/Interval, and
+// the latter come from the backend.DataQuery envelope rather than the JSON
+// body, so a full sqlutil.Query decode is unnecessary. On any (un)marshal
+// failure the original query is returned unchanged.
 func expandMacrosInQuery(q backend.DataQuery) backend.DataQuery {
-	var sqq sqlutil.Query
-	if err := json.Unmarshal(q.JSON, &sqq); err != nil {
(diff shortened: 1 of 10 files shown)


pkg/plugin/driver.go
```diff
@@ -418,6 +423,97 @@ func injectGrafanaUserHeader(ctx context.Context, req *backend.QueryDataRequest)
...
+	}
+	return replaceRawSQL(q, expanded)
+}
+
+// replaceRawSQL writes newRawSQL into q.JSON while preserving plugin-specific
+// fields (meta.timezone, format, etc.) that sqlutil.Query does not know about.
+// On any serialization failure the original query is returned unchanged.
+func replaceRawSQL(q backend.DataQuery, newRawSQL string) backend.DataQuery {
+	var raw map[string]json.RawMessage
+	if err := json.Unmarshal(q.JSON, &raw); err != nil {
```
aangelisc: nit: can we avoid the unmarshal here and in `expandMacrosInQuery`?

aangelisc (commented):
One nit, otherwise LGTM

aangelisc (approved):
Nice!

PR #1820, opened 2026-04-30
fix: dedupe autocomplete suggestions and dispose Monaco providers on unmount

adamyeats:
## Type of Change

- [x] 🐛 Bug Fix

---

### What is the bug?

The SQL editor's autocomplete returned duplicate suggestions — the same column or keyword appearing multiple times in the list. After merging #1779 (case-insensitive matching) the duplicates became less frequent but never disappeared.

Two root causes:

1. **Provider-registration leak in `registerSQL`.** `monaco.languages.registerCompletionItemProvider('sql', …)` returns an `IDisposable`, but `registerSQL` did not capture or expose it. Every time `SqlEditor` mounted (panel open/close, dashboard navigation, inspector toggle) a *new* completion provider was registered for `sql`. Monaco calls every registered provider for the language and merges results, so suggestions multiplied per re-mount.
2. **No within-call dedupe.** Even with a single provider, upstream sources can occasionally emit the same label twice (e.g., `system.functions` rows for case-variant aliases on some ClickHouse builds). The hot path had no guard.

### How to reproduce

1. Open a Grafana dashboard backed by the ClickHouse data source.
2. Open the panel editor → SQL editor mode.
3. Trigger autocomplete (e.g., type `SELECT ` and pause). Note the suggestion list.
4. Close the panel editor; reopen it. Trigger autocomplete again.
5. Each round-trip approximately doubles the number of times each column / function / macro appears in the list.

### Related Issues

Source feedback: grafana/data-sources#1108 (item 2). Follow-up to #1779, which closed the case-sensitivity half of the same item but explicitly deferred the duplicate fix.

---

## User-facing impact

**What users will see:** the autocomplete popup in the SQL editor now shows each column, function, macro, and keyword exactly once, regardless of how many times the user has opened/closed the panel editor or navigated between dashboards in the same browser session. Previously, the suggestion list grew over a session — the same column could appear 2×, 3×, or more depending on how many times the editor had been mounted.

No change to suggestion contents, ordering, or filter behaviour. No new config. No migration.

---

## Fix

- `src/components/sqlProvider.ts` — `registerSQL` captures both registration disposables and returns `{ monacoEditor, dispose }`.
- `src/components/SqlEditor.tsx` — stores `dispose` in a ref and invokes it from `onEditorWillUnmount`.
- `src/components/suggestions.ts` — defense-in-depth dedupe of the final result by `(label, kind, insertText)`. Cheap and bulletproof against any future upstream change that re-introduces duplicates.

---

## Tests

**Unit:**
- `suggestions.test.ts` — schema deliberately returns `CodeFile` twice and `toDateTime` twice; assert each appears exactly once in suggestions.
- `sqlProvider.test.ts` — assert `dispose()` invokes both provider disposables; assert each `registerSQL` call creates a fresh registration (so the caller is responsible for disposing).
- `SqlEditor.test.tsx` — render + unmount the editor and assert both Monaco providers' `dispose` were called.

**E2E (`tests/e2e/sqlAutocomplete.spec.ts`):**
- Navigate to Explore, trigger macro autocomplete via `$__`, capture labels, assert no duplicates.
- Navigate away and back to force a remount, repeat the trigger, assert no duplicates and identical label set across mounts. This is the regression-driver: without the dispose fix the second mount stacks a provider and every macro doubles.

All 676 unit tests pass; tsc, eslint, and cspell are clean on the changed files.

---

## Please check that:

The dispose-on-unmount fix is the dominant cause; the within-call dedupe is belt-and-braces. They are independent — either alone would reduce duplicates, but together they cover both the leaked-provider case and any future upstream weirdness (e.g., system.functions duplicates).


src/components/suggestions.ts
```diff
@@ -141,7 +141,23 @@ export async function getSuggestions(
...
+  const results = await getSuggestionsFromCursorData(cursorData, schema, range);
+  return dedupeSuggestions(results);
+}
+
+function dedupeSuggestions(
+  items: monacoTypes.languages.CompletionItem[]
+): monacoTypes.languages.CompletionItem[] {
+  const seen = new Set<string>();
+  return items.filter((item) => {
+    const label = typeof item.label === 'string' ? item.label : item.label.label;
```
bossinc, line 151: Why is it item.label.label? Can `item.label` be null?

src/components/suggestions.ts
```diff
@@ -141,7 +141,23 @@ export async function getSuggestions(
...
-  return await getSuggestionsFromCursorData(cursorData, schema, range);
+  const results = await getSuggestionsFromCursorData(cursorData, schema, range);
+  return dedupeSuggestions(results);
+}
+
+function dedupeSuggestions(
+  items: monacoTypes.languages.CompletionItem[]
+): monacoTypes.languages.CompletionItem[] {
+  const seen = new Set<string>();
+  return items.filter((item) => {
```
bossinc, line 150: Since we are building lists here (seen and the return of filter) should we make a loop instead of filtering and return seen?

bossinc: talked about in slack. it doesnt really matter and we would be alloc another list if we did my suggestion anyways. this way is cleaner

bossinc (approved):

PR #1824, opened 2026-05-05
feat: implement DataSourceWithToggleableQueryFiltersSupport interface

karl-power:
### Feature
Implements Grafana's `DataSourceWithToggleableQueryFiltersSupport` interface so the existing **+ / −** "Filter for / Filter out value" buttons in expanded log-row details now show as **active (blue) when a matching filter is already on the query**, and clicking the active button removes it.

Two new methods on `Datasource`:

- `queryHasFilter` — Grafana calls this to decide whether to render the button in its active state.
- `toggleQueryFilter` — handles the click: adds the filter if absent, removes it if it already matches (toggle off), or swaps the operator if the opposite filter exists.

Both reuse the column-resolution path from `modifyQuery` (extracted into shared helpers), so log-alias hints (`level` → `LogLevel`) and OTel map keys (`ResourceAttributes.service.name`) work the same way they already do for the existing +/− buttons. SQL-mode queries are returned unchanged.

### How to test this feature

1. In Explore, select the ClickHouse datasource, switch to Query Builder, choose Query Type **Logs**, configure a table with a log-level column, and run the query.
2. Expand a log row and click **+** on the level field — filter is added, the **+** button turns blue.
3. Click the blue **+** again — filter is removed, button returns to default state.
4. Click **−** on a different value, then **+** on that same value — the `!=` filter is replaced by `=`.
5. Repeat with an OTel map key (`ResourceAttributes.service.name`) and confirm the same toggle behavior.
6. Switch to SQL mode and click **+** — the query must not be modified.

### Screenshots
Screenshot 2026-05-05 at 11 29 08
Screenshot 2026-05-05 at 11 28 56


SpencerTorres (approved):

adamyeats:
@karl-power Please rebase and sign your commits

karl-power:
> @karl-power Please rebase and sign your commits

@adamyeats signed and rebased

src/data/CHDatasource.ts
```diff
@@ -615,6 +630,82 @@ export class Datasource
...
+    return filters.some(
+      (f) =>
+        filterMatchesColumn(f, resolved) &&
+        f.operator === FilterOperator.Equals &&
+        'value' in f &&
+        String(f.value) === value
+    );
+  }
+
+  toggleQueryFilter(query: CHQuery, filter: ToggleFilterAction): CHQuery {
```
adamyeats, line 661: `FILTER_FOR` only removes an exact same-value filter (toggle off) or the opposite operator at the same value, never a same-operator filter with a different value. So with `level = 'info' ` set, clicking + on error produces level = 'info' AND level = 'error' = zero rows.

`modifyQuery`'s `ADD_FILTER` already guards this (see CHDatasource.ts:575), but implementing this interface routes + through `toggleQueryFilter,` bypassing it. Scope the fix to the Equals branch: `FILTER_OUT`'s `!= a AND != b` is correct. A unit test for this case would be good too.

src/data/CHDatasource.ts
```diff
@@ -615,6 +630,82 @@ export class Datasource
...
+    if (query.editorType !== EditorType.Builder) {
+      return query;
+    }
+
+    const key = filter.options.key;
+    const value = filter.options.value;
+    if (!key || !value) {
```
adamyeats: Nit: `!value` rejects empty-string values, so filtering on a field with value `''` is a no-op here while `modifyQuery` would still add it. Guard on key only unless empty values shouldn't be filterable.

adamyeats (changes requested):

adamyeats (approved):
Tested and working! Thanks @karl-power ! 🚀

PR #1825, opened 2026-05-05
chore: add claude.md and agents.md

karl-power:
Add `agents.md` and `claude.md`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,816 @@
+# agents.md — grafana/clickhouse-datasource
+
+This document is a comprehensive reference for AI coding agents (and human contributors) working in the [`grafana/clickhouse-datasource`](https://github.com/grafana/clickhouse-datasource) repository. It covers architecture, code structure, development workflow, testing, and contribution conventions.
+
+---
+
+## Table of Contents
+
+1. [Repository Overview](#repository-overview)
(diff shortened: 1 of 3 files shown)


SpencerTorres (approved):
Not sure how advanced some agents are in this space, but this file is a bit long and I want to be considerate of token usage for local agents. I know Claude has no problems with this, but does other agent tooling know how to consume this file in portions? Curious if a split directory of files may be necessary.

We can change it if needed later, lets get this added for now 👌

alex-fedotyev:
Good question. The short answer: most hosted agents (Claude Code, Codex, Copilot) load the full file and handle 800 lines without issues. Local models are a different story since their context windows are much smaller.

A few options if splitting becomes worth it later:

- A top-level short AGENTS.md (50-100 lines: build, test, common gotchas) with links to a per-topic `agent-docs/` directory for the detail sections. Agents needing deep context can follow the links; agents doing quick edits get the essentials upfront.
- Or a minimal `AGENTS.md` that just points at the existing `DEV_GUIDE.md` and `CONTRIBUTING.md` for sections already covered there, and adds agent-specific notes (tool restrictions, PR conventions, test commands) in a focused 200-line file.

Happy to do that split in a follow-up if the team prefers that shape. For now, merging as-is and opening an issue to track the split makes sense.

PR #1828, opened 2026-05-06
feat(queryBuilder): add SchemaPicker for cascading schema navigation

alex-fedotyev:
## Type of Change

- [x] Feature
- [ ] Bug Fix
- [ ] Documentation
- [ ] Refactor
- [ ] Chore

### What

Adds `src/components/queryBuilder/SchemaPicker.tsx`, a reusable cascading picker for **Database -> Table -> Column -> Map Key** with a configurable depth. The depth is controlled by a `level` prop:

| `level` | Renders |
| ------- | ------- |
| `'database'` | Database |
| `'table'` | Database + Table |
| `'column'` (default) | Database + Table + Column |
| `'mapKey'` | Database + Table + Column + Map Key (Map Key only when the selected column is a `Map(...)` type) |

Reuses the existing `useDatabases`, `useTables`, `useColumns`, and `useUniqueMapKeys` hooks, so it stays consistent with the rest of the query builder. Changing a parent value clears every downstream selection in a single `onChange` emission.

The component API:

```ts
export type SchemaPickerLevel = 'database' | 'table' | 'column' | 'mapKey';

export interface SchemaPickerValue {
  database?: string;
  table?: string;
  column?: string;
  mapKey?: string;
}

export interface SchemaPickerProps {
  datasource: Datasource;
  value: SchemaPickerValue;
  onChange: (value: SchemaPickerValue) => void;
  level?: SchemaPickerLevel;
  labels?: Partial<Record<SchemaPickerLevel, string>>;
}
```

No consumer migrates in this PR.

### Why

This is the foundation for several upcoming editor improvements. The depth-mode design lets the same primitive cover:

- variable editor improvements (column-values + Map keys),
- annotation editor improvements (deployment / event change detection on Map keys),
- datasource configuration (replacing the plain `<Input>` text fields for default database / table),
- the upcoming Compact Query Mode (override picker above `CompactMetricsBar`),
- consolidation of the existing `DatabaseTableSelect` cascade.

Each of those is tracked as a separate ticket so this PR can stay small and focused. Sub-150-line, single-concern PRs in this repo land same-day to two days based on the recent review cadence (#1818, #1819, #1820), so I want to get the primitive in first and migrate consumers behind it.

### Who

Anyone who edits the ClickHouse datasource: query authors, dashboard authors, and the upcoming variable / annotation flows.

### How to test

1. `yarn jest src/components/queryBuilder/SchemaPicker.test.tsx` (15 tests covering all four depth modes, cascade invalidation, and the `labels` override).
2. `yarn jest` (701 tests pass, no regressions).
3. `npm run lint` and `npm run prettier:check`: clean for the new files. The remaining warnings are pre-existing on `main`.
4. The PR has no consumer in this branch, so manual UI testing was done against a local visual harness rendered inside the existing query editor in Grafana 12 (see screenshots below); the harness is **not** part of this diff.

## Screenshots / Videos

Verified in Grafana 12 against the public ClickHouse demo (`sql-clickhouse.clickhouse.com/otel_v2`) using a temporary harness that renders the picker at all four depths. The harness lives on a throwaway branch (`feat/schema-picker-harness`) and is **not** part of this PR.

### Dark theme

Empty state, all four levels:

SchemaPicker dark, empty

Populated, levels database / table:

SchemaPicker dark, populated top

Populated, levels column / mapKey (LogAttributes is a Map column, so the Map Key picker appears):

SchemaPicker dark, populated bottom

### Light theme

Empty state:

SchemaPicker light, empty

Populated, levels database / table / column:

SchemaPicker light, populated top

Populated mapKey level alongside the existing query builder Database / Table picker (visual parity check):

SchemaPicker light, mapKey populated

The picker reuses the same `gf-form` / `InlineFormLabel` / `Select` primitives as `DatabaseTableSelect` and `ColumnSelect`, so the rendering stays consistent with the rest of the editor. You can see this in the last screenshot, where the harness pickers above and the query builder pickers below have identical labels, spacing, and dropdown chrome.

## Please check that:

- [x] Tests added / updated (`SchemaPicker.test.tsx`, 15 cases)
- [ ] Docs updated (no user-visible config in this PR; consumers will document their own surfaces)

- The component intentionally has no consumer in this PR. Migrations land separately so review can stay focused on the primitive.
- Cascade invalidation is verified by unit tests. A user changing the database with `level='mapKey'` selected gets a single `onChange` emission with `table`, `column`, and `mapKey` cleared.
- The Map Key picker only renders when `level='mapKey'` AND the selected column type starts with `Map(`. For non-Map columns at `level='mapKey'`, the picker behaves exactly like `level='column'`.
- I did not introduce a `columnFilter` prop or other speculative API. They can be added when a consumer actually needs them.
- The deprecated `Select` warning is pre-existing in this directory; switching to `Combobox` would diverge from `DatabaseTableSelect` / `ColumnSelect`. Happy to do the migration as a separate PR if useful.

Add the `changelog` label.

### DIFF
--- a/src/components/queryBuilder/SchemaPicker.tsx
+++ b/src/components/queryBuilder/SchemaPicker.tsx
@@ -1,0 +1,172 @@
+import React from 'react';
+import { SelectableValue } from '@grafana/data';
+import { InlineFormLabel, Select } from '@grafana/ui';
+import { Datasource } from 'data/CHDatasource';
+import useColumns from 'hooks/useColumns';
+import useDatabases from 'hooks/useDatabases';
+import useTables from 'hooks/useTables';
+import useUniqueMapKeys from 'hooks/useUniqueMapKeys';
+import labels from 'labels';
+import { styles } from 'styles';
+
+/**
+ * How deep the cascading picker goes. Each level builds on the one above.
+ *  - 'database': database picker only
+ *  - 'table': database + table
+ *  - 'column': database + table + column (default)
+ *  - 'mapKey': database + table + column + map key (when the column is a Map type)
+ */
+export type SchemaPickerLevel = 'database' | 'table' | 'column' | 'mapKey';
+
+/** Selected schema parts. Downstream fields are cleared when an upstream field changes. */
+export interface SchemaPickerValue {
+  database?: string;
+  table?: string;
+  column?: string;
+  mapKey?: string;
+}
+
+export interface SchemaPickerProps {
+  datasource: Datasource;
+  value: SchemaPickerValue;
+  onChange: (value: SchemaPickerValue) => void;
+  /** How deep the cascade renders. Defaults to 'column'. */
+  level?: SchemaPickerLevel;
+  /** Override visible labels per level. */
+  labels?: Partial<Record<SchemaPickerLevel, string>>;
+}
+
+const LEVEL_ORDER: SchemaPickerLevel[] = ['database', 'table', 'column', 'mapKey'];
+const MAP_TYPE_PREFIX = 'Map(';
+
+function levelIndex(level: SchemaPickerLevel): number {
+  return LEVEL_ORDER.indexOf(level);
+}
+
+/**
+ * Reusable cascading schema picker: Database -> Table -> Column -> Map Key.
+ *
+ * Reuses the existing schema-fetch hooks (useDatabases, useTables, useColumns,
+ * useUniqueMapKeys) so it stays consistent with the rest of the query builder.
+ *
+ * Usage: pick a depth via the `level` prop. Changing a parent value clears
+ * every downstream selection in a single onChange emission.
+ */
+export const SchemaPicker = (props: SchemaPickerProps) => {
+  const { datasource, value, onChange, level = 'column', labels: labelOverrides } = props;
+  const maxLevel = levelIndex(level);
+
+  const databases = useDatabases(datasource);
+  const tables = useTables(datasource, value.database || '');
+  const columns = useColumns(datasource, value.database || '', value.table || '');
+
+  const selectedColumn = columns.find((c) => c.name === value.column);
+  const isMapColumn = selectedColumn ? selectedColumn.type.startsWith(MAP_TYPE_PREFIX) : false;
+  const mapColumnName = isMapColumn && value.column ? value.column : '';
+  const mapKeys = useUniqueMapKeys(datasource, mapColumnName, value.database || '', value.table || '');
+
+  const labelText = (key: SchemaPickerLevel, fallback: string): string => labelOverrides?.[key] ?? fallback;
+
+  const handleDatabaseChange = (selected: SelectableValue<string>) => {
+    onChange({ database: selected.value || '', table: '', column: '', mapKey: '' });
+  };
+
+  const handleTableChange = (selected: SelectableValue<string>) => {
+    onChange({ ...value, table: selected.value || '', column: '', mapKey: '' });
+  };
+
+  const handleColumnChange = (selected: SelectableValue<string>) => {
+    onChange({ ...value, column: selected.value || '', mapKey: '' });
+  };
+
+  const handleMapKeyChange = (selected: SelectableValue<string>) => {
+    onChange({ ...value, mapKey: selected.value || '' });
+  };
+
+  const databaseTooltip = labels.components.DatabaseSelect.tooltip;
+  const tableTooltip = labels.components.TableSelect.tooltip;
+
+  const showTable = maxLevel >= 1;
+  const showColumn = maxLevel >= 2;
+  const showMapKey = maxLevel >= 3 && isMapColumn && Boolean(value.column);
+
+  return (
+    <>
+      <div className="gf-form">
+        <InlineFormLabel width={10} className="query-keyword" tooltip={databaseTooltip}>
+          {labelText('database', labels.components.DatabaseSelect.label)}
+        </InlineFormLabel>
+        <Select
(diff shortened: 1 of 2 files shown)


alex-fedotyev:
@SpencerTorres @kevinwcyu @karl-power requesting review when you have a moment. Could a maintainer also apply the `changelog` label so this surfaces in release notes? Thanks.

alex-fedotyev:
Filed [#1833 (Friendlier dashboard variables for authoring)](https://github.com/grafana/clickhouse-datasource/issues/1833) as the customer-facing use case this picker unblocks. The guided variable editor described there consumes `SchemaPicker` at `level='mapKey'` to give authors first-class "values of a Map key" variables for OTel resource attributes.

SpencerTorres (approved):

alyssajoyner (dismissed):

alex-fedotyev:
@adamyeats - should I do anything with this (sign, rebase) or not since it was approved already?

adamyeats:
@alex-fedotyev Yes please to both! 👍

alex-fedotyev:
@adamyeats done. Rebased onto the latest `main` and re-signed the commit; the head is now `ffb18bb` and shows the Verified badge. The diff is unchanged from what was approved (same single SchemaPicker commit, just replayed on current `main`). CI is re-running on the new SHA.

src/components/queryBuilder/SchemaPicker.tsx
```diff
@@ -0,0 +1,172 @@
...
+  const databaseTooltip = labels.components.DatabaseSelect.tooltip;
+  const tableTooltip = labels.components.TableSelect.tooltip;
+
+  const showTable = maxLevel >= 1;
+  const showColumn = maxLevel >= 2;
+  const showMapKey = maxLevel >= 3 && isMapColumn && Boolean(value.column);
+
+  return (
+    <>
+      <div className="gf-form">
```
adamyeats: `gf-form` is deprecated and shouldn't be used. Use the available [Form components](https://developers.grafana.com/ui/) to replace it. Also see https://github.com/grafana/grafana/blob/main/packages/grafana-ui/src/components/Forms/Form.tsx#L20. Note that we have a TODO to remove the existing usages of `gf-form` in this repo soon.

src/components/queryBuilder/SchemaPicker.tsx
```diff
@@ -0,0 +1,172 @@
...
+    onChange({ ...value, table: selected.value || '', column: '', mapKey: '' });
+  };
+
+  const handleColumnChange = (selected: SelectableValue<string>) => {
+    onChange({ ...value, column: selected.value || '', mapKey: '' });
+  };
+
+  const handleMapKeyChange = (selected: SelectableValue<string>) => {
+    onChange({ ...value, mapKey: selected.value || '' });
+  };
```
adamyeats: All four `onChange` handlers dereference `selected.value` with no null guard, while every `Select` below them is `isClearable`. On clear, Grafana passes `selected === null`, so `selected.value` throws `TypeError: Cannot read properties of null (reading 'value')`. Can we add some guards against `null` here?

src/components/queryBuilder/SchemaPicker.tsx
```diff
@@ -0,0 +1,172 @@
...
+  return (
+    <>
+      <div className="gf-form">
+        <InlineFormLabel width={10} className="query-keyword" tooltip={databaseTooltip}>
+          {labelText('database', labels.components.DatabaseSelect.label)}
+        </InlineFormLabel>
+        <Select
+          className={`width-15 ${styles.Common.inlineSelect}`}
+          options={databases.map((d) => ({ label: d, value: d }))}
+          value={value.database || null}
```
adamyeats: The four `value=` props pass a bare string, which `SelectBase` drops via `cleanValue` when it isn't in `options`. Each of these needs the existing components' fallback of appending the current value to `options` when it's missing from the fetched list.

adamyeats (changes requested):
Tested in the browser with a harness and this works! However, the Database clear (✕) button produced a crash that is reproducible. Added this and some other comments below.

src/components/queryBuilder/SchemaPicker.tsx
alex-fedotyev: Good catch. Fixed in 621444c. All four handlers now treat the `null` that Grafana passes on clear as an empty selection, and clearing a field also clears its downstream selections. Added regression tests for clearing the database and the column.

src/components/queryBuilder/SchemaPicker.tsx
alex-fedotyev: Done in 621444c. Each option list now appends the current value when the fetched list doesn't include it, matching the fallback already in `DatabaseSelect` / `TableSelect` / `ColumnSelect`.

src/components/queryBuilder/SchemaPicker.tsx
alex-fedotyev: Done in 621444c. Replaced the `gf-form` / `InlineFormLabel` layout with `InlineField` / `InlineFieldRow`.

alex-fedotyev:
@adamyeats addressed all three in 621444c:

- Null guards on every `onChange`, so the Database clear button no longer crashes; clearing a field clears the fields below it too.
- Each `Select` now keeps its current value visible when it isn't in the fetched list, matching the existing schema selects.
- Swapped the `gf-form` layout for `InlineField` / `InlineFieldRow`.

Added regression tests for the clear path. Mind taking another look when you get a chance?

adamyeats (approved):
Changes look good and tested locally, LGTM! 🚀

PR #1830, opened 2026-05-07
fix(config): trim whitespace from host setting

### LINKED ISSUE #1823: Extra space in Server address is breaking the form submission.
**What happened**: Because of extra space in server address input. User can't submit the form.

**What you expected to happen**: Form should submit even though there is a space in server address input.

**How to reproduce it (as minimally and precisely as possible)**: Add some space in server address input and try to click Save and test button

**Screenshots**

Image

Image

**Environment**:

- Grafana version: Grafana v12.4.0 (d1729c53a7)
- Plugin version:
- OS Grafana is installed on: MacOs
- User OS & Browser: Chrome
- Others:

alex-fedotyev:
## Summary

When the user typed a leading or trailing space in the Server address field, Save and Test would fail with a confusing connection error instead of a clear validation message. A whitespace-only host slipped past both the frontend's `if (!jsonData.host)` check (truthy in JS) and the backend's `Host == ""` check, so the request hit the ClickHouse driver with a malformed hostname and died there.

This fixes it in `LoadSettings` by trimming whitespace when reading both the v4 `host` field and the legacy v3 `server` field. One change covers V1 editor, V2 editor (gated on `newClickhouseConfigPageDesign`), provisioning YAML, and any API-driven config.

After the fix:
- `"   "` -> trimmed to `""` -> existing `ErrorMessageInvalidHost` triggers with a clear message
- `"  ch.example.com  "` -> trimmed to `"ch.example.com"` -> connection works

The pattern matches existing whitespace handling in the same file (`headerName = strings.TrimSpace(...)`) and on the frontend (`HttpHeadersConfigV2.tsx`, `AliasTableConfigV2.tsx`).

Fixes #1823

## Test plan

- [x] `go test ./pkg/plugin/... -run TestLoadSettings -v` passes (added 2 positive cases for v4 `host` and v3 `server` trimming, 1 negative case for whitespace-only host)
- [x] `go test ./pkg/...` full suite green
- [x] `go vet ./pkg/...` clean

### DIFF
--- a/pkg/plugin/settings.go
+++ b/pkg/plugin/settings.go
@@ -85,10 +85,10 @@ func LoadSettings(ctx context.Context, config backend.DataSourceInstanceSettings
 
 	// Deprecated: Replaced with Host for v4. Deserializes "server" field for old v3 configs.
 	if jsonData["server"] != nil {
-		settings.Host = jsonData["server"].(string)
+		settings.Host = strings.TrimSpace(jsonData["server"].(string))
 	}
 	if jsonData["host"] != nil {
-		settings.Host = jsonData["host"].(string)
+		settings.Host = strings.TrimSpace(jsonData["host"].(string))
 	}
 
 	if jsonData["port"] != nil {
--- a/pkg/plugin/settings_test.go
+++ b/pkg/plugin/settings_test.go
@@ -294,6 +294,50 @@ func TestLoadSettings(t *testing.T) {
 				wantErr: nil,
 				testCtx: ctx,
 			},
+			{
+				name: "should trim whitespace from host",
+				args: args{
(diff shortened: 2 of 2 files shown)


SpencerTorres (approved):
Nice 👌

adamyeats (approved):
LGTM 🚀

alex-fedotyev:
Bumping CI: the compatibility check failed because npm resolved vitest@3.2.5 (broken vite-node@3.2.5 dep). vitest@3.2.6 (with vite-node@3.2.4) published shortly after the run cutoff. Reopening to retrigger.

SpencerTorres (approved):

PR #1831, opened 2026-05-07
fix(config-v2): trim whitespace from host on blur

### LINKED ISSUE #1823: Extra space in Server address is breaking the form submission.

alex-fedotyev:
## Summary

Companion to #1830. The backend `LoadSettings` fix in that PR catches whitespace at submission time so Save and Test stops failing silently, but the V2 config editor's frontend validation still treats whitespace-only as valid input (truthy `if (jsonData.host)` check). That means whitespace-only or whitespace-padded values pass inline validation without any visual feedback.

This PR fixes the V2 frontend behavior:

- The validation effect's host checks use `.trim()` so whitespace-only is treated as empty.
- The host input's onBlur trims `e.currentTarget.value` and persists the cleaned value back via `onOptionsChange`. Whitespace-only triggers the existing "Server address required" inline error.

V2 is gated on the `newClickhouseConfigPageDesign` feature flag and is the upcoming default config experience. Pattern matches existing whitespace handling in this directory: `HttpHeadersConfigV2.tsx:35` (`header.name.trim()`) and `AliasTableConfigV2.tsx:58-61` (alias entry trim).

Fixes #1823 (in combination with #1830 on the backend).

## Test plan

- [x] `yarn jest --testPathPattern='config-v2/ServerAndEncryptionSection'` passes (10/10, 2 new cases)
  - trims leading and trailing whitespace from host on blur
  - shows "Server address required" when host is whitespace-only on blur
- [x] `yarn jest --testPathPattern='config-v2'` full V2 suite green (45/45)
- [x] `yarn lint` no errors (69 pre-existing warnings unchanged)

## Note on V1

The legacy V1 editor (`src/views/CHConfigEditor.tsx`) has the same JS truthy bug but is the deprecated default and will go away when the V2 feature flag becomes the default. The backend fix in #1830 covers it at submission time, so I'm not touching V1 here.

### DIFF
--- a/src/views/config-v2/ServerAndEncryptionSection.test.tsx
+++ b/src/views/config-v2/ServerAndEncryptionSection.test.tsx
@@ -45,6 +45,60 @@ describe('ServerAndEncryptionSection', () => {
     expect(onOptionsChangeMock).toHaveBeenCalled();
   });
 
+  it('trims leading and trailing whitespace from host on blur', () => {
+    const props = createTestProps({
+      options: {
+        jsonData: {
+          host: '  clickhouse-example.com  ',
+          secure: false,
+          protocol: Protocol.Native,
+          port: undefined,
+          pdcInjected: false,
+        },
+        secureJsonData: {},
+        secureJsonFields: {},
+      },
+      mocks: {
+        onOptionsChange: onOptionsChangeMock,
+      },
+    });
+    render(<ServerAndEncryptionSection {...props} />);
+
+    const input = screen.getByTestId('clickhouse-v2-config-host-input');
+    fireEvent.blur(input);
+
+    expect(onOptionsChangeMock).toHaveBeenCalledWith(
+      expect.objectContaining({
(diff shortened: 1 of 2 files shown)


SpencerTorres (approved):

SpencerTorres (approved):

alex-fedotyev:
Rebased onto main to clear the merge conflict. It was a formatting overlap in ServerAndEncryptionSection.tsx; the trim behavior is unchanged from the previously reviewed version. CI is green and the PR is mergeable again, so it is ready for another look when you have a chance.

adamyeats (approved):

PR #1836, opened 2026-05-12
set enable scripts false in .npmrc

itsgareth:
this repo uses npm (source: package-lock). updated the packageManager reference and set `ignore-scripts=true`

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,5 +111,5 @@
     "sql-formatter": "^15.6.12",
     "tslib": "2.8.1"
   },
-  "packageManager": "npm"
+  "packageManager": "npm@11.12.1"
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -111,5 +111,5 @@
     "sql-formatter": "^15.6.12",
     "tslib": "2.8.1"
   },
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+  "packageManager": "npm"
```
itsmylife: Can you set the exact npm version? Like `npm@<version-number>` I see repository prefers node v22. So it must be `npm@11.12.1` but please double check.

itsmylife (approved):

PR #1840, opened 2026-05-17
fix: scope variables support with the datasource

### LINKED ISSUE #122553: Section (tab) variables cannot reference other section variables in the same tab
### What happened?

I am using Grafana 13 with the feature for section-level variables enabled ('dashboardSectionVariables'):
https://github.com/grafana/grafana/pull/119572

When using section (tab) variables, a variable defined in a tab cannot reference another variable from the same tab.

Specifically:
- Variable A: 'namespace'
- Variable B: 'serviceName', which depends on 'namespace'

Variable B does not resolve '${namespace}' in its query.

However, '${namespace}' works correctly when used inside panel queries within the same tab.

### What did you expect to happen?

I expect section-level variables to support chaining (variable dependencies), similar to dashboard-level template variables.

Variable B should resolve '${namespace}' and refresh its values when 'namespace' changes.

Since section variables are scoped to a tab, variables within the same tab should be able to reference each other.

### Did this work before?

This is a new feature (section-level variables), and I have not found a version where this worked correctly.

### How do we reproduce it?

1. Enable feature flag:
   dashboardSectionVariables

2. Create a dashboard with tabs (sections)

3. In one tab, create variable A:
   SELECT DISTINCT namespace FROM your_table

4. In the same tab, create variable B:
   SELECT DISTINCT serviceName
   FROM your_table
   WHERE namespace = '${namespace}'

5. Create a panel in the same tab:
   SELECT *
   FROM your_table
   WHERE namespace = '${namespace}'

6. Change the value of 'namespace'

Result:
- Panel updates correctly
- Variable B does not update and does not resolve '${namespace}'

### Is the bug inside a dashboard panel?

_No response_

### Environment (with versions)?

Grafana:13.0.0

### Grafana platform?

None

### Datasource(s)?

_No response_

Orenico10:
## Type of Change

- [ ] 🚀 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧹 Refactor / Chore

---

### What is the bug?

Referencing Section vars in query doesn't work, the bug is explained here: [#12253](https://github.com/grafana/grafana/issues/122553)

### How to reproduce

1. Create a dashboard with the new dashboardSectionVariables enabled
2. Create a query variable that references a scope variable 
3. The query returns full without the filtering

### Related Issues

Closes https://github.com/grafana/grafana/issues/122553

## Please check that:

### DIFF
--- a/src/data/CHDatasource.test.ts
+++ b/src/data/CHDatasource.test.ts
@@ -69,6 +69,21 @@ describe('ClickHouseDatasource', () => {
       const values = await createInstance({ queryResponse }).metricFindQuery('mock', {});
       expect(values).toEqual(expectedValues);
     });
+
+    it('forwards scopedVars from options to the query request', async () => {
+      const queryResponse = {
+        fields: [{ name: 'field', type: 'string', values: ['val1'] }],
+      };
+      const instance = createInstance({ queryResponse });
+      const querySpy = jest.spyOn(instance, 'query');
+      const scopedVars = { namespace: { text: 'prod', value: 'prod' } };
+
+      await instance.metricFindQuery('SELECT 1', { scopedVars });
+
+      expect(querySpy).toHaveBeenCalledWith(
+        expect.objectContaining({ scopedVars })
+      );
+    });
   });
 
   describe('applyTemplateVariables', () => {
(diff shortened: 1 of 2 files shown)


aangelisc (approved):
Thank you for the contribution @Orenico10 👍

PR #1841, opened 2026-05-18
feat: compact query mode for single table data sources

fleon:
## Type of Change

- [x] 🚀 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧹 Refactor / Chore

---

### What is this feature?

Adds a compact query editor for single-table logs and traces datasources. The compact editor provides a focused logs/traces workflow with search, filter controls, advanced options, query history access, and a generated SQL preview with copy and edit-as-SQL actions.

The compact filter controls also hide builder-owned timestamp columns and tailor available operators based on inferred field type: numeric fields get comparison operators, while string fields get contains/equality operators.

Single-table logs and traces datasources do not need the full database/table/query-type builder surface. A compact editor keeps common Explore workflows faster and reduces confusing filter options, while preserving access to the generated SQL when users need to inspect or customize it.

### Who is this feature for?

Users querying logs or traces from a datasource configured for a focused single-table setup.

### How to test this feature

1. Configure a datasource in single-table mode for logs or traces.
2. Open Explore and confirm the compact query editor renders instead of the full database/table/query-type builder.
3. For logs, use the search input and add filters from the compact filter bar.
4. Confirm configured time columns are not offered in the Add filter column list.
5. Confirm numeric fields show numeric comparison operators and string fields show contains/equality operators.
6. Open the SQL preview, copy the generated SQL, and use Edit as SQL to switch to the SQL editor.

---

## Screenshots / Videos

|Case | Screenshot |
| ------ | ----- |
| Compact logs view  |    CleanShot 2026-05-28 at 17 16 26@2x |
| Compact traces view | CleanShot 2026-05-28 at 17 17 17@2x |

---

## Please check that:

### DIFF
--- a/src/components/queryBuilder/FilterPopover.tsx
+++ b/src/components/queryBuilder/FilterPopover.tsx
@@ -1,0 +1,256 @@
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import { css } from '@emotion/css';
+import { GrafanaTheme2, SelectableValue } from '@grafana/data';
+import { AsyncSelect, Button, Select, useStyles2 } from '@grafana/ui';
+import { Datasource } from 'data/CHDatasource';
+import { Filter, FilterOperator, NumberFilter, StringFilter, TableColumn } from 'types/queryBuilder';
+
+interface FilterPopoverProps {
+  datasource: Datasource;
+  database: string;
+  table: string;
+  allColumns: readonly TableColumn[];
+  onAddFilter: (filter: Filter) => void;
+  onClose: () => void;
+}
+
+type FilterValueKind = 'number' | 'string';
+
+const numberOperatorOptions: Array<SelectableValue<FilterOperator>> = [
+  { label: '>', value: FilterOperator.GreaterThan },
+  { label: '<', value: FilterOperator.LessThan },
+  { label: '>=', value: FilterOperator.GreaterThanOrEqual },
+  { label: '<=', value: FilterOperator.LessThanOrEqual },
+  { label: '=', value: FilterOperator.Equals },
+  { label: '!=', value: FilterOperator.NotEquals },
+  { label: 'IS NULL', value: FilterOperator.IsNull },
+  { label: 'IS NOT NULL', value: FilterOperator.IsNotNull },
+];
+
+const stringOperatorOptions: Array<SelectableValue<FilterOperator>> = [
+  { label: 'contains', value: FilterOperator.Like },
+  { label: 'does not contain', value: FilterOperator.NotLike },
+  { label: '=', value: FilterOperator.Equals },
+  { label: '!=', value: FilterOperator.NotEquals },
+  { label: 'IS NULL', value: FilterOperator.IsNull },
+  { label: 'IS NOT NULL', value: FilterOperator.IsNotNull },
+];
+
+const defaultOperatorByKind: Record<FilterValueKind, FilterOperator> = {
+  number: FilterOperator.GreaterThan,
+  string: FilterOperator.Like,
+};
+
+const getMapValueType = (type: string): string => {
+  return type.match(/Map\(\s*.+\s*,\s*(.+)\s*\)/)?.[1]?.trim() || type;
+};
+
+export const getFilterValueKind = (type = ''): FilterValueKind => {
+  const normalizedType = getMapValueType(type)
+    .toLowerCase()
+    .replace(/\(/g, '')
+    .replace(/\)/g, '')
+    .replace(/nullable/g, '')
+    .replace(/lowcardinality/g, '');
+
+  return ['int', 'float', 'decimal'].some((numberType) => normalizedType.includes(numberType)) ? 'number' : 'string';
+};
+
+export const getOperatorOptions = (kind: FilterValueKind): Array<SelectableValue<FilterOperator>> => {
+  return kind === 'number' ? numberOperatorOptions : stringOperatorOptions;
+};
+
+const getStyles = (theme: GrafanaTheme2) => ({
+  popover: css`
+    display: flex;
+    align-items: flex-end;
+    gap: ${theme.spacing(0.75)};
+    padding: ${theme.spacing(0.75)} 0;
+    flex-wrap: wrap;
+  `,
+  field: css`
+    display: flex;
+    flex-direction: column;
+    gap: 2px;
+  `,
+  fieldLabel: css`
+    font-size: 11px;
+    color: ${theme.colors.text.secondary};
+    font-weight: ${theme.typography.fontWeightMedium};
+  `,
+  actions: css`
+    display: flex;
+    gap: ${theme.spacing(0.5)};
+  `,
+});
+
+export const FilterPopover = (props: FilterPopoverProps) => {
+  const { datasource, database, table, allColumns, onAddFilter, onClose } = props;
+  const styles = useStyles2(getStyles);
+
+  const [selectedColumn, setSelectedColumn] = useState('');
+  const [selectedMapKey, setSelectedMapKey] = useState('');
+  const [operator, setOperator] = useState<FilterOperator>(FilterOperator.Equals);
+  const [value, setValue] = useState('');
+  const [mapKeys, setMapKeys] = useState<string[]>([]);
+
+  const selectedColDef = allColumns.find((column) => column.name === selectedColumn);
+  const isMapColumn = selectedColDef?.type?.startsWith('Map(') || false;
+  const filterKind = getFilterValueKind(selectedColDef?.type);
+  const currentOperatorOptions = useMemo(() => getOperatorOptions(filterKind), [filterKind]);
+
+  useEffect(() => {
+    if (isMapColumn && selectedColumn && database && table) {
+      datasource
+        .fetchUniqueMapKeys(selectedColumn, database, table)
+        .then(setMapKeys)
+        .catch(() => setMapKeys([]));
+    } else {
+      setMapKeys([]);
+      setSelectedMapKey('');
+    }
+  }, [datasource, database, table, selectedColumn, isMapColumn]);
+
+  const columnOptions: Array<SelectableValue<string>> = allColumns.map((column) => ({
+    label: column.label || column.name,
+    value: column.name,
+    description: column.type,
+  }));
+
+  const loadValueOptions = useCallback(
+    async (inputValue: string): Promise<Array<SelectableValue<string>>> => {
+      if (!selectedColumn || !database || !table) {
+        return [];
+      }
+      try {
+        const values =
+          isMapColumn && selectedMapKey
+            ? await datasource.fetchDistinctMapValues(selectedColumn, selectedMapKey, database, table)
+            : !isMapColumn
+              ? await datasource.fetchDistinctValues(selectedColumn, database, table)
+              : [];
+
+        return values
+          .filter((nextValue) => !inputValue || nextValue.toLowerCase().includes(inputValue.toLowerCase()))
+          .map((nextValue) => ({ label: nextValue, value: nextValue }));
+      } catch (err) {
+        console.error('FilterPopover: failed to load values for column', selectedColumn, err);
+        return [];
+      }
+    },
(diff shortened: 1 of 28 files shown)


alex-fedotyev:
@fleon - I reviewed the UI, very nice!
Could you please adjust couple things?

Collapse SQL section by default to save space and confuse new users less:
image

When switching to SQL mode, could you select view by default to be `traces` for traces datasource and same for `logs`? Right now it shows `table` by default.
image

alex-fedotyev:
@fleon - is there a way to take advantage of bigger screen for traces? Now there is more space, it looks strange to have grid not dynamic to the available space.
image

fleon:
@alex-fedotyev great catches. Fixed the first two.

1. SQL section stays collapsed by default in compact query mode.
2. SQL mode in Traces view defaults to traces.

However for the traces table not taking up full height, that CSS doesn't belong in the plugin. This is something that would need to be fixed in the main grafana repo.

EDIT: On a second note, I would avoid fixing the height issue altogether. In case people have multiple queries, keeping the table as fixed height makes better sense. Alternatively, we could make it set fluid height, but restrict min-height to something specific.

CleanShot 2026-05-26 at 12 30 44@2x

adamyeats:
@fleon  Please rebase and sign your commits

fleon:
@adamyeats Done!

src/views/CHQueryEditor.tsx
```diff
@@ -60,10 +73,24 @@ const CHEditorByType = (props: CHQueryEditorProps) => {
...
   lastEditorType.current = query.editorType;
 
-  // Prevent trying to run empty query on load
+  const propBuilderOptions =
+    query.editorType === EditorType.Builder ? (query as CHBuilderQuery).builderOptions : undefined;
+  const lastPropBuilderOptions = useRef<QueryBuilderOptions | undefined>(propBuilderOptions);
+  if (propBuilderOptions && propBuilderOptions !== lastPropBuilderOptions.current) {
+    lastPropBuilderOptions.current = propBuilderOptions;
+    if (propBuilderOptions !== builderOptions) {
+      builderOptionsDispatch(setAllOptions(propBuilderOptions));
```
adamyeats, line 92: This is an infinite render loop that crashes the compact editor (`React #185`, max update depth) on Grafana 11.6 and 12.2; works only on 13.x. This re-dispatches `setAllOptions` whenever Explore returns a fresh `builderOptions` reference, which re-fires the `[builderOptions]` onChange effect. 

Compare by value (or don't re-dispatch the options just emitted) so it doesn't depend on host object-identity behaviour.

src/components/queryBuilder/FilterPopover.tsx
```diff
@@ -0,0 +1,256 @@
...
+      try {
+        const values =
+          isMapColumn && selectedMapKey
+            ? await datasource.fetchDistinctMapValues(selectedColumn, selectedMapKey, database, table)
+            : !isMapColumn
+              ? await datasource.fetchDistinctValues(selectedColumn, database, table)
+              : [];
+
+        return values
+          .filter((nextValue) => !inputValue || nextValue.toLowerCase().includes(inputValue.toLowerCase()))
```
adamyeats: `nextValue.toLowerCase()` throws on numeric columns (`values()` returns numbers, not strings). Typing in the value box empties the dropdown. Coerce with `String(nextValue)`. The `Promise<string[]>` annotation seems to also be incorrect.

src/data/CHDatasource.ts
```diff
@@ -944,6 +993,16 @@ export class Datasource
     return this.fetchData(rawSql);
   }
 
+  async fetchDistinctValues(column: string, db: string, table: string): Promise<string[]> {
```
adamyeats: `fetchDistinctValues`/`fetchDistinctMapValues` interpolate identifiers and `mapKey` with no escaping. Every sibling fetcher uses `escapeIdentifier`. A map key with a `'` breaks the query / is an injection vector.

src/data/CHDatasource.ts
```diff
@@ -548,7 +556,7 @@ export class Datasource
         key: column && column.hint ? '' : columnName,
         hint: column && column.hint ? column.hint : undefined,
         mapKey: hasMapKey ? mapKey : undefined,
-        type: hasMapKey ? (columnType.startsWith('Map') ? 'Map(String, String)' : 'JSON') : 'String',
+        type: hasMapKey ? (columnType.startsWith('JSON') ? 'JSON' : 'Map(String, String)') : 'String',
```
adamyeats: An untyped attributes column now defaults to `Map` instead of `JSON`, is that intentional? Will this have impact anywhere else?

src/data/CHDatasource.ts
```diff
@@ -502,11 +502,19 @@ export class Datasource
...
-      const prefixIndex = columnName.indexOf('.');
-      mapKey = columnName.substring(prefixIndex + 1);
-      columnName = columnName.substring(0, prefixIndex);
+    // Convert flattened/merged OTel attributes into column+path pair.
+    // Grafana log details can emit semantic, normalized keys like
+    // `log_attributes.log.file.path`, while the query may use a configured
+    // column name like `LogAttributes`. Resolve the prefix through selected
+    // column hints/names/aliases so compact and classic builders behave alike.
+    const prefixIndex = columnName.indexOf('.');
+    if (prefixIndex > -1) {
```
adamyeats: Attribute-prefix resolution now only splits when the column is in `builderOptions.columns`. A filter like `LogAttributes.foo.bar` whose column isn't selected no longer becomes a map-path filter (regresses the old unconditional split).

src/views/CHQueryEditor.tsx
```diff
@@ -128,6 +195,44 @@ const CHEditorByType = (props: CHQueryEditorProps) => {
...
+      builderOptions,
+      format: mapQueryBuilderOptionsToGrafanaFormat(builderOptions),
+    });
+  };
+
+  return (
+    <>
+      <div className={'gf-form ' + styles.QueryEditor.queryType}>
+        <Button variant="secondary" onClick={onSwitchToBuilder}>
+          Switch to compact view
```
adamyeats, line 255: "Switch to compact view" overwrites `rawSql` with regenerated defaults, silently discarding hand-edited SQL. Add a confirm or preserve the query.

src/views/CHQueryEditor.tsx
```diff
@@ -128,6 +195,44 @@ const CHEditorByType = (props: CHQueryEditorProps) => {
...
+      editorType: EditorType.Builder,
+      rawSql: generateSql(builderOptions),
+      builderOptions,
+      format: mapQueryBuilderOptionsToGrafanaFormat(builderOptions),
+    });
+  };
+
+  return (
+    <>
+      <div className={'gf-form ' + styles.QueryEditor.queryType}>
```
adamyeats: `gf-form` is deprecated and shouldn't be used. Use the available [Form components](https://developers.grafana.com/ui/) to replace it. Also see https://github.com/grafana/grafana/blob/main/packages/grafana-ui/src/components/Forms/Form.tsx#L20. Note that we have a TODO to remove the existing usages of `gf-form` in this repo soon.

src/components/queryBuilder/FilterPopover.tsx
```diff
@@ -0,0 +1,256 @@
...
+  database: string;
+  table: string;
+  allColumns: readonly TableColumn[];
+  onAddFilter: (filter: Filter) => void;
+  onClose: () => void;
+}
+
+type FilterValueKind = 'number' | 'string';
+
+const numberOperatorOptions: Array<SelectableValue<FilterOperator>> = [
```
adamyeats: `numberOperatorOptions`/`stringOperatorOptions` duplicate the operator/label tables in `FilterEditor.tsx`. Reuse the shared source if you can.

src/components/queryBuilder/FilterPopover.tsx
```diff
@@ -0,0 +1,256 @@
...
+const defaultOperatorByKind: Record<FilterValueKind, FilterOperator> = {
+  number: FilterOperator.GreaterThan,
+  string: FilterOperator.Like,
+};
+
+const getMapValueType = (type: string): string => {
+  return type.match(/Map\(\s*.+\s*,\s*(.+)\s*\)/)?.[1]?.trim() || type;
+};
+
+export const getFilterValueKind = (type = ''): FilterValueKind => {
```
adamyeats, line 55: Re-implements `isNumberType`/`isStringType` from `queryBuilder/utils.ts`

src/components/queryBuilder/compactQueryDefaults.ts
```diff
@@ -0,0 +1,189 @@
...
+    hint: ColumnHint.LogLevel,
+    condition: 'AND',
+    value: '',
+  } as StringFilter,
+];
+
+const getLogsDefaultOrderBy = (): OrderBy[] => [
+  { name: '', hint: ColumnHint.FilterTime, dir: OrderByDirection.DESC, default: true },
+  { name: '', hint: ColumnHint.Time, dir: OrderByDirection.DESC, default: true },
+];
```
adamyeats: `getLogsDefaultFilters`/`getLogsDefaultOrderBy` (+ traces equivalents) are verbatim copies of the arrays in `logsQueryBuilderHooks.ts`/`traceQueryBuilderHooks.ts`. Let's consolidate these so classic vs compact defaults don't drift.

src/components/queryBuilder/QueryBuilder.tsx
```diff
@@ -96,6 +132,129 @@ export const QueryBuilder = (props: QueryBuilderProps) => {
...
+          ...nextOptions.meta,
+        },
+      },
+      shouldRunQuery
+    );
+  };
+
+  return (
+    <div data-testid="query-editor-section-builder">
+      <CompactModeBar
```
adamyeats, line 228: `CompactModeBar` is called with no-op `onModeChange`/`onSearchSubmit` and a hardcoded `mode`, so the whole mode-dropdown branch (`getModeOptions`, `raw-sql`) is unreachable dead code.

adamyeats (changes requested):
@fleon Nice UX improvement, and it works on latest Grafana. 🚀  Two blockers before merge, though.

- **1. The compact editor crashes on older Grafana.** The error boundary shows `React error #185` (max update depth) on 11.6 and 12.2; works on 13.x.
- **2. This needs E2E coverage.** CI is green only because no E2E test exercises the compact path against a real Grafana. The unit tests mock `onChange`/`dispatch`, so the render loop never forms. Could we add an E2E spec that opens a single-table datasource in Explore (run against the floor version)?

src/components/queryBuilder/QueryBuilder.tsx
fleon: Removed.

src/views/CHQueryEditor.tsx
fleon: Done. Also replaced AsyncSelect/MultiSelect/Select with Combobox.

src/data/CHDatasource.ts
fleon: Fixed

src/data/CHDatasource.ts
fleon: Fixed. Attribute prefix resolution now keeps the normalized prefix matching path, but falls back to the legacy inbuilt attributes split when the attribute column is not selected. I also restored the previous untyped map-path fallback to JSON, while still preserving Map and JSON types when the selected column type is known.

src/components/queryBuilder/FilterPopover.tsx
fleon: Added escaping, good catch 👍

src/views/CHQueryEditor.tsx
fleon: Good catch. Can confirm this was an issue that could only be fixed by refreshing the page. Fixed now.

fleon:
@adamyeats Thanks for the review. I've addressed your comments now. :)

adamyeats (approved):
Looks excellent! LGTM 🚀

PR #1847, opened 2026-05-26
chore(security): hardening pass — renovate cooldown, dep reduction, workflow least-privilege

adamyeats:
## Type of Change

- [x] 🧹 Refactor / Chore

---

## What this changes

A repo-wide security hardening pass covering three areas: dependency reduction, package-management hardening, and GitHub Actions least-privilege. No behavior change in the plugin itself.

### Renovate (highest-impact)

The repo extends only `presets/automerge` from `grafana-renovate-config`. None of the other presets (`npm`, `github-actions`, `rate-limits`, `base`, …) were being applied. The local `packageRules` that disable `minimumReleaseAge` for `@grafana/*` packages were no-ops because there was no baseline cooldown to override.

- Extend the missing presets so a 3-day release-age cooldown applies to npm and GitHub Actions, action versions get digest-pinned, and `gomodTidy` runs on Go updates.
- Add an explicit rule applying `minimumReleaseAge: \"3 days\"` to non-Grafana **devDependencies** — the upstream npm preset only covers production deps.
- Restrict the `automerge-minor` rule to `devDependencies` only. Production minor bumps (`@grafana/data`, `pgsql-ast-parser`, etc.) now require human review.

### Dependency reduction

| Package | Action | Reason |
|---|---|---|
| `@openfeature/web-sdk` | Removed from `dependencies` | Zero imports; still pulled transitively by `@grafana/runtime` |
| `react-router-dom`, `@types/react-router-dom` | Removed | Not imported; host-provided via webpack externals; still pulled transitively by `@grafana/ui` |
| `es-toolkit` | Removed | Single usage of `isEqual` swapped to `lodash.isEqual`; lodash is a webpack external (zero bundle cost) |
| `semver`, `@types/semver` | Removed | Two `gte()` call sites swapped to the existing internal `isVersionGtOrEq` helper in `src/utils/version.ts` |
| `github.com/pkg/errors` | Removed | Frozen since 2020; only `errors.New`/`Is`/`As` used, all in stdlib |

### Package management

- `\"sign\": \"npx --yes @grafana/sign-plugin@latest\"` → `\"npx --package=@grafana/sign-plugin@3.2.2 -- sign-plugin\"`. Removes `@latest` resolution at publish time.

### GitHub Actions

- Added explicit `permissions:` blocks to `detect-breaking-changes.yml`, `integration.yml`, `cron.yml`, and `issue_commands.yml` (previously inherited repo default scopes).
- `detect-breaking-changes.yml`: `npm install` → `npm ci`; trigger narrowed from `[push, pull_request]` to `pull_request` (removes duplicate PR-branch runs).
- `integration.yml`: `${{ matrix.clickhouse }}` moved from `run:` template to `env:` (avoids template-injection-shaped code even though the matrix is currently static).
- `issue_commands.yml`: SHA-pinned the `grafana/grafana-github-actions` checkout `ref:` (was `main`).
- `push.yml` and `cron.yml`: added policy comments explaining the intentional `grafana/data-sources-ci-workflows@main` refs (the shared-workflow repo doesn't tag releases for these yet; allowed by the existing `.github/zizmor.yml` policy).

---

## Please check that:

- [x] Tests for this change have been added/updated. (No behavior change; existing tests pass)
- [ ] Documentation has been added/updated (where applicable). (N/A)

Three things to look at carefully:

1. **`renovate.json` extends list ordering.** I put `automerge` last so its `packageRules` are applied after the npm/github-actions presets — Renovate merges preset packageRules in order, and we want the automerge rules to take effect after the cooldown rules. Worth a sanity check from someone who knows Renovate's merge semantics better than me.
2. **The `grafana-github-actions` SHA pin** in `issue_commands.yml` was set to the current `main` tip at the time of writing. Renovate's shared-workflows preset does not manage checkout `ref:` values, so this will need manual bumps. Acceptable trade-off, but flagging.
3. **`docs/`** has several untracked planning files unrelated to this PR — they were left out of this commit on purpose.

Verification done locally:
- `npm run typecheck` ✓
- `go build ./...`, `go vet ./...` ✓
- `go test ./pkg/plugin/...` (TestCategorizeConnectionError, 37 assertions) ✓
- `govulncheck ./...` → 0 reachable vulnerabilities ✓
- `cspell` on all changed files ✓


package.json
```diff
@@ -101,13 +99,9 @@
     "@grafana/runtime": "12.4.2",
     "@grafana/schema": "12.4.2",
     "@grafana/ui": "12.4.2",
-    "@openfeature/web-sdk": "^1.7.3",
```
aangelisc, line 104: I think we will still need this for feature toggles so best to keep?

aangelisc (dismissed):
LGTM, one question in the comments!

package.json
adamyeats: Good catch — restored in 56eb379. Agreed; even though it's pulled transitively via `@grafana/runtime`, keeping the direct declaration documents intent and gives us an explicit version pin for the upcoming feature-toggle work.

aangelisc (approved):

PR #1853, opened 2026-05-27
fix: trace optimisation defaults true

### LINKED ISSUE #1842: Regression: Trace View fails with incorrect dependency <trace_table>_trace_id_ts (works after manual re-run)
**What happened**:

Right-clicking on a TraceId and choosing "View Trace" fails to build the trace view with the following error:

```
error querying the database: code: 60, message: Unknown table expression identifier 'otel.traces_dist_trace_id_ts' in scope (SELECT min(Start) FROM otel.traces_dist_trace_id_ts WHERE TraceId = trace_id) AS trace_start
(Trace ID: 1234567890)
```

The plugin error is referencing a `<trace_table>_trace_id_ts` table as a dependency. This table is, however, **not referenced** by the SQL query that actually builds the trace view. Clicking the "Show full query" button confirms this. The SQL statement building the actual trace view is as follows:

```
SELECT "TraceId" as traceID, "SpanId" as spanID, "ParentSpanId" as parentSpanID, "ServiceName" as serviceName, "SpanName" as operationName, multiply(toUnixTimestamp64Nano("Timestamp"), 0.000001) as startTime, multiply("Duration", 0.000001) as duration, arrayMap(key -> map('key', key, 'value',"SpanAttributes"[key]), mapKeys("SpanAttributes")) as tags, arrayMap(key -> map('key', key, 'value',"ResourceAttributes"[key]), mapKeys("ResourceAttributes")) as serviceTags, if("StatusCode" IN ('Error', 'STATUS_CODE_ERROR'), 2, 0) as statusCode, arrayMap((name, timestamp, attributes) -> tuple(name, toString(toUnixTimestamp64Milli(timestamp)), arrayMap( key -> map('key', key, 'value', attributes[key]), mapKeys(attributes)))::Tuple(name String, timestamp String, fields Array(Map(String, String))), "Events".Name, "Events".Timestamp, "Events".Attributes) AS logs, arrayMap((traceID, spanID, attributes) -> tuple(traceID, spanID, arrayMap(key -> map('key', key, 'value', attributes[key]), mapKeys(attributes)))::Tuple(traceID String, spanID String, tags Array(Map(String, String))), "Links".TraceId, "Links".SpanId, "Links".Attributes) AS references, "SpanKind" as kind, "StatusMessage" as statusMessage, "TraceState" as traceState FROM "otel"."traces_dist" WHERE traceID = '1234567890'
````

Note, that this SQL query references only one table, in this case `traces_dist`. It DOES NOT require a `traces_dist_trace_id_ts` table. In fact, simply clicking "Run Query" after the error appears, builds the trace view correctly.

When clicking on "View Trace", the trace view builds without an error using the SQL statement above as it did, in fact, in version 4.14.

1. Install Grafana clickhouse datasource plugin 4.15 or higher.
2. Ensure you have no `<trace_table>_trace_id_ts` table in Clickhouse
3. Query trace table to get some TraceIds in the output. For example `SELECT * FROM "{{db}}.{{trace_table}}" LIMIT 1000
4. Right click on any TraceId
5. Choose "View Trace"
6. As the error appears, click "Run Query" to see the trace

**Screenshots**

See below

Version 4.14 worked, 4.15+ reintroduced references to a not required table.

**Environment**:

- Grafana version: Grafana v12.0.2+security-01 (76ea754dbb)
- Plugin version: 4.15+
- OS Grafana is installed on: Azure Cloud
- User OS & Browser: Linux/macOS/Windows & Chrome/Edge/Firefox
- Others:

Image
Image

### PR DESCRIPTION
**Pull request by karl-power**
## Type of Change

- [x] 🐛 Bug Fix

---

### What is the bug?

The "View trace" data link generated by trace search, logs, and table panels could fail on the first click when the OTel `_trace_id_ts` companion table was missing.

The root cause was in `transformQueryResponseWithTraceAndLogLinks` (`src/data/utils.ts`), which set `meta.hasTraceTimestampTable: Boolean(otelVersion)` on the generated link query. With OTel enabled, that value was always `true`, so the link's `rawSql` was generated with the two-step `WITH trace_start / trace_end` shape that joins against `<table>_trace_id_ts`. On environments without the companion, the query failed with "Unknown table expression identifier". On environments where the companion existed but had no row for the trace, the optimised SQL silently returned 0 rows (demonstrated by the new e2e test).

`CHQueryEditor` had a separate `useTables` + `useMemo` check that would eventually correct the `meta` value, but it ran inside the editor and depended on a table-list fetch completing. On a fresh page load — particularly when the editor was minimised on a deep-link — the link's `rawSql` was already generated by the response-transform path before the editor's effect ran. Two related issues also surfaced once the response transform was made the source of truth:

1. The trace→trace branch of the response transform spread `originalQuery.builderOptions.meta` and never set `hasTraceTimestampTable` itself, so it inherited whatever value was on the source query's meta rather than consulting the actual schema.
2. Neither the generated trace ID link query nor the generated View Logs link query had a `format` set. Without it, the sqlds backend tagged the response with the wrong `preferredVisualisationType` and Grafana could route the frame to the wrong panel before the editor's `setAllOptions` had a chance to run.

### How to reproduce

1. Configure a ClickHouse datasource with OTel tracing enabled, pointing at a trace table that has **no** `<table>_trace_id_ts` companion (or one where the trace you click has no row in the companion).
2. Run a logs or table query that surfaces trace IDs. Click **View trace** — the query fails with "Unknown table expression identifier" or returns 0 spans.
3. Run a trace search, then click **View trace** on a result row — the generated SQL is unoptimised even when the companion does exist.

### Related Issues

Closes #1842
Relates to #1705, #1786

---

## Please check that:

**Files changed**

`src/data/CHDatasource.ts` — Adds `async hasTraceTimestampTable(database, table): Promise<boolean>` on the datasource. Backed by a `Map<string, Promise<boolean>>` keyed by `${database}.${table}`. Caching the Promise (not the resolved value) means concurrent callers share a single `SHOW TABLES` round-trip; repeat callers resolve in a microtask. On fetch failure the entry is evicted so the next call retries, and the failing call resolves `false` (the safe, unoptimised path). The `Datasource.query()` pipe operator changes from `map` to `concatMap` so the now-async transform can be awaited before the response is emitted.

`src/data/utils.ts` — `transformQueryResponseWithTraceAndLogLinks` becomes `async`. Both branches (trace search → trace, and logs/table → trace) now `await datasource.hasTraceTimestampTable(...)` before generating `rawSql`, replacing the `Boolean(otelVersion)` default in the logs/table branch and adding an explicit lookup in the trace search branch (which previously inherited the value from the source query's meta). `format` is now set on both the trace ID link query and the View Logs link query via `mapQueryBuilderOptionsToGrafanaFormat`, so the response is tagged with the correct `preferredVisualisationType` from the first request.

`src/hooks/useHasTraceTimestampTable.ts` — New hook that wraps `datasource.hasTraceTimestampTable()` for use in the editor, with cancellation on unmount or dep change.

`src/views/CHQueryEditor.tsx` — Replaces the `useTables` + `useMemo` table-list check with `useHasTraceTimestampTable`, sharing the same datasource-level cache as the response transform. The `traceTables.length === 0` short-circuit in the effect is removed (no longer applicable now that the hook returns a stable boolean instead of deriving from a list that may be transiently empty).

**Tests**

- New unit tests on `Datasource.hasTraceTimestampTable` cover the empty-inputs case, present/missing companion, repeat-call caching, concurrent-call dedupe, error-evict-and-retry, and configured custom suffix.
- `src/data/utils.test.ts` is updated to `await` the now-async transform. New cases assert that the datasource lookup overrides stale `meta.hasTraceTimestampTable` (true→false and false→true), that the generated trace ID link carries `format: 3` (Trace) and the View Logs link carries `format: 2` (Logs), and that the logs→trace and table→trace branches both ship unoptimised SQL when the companion is missing (#1842).
- E2E (`tests/e2e/traceTimestampTable.spec.ts` + fixture `tests/e2e/fixtures/trace_id_ts.sql`) drives real ClickHouse with two seeded traces: `e2e-ts-trace-a` has a companion row, `e2e-ts-trace-b` deliberately does not. The tests assert that the optimised SQL returns all spans for trace-a, returns 0 rows for trace-b (the #1842 failure mode made concrete), and that the fallback SQL returns spans correctly for both. A final test confirms `SHOW TABLES` returns the companion, i.e. that `hasTraceTimestampTable` would resolve `true` for the seeded schema.

### DIFF
--- a/src/data/CHDatasource.ts
+++ b/src/data/CHDatasource.ts
@@ -742,7 +742,7 @@ export class Datasource
    * Resolves whether the `<table>_trace_id_ts` companion exists for the given
    * (database, table). Caches the Promise so concurrent and repeat callers share
    * a single `SHOW TABLES` round-trip; on failure, evicts so the next caller
-   * retries and meanwhile returns `false` (the safe, unoptimised path).
+   * retries and meanwhile returns `false` (the safe, unoptimized path).
    */
   async hasTraceTimestampTable(database: string, table: string): Promise<boolean> {
     if (!database || !table) {
--- a/src/data/CHDatasource.ts
+++ b/src/data/CHDatasource.ts
@@ -87,7 +87,10 @@ export class Datasource
   // callers and lets cache hits resolve in a microtask. Entries expire after
   // TRACE_TIMESTAMP_TABLE_CACHE_TTL_MS so a companion table created after the datasource
   // loaded is detected without a page reload.
-  private traceTimestampTableCache = new Map<string, { pending: Promise<boolean>; expiresAt: number }>();
+  private traceTimestampTableCache = new Map<
+    string,
+    { pending: Promise<boolean>; resolved?: boolean; expiresAt: number }
+  >();
 
   constructor(instanceSettings: DataSourceInstanceSettings<CHConfig>) {
     super(instanceSettings);
@@ -757,10 +760,6 @@ export class Datasource
 
     let entry = this.traceTimestampTableCache.get(key);
 
-    if (entry) {
-      console.log({ expired: entry.expiresAt <= now });
-    }
-
     if (!entry || entry.expiresAt <= now) {
       const pending = (async () => {
         try {
@@ -774,9 +773,33 @@ export class Datasource
 
       entry = { pending, expiresAt: now + Datasource.TRACE_TIMESTAMP_TABLE_CACHE_TTL_MS };
       this.traceTimestampTableCache.set(key, entry);
+
+      // Record the settled value so peekTraceTimestampTable() can read it synchronously.
+      const created = entry;
+      pending.then((v) => {
+        created.resolved = v;
+      });
     }
 
     return entry.pending;
+  }
+
+  /**
+   * Synchronously read the cached trace timestamp table result without awaiting.
+   * Returns the resolved boolean when a non-expired cache entry has settled, or
+   * undefined when there is no entry, it is still pending, or it has expired.
+   * Lets the React hook seed its initial state from a warm cache and skip a
+   * false→true render that would briefly clobber a known-good meta value.
+   */
+  peekTraceTimestampTable(database: string, table: string): boolean | undefined {
+    if (!database || !table) {
+      return undefined;
+    }
+    const entry = this.traceTimestampTableCache.get(`${database}.${table}`);
+    if (!entry || entry.expiresAt <= Date.now()) {
+      return undefined;
+    }
+    return entry.resolved;
   }
 
   /**
--- a/src/data/CHDatasource.ts
+++ b/src/data/CHDatasource.ts
@@ -27,7 +27,7 @@ import LogsContextPanel from 'components/LogsContextPanel';
 import { cloneDeep, isEmpty, isString } from 'lodash';
 import otel from 'otel';
 import { createElement as createReactElement, ReactNode } from 'react';
-import { firstValueFrom, from, map, mergeMap, Observable } from 'rxjs';
+import { concatMap, firstValueFrom, Observable } from 'rxjs';
 import { CHConfig } from 'types/config';
 import {
   AggregateColumn,
@@ -81,6 +81,10 @@ export class Datasource
   // `asMapAccess()` strips any `db.` prefix from the lookup source so callers
   // can pass either form.
   private mapColumnsByTable: Map<string, Set<string>> = new Map();
+  // Caches the in-flight or resolved existence check for the `<table>_trace_id_ts`
+  // companion, keyed by `${database}.${table}`. Caching the Promise dedupes
+  // concurrent callers and lets cache hits resolve in a microtask.
+  private traceTimestampTableCache = new Map<string, Promise<boolean>>();
 
   constructor(instanceSettings: DataSourceInstanceSettings<CHConfig>) {
     super(instanceSettings);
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Review by SpencerTorres — APPROVED** [review]

**Review by SpencerTorres — APPROVED** [review]

**Comment by s4njur0** [comment]
@SpencerTorres , @karl-power I don't understand exactly where the hold up is. Does it lie with Grafana? Could you nudge them maybe? We have customers on Clickhouse that are affected. It would be great to tell them when this works again.

File: src/data/CHDatasource.ts [hunk 1]
```diff
@@ -734,6 +738,38 @@ export class Datasource
...
+   * retries and meanwhile returns `false` (the safe, unoptimized path).
+   */
+  async hasTraceTimestampTable(database: string, table: string): Promise<boolean> {
+    if (!database || !table) {
+      return false;
+    }
+
+    const key = `${database}.${table}`;
+
+    let pending = this.traceTimestampTableCache.get(key);
```
adamyeats: The cache has no TTL and is only evicted on rejection, so a resolved `true` sticks until page reload. If the companion table is dropped mid-session, every "View trace" click will then ship the optimised `WITH trace_start / trace_end` shape against a table that no longer exists and fail with `Unknown table expression identifier`, which gets us back to where we started with #1842, just behind a reload gate instead of a fresh-load one. A user who hits it has no in-product signal that refreshing fixes it. 

Narrow window in practice, but worth a short TTL or a setting-change eviction hook.

src/hooks/useHasTraceTimestampTable.ts
```diff
@@ -0,0 +1,23 @@
+import { useState, useEffect } from 'react';
+import { Datasource } from 'data/CHDatasource';
+
+export default (datasource: Datasource, database: string, table: string): boolean => {
+  const [result, setResult] = useState(false);
```
adamyeats: This means the first render is always `false`, even when the datasource cache already has true resolved.

The promise resolves in a microtask but the editor's effect still fires once with false and briefly writes `meta.hasTraceTimestampTable: false` over a known-good true. The old `traceTables.length === 0` short-circuit existed to avoid exactly this thrash. Cosmetic since the response transform does its own lookup, but a sync `peekTraceTimestampTable(db, table)` on the datasource would let the hook seed `useState` with the cached value and skip the extra render entirely.

adamyeats (changes requested):
Looks good @karl-power ! 🙌 Couple of points before we merge.

src/data/CHDatasource.ts
karl-power: added a 30s TTL to the cache

src/hooks/useHasTraceTimestampTable.ts
karl-power: added the peekTraceTimestampTable method

adamyeats (approved):
Looks great! Thanks for addressing the comments, we'll cut a new release for this next week. 🚀

PR #1866, opened 2026-05-31
feat: json attributes (tags) support

Orenico10:
## Type of Change

- [x] 🚀 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧹 Refactor / Chore

---

### What is this feature?

This feature allows for using JSON instead of Map for fields like attributes with the OTel schema, and allows for queries with SpanAttributes/ResourceAttributes & more (Also works for logs and metrics) that are JSON.

When setting these columns to be JSON, the queries don't work properly, and you cannot use the full trace view, as it is using functions in the query that can only be used on Maps.

### Who is this feature for?

Anyone who is using ClickHouse with the OpenTelemetry schema with the [JSON feature enabled](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/clickhouseexporter/README.md#experimental-json-support) in the OTel Collector.

### How to test this feature

_Provide step-by-step instructions for reviewers to test the feature._

1. Use the OTel Collector to send your telemetry to your ClickHouse instance with the [JSON feature enabled](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/clickhouseexporter/README.md#experimental-json-support).
2. Connect your ClickHouse cluster to your Grafana instance.
3. Query your data like you normally would

Or use the e2e tests, which now have a `trace_spans_json` table fixture that has the full OTel schema (with json instead of maps)

---

## Screenshots / Videos

| Before | After |
| ------ | ----- |
| image       |     image  |
|    image     |   image       |

---

## Please check that:

I was not sure where I should document this. Please let me know where I should add it :)


src/data/utils.ts
```diff
@@ -223,10 +315,33 @@ export const transformQueryResponseWithTraceAndLogLinks = (
...
+      // fetchColumns so saved queries (where useOtelColumns doesn't re-run) still work.
+      let columns = originalQuery.builderOptions.columns;
+      const db = originalQuery.builderOptions.database;
+      const tbl = originalQuery.builderOptions.table;
+      // fetchedLiveSchema tracks whether we actually queried ClickHouse for column types.
+      // The meta.tagsAreJSON fallback is only used when we couldn't fetch (empty db/table).
+      const fetchedLiveSchema = !!(db && tbl);
+      try {
+        if (fetchedLiveSchema) {
+          const allCols = await getCachedColumns(db, tbl);
```
adamyeats, line 375: `getCachedColumns` fires unconditionally whenever `db` and `tbl` are set, and the cache is per-invocation and every response pays the round-trip cold. For users on `Map(String,String)` (still the majority afaik), this is pure overhead added to every trace/log query, not just the JSON-schema users this PR targets.

Could we short-circuit when the type is already known?

```
const tagsCol = options.columns?.find(
  (c) => c.hint === ColumnHint.TraceTags || c.hint === ColumnHint.TraceServiceTags
);
const typeKnown = tagsCol?.type !== undefined;
const metaSet = originalQuery.builderOptions.meta?.tagsAreJSON !== undefined;
if (!typeKnown && !metaSet && options.database && options.table) {
  // existing fetchColumns + stampJsonColumnTypes
}
```

Fresh queries get the type from `useOtelColumns`. Saved queries that have run once already carry `meta.tagsAreJSON`. Steady-state should skip the fetch. Wdyt?

src/data/sqlGenerator.ts
```diff
@@ -682,7 +698,7 @@ const getTableIdentifier = (database: string, table: string): string => {
   return `${escapeIdentifier(database)}${sep}${escapeIdentifier(table)}`;
 };
 
-const escapeIdentifier = (id: string): string => {
+export const escapeIdentifier = (id: string): string => {
   return id ? `"${id}"` : '';
```
adamyeats:
```suggestion
  return id ? `"${id.replace(/"/g, '""')}"` : '';
```

Needs a internal escape.

src/data/sqlGenerator.ts
```diff
@@ -178,24 +186,34 @@ const generateTraceIdQuery = (options: QueryBuilderOptions): string => {
...
   const flattenNested = Boolean(options.meta?.flattenNested);
+  // Events/links attributes must be Array(Map(String,String)) in the typed tuple cast so
+  // the conversion must happen in SQL. JSONExtractKeysAndValuesRaw returns all key-value
+  // pairs from a JSON string as Array(Tuple(String,String)); values are raw JSON literals
+  // (strings are quoted, numbers/booleans are not). replaceRegexpOne strips the surrounding
+  // quotes from string values so every value lands as a plain string.
+  const jsonAttrsToFields = (expr: string): string => {
+    const json = `toJSONString(${expr})`;
+    return `arrayMap(kv -> map('key', kv.1, 'value', replaceRegexpOne(kv.2, '^"(.*)"$', '\\1')), JSONExtractKeysAndValuesRaw(${json}))`;
+  };
```
adamyeats: `JSONExtractKeysAndValuesRaw` only returns top-level pairs, so `{"db.rows_examined":"42"}` (stored internally as `{"db":{"rows_examined":"42"}})` renders as `{key:"db", value:'{"rows_examined":"42"}'}` instead of the dotted-key form that top-level tags get from flattenJsonTags. Matching the e2e test's shape would fix it:

```suggestion
const jsonAttrsToFields = (expr: string): string =>
  `arrayMap(path -> map('key', path, 'value', ifNull(JSON_VALUE(toJSONString(${expr}), concat('$.', path)), '')), JSONAllPaths(${expr}))`;
```

adamyeats (changes requested):
Looks great so far @Orenico10, thanks for the submission! I've added a couple of comments. Could you also address the `cspell` errors, and add any relevent documentation, perhaps in a new section in `docs/sources/query-editor.md`?

src/data/utils.ts
Orenico10: Should be fixed :)

src/data/sqlGenerator.ts
Orenico10: Should be fixed :)

src/data/sqlGenerator.ts
Orenico10: The suggested fix does not work as it results with code 44 error from the clickhouse. I implemented a fix that should address this issue, let me know if this is better than the previous implementation :)

Orenico10:
Why are you force pushing btw?

adamyeats:
> Why are you force pushing btw?

Pulling latest from `main` to resolve merge conflicts. Merge with rebase is company convention.

adamyeats (dismissed):
Thanks @Orenico10 , this is good to merge once green! Could you please fix the lint issues? Then re-request a review from me, then we'll merge this. 🚀

src/data/sqlGenerator.ts
adamyeats: Nice, LGTM! 👍

adamyeats (approved):

PR #1868, opened 2026-06-01
feat(variables): guided variable editor with column values and Map-key support

### LINKED ISSUE #1833: Friendlier dashboard variables for authoring
Building a Grafana dashboard for a team usually starts with variables. Service name dropdown, severity dropdown, k8s.namespace dropdown, host.name dropdown, region picker, the obvious slicers consumers expect on day one. Every variable is a separate query the dashboard author has to write by hand.

For ClickHouse today that means `SELECT DISTINCT ServiceName FROM otel_logs ORDER BY 1 LIMIT 1000` for each, sometimes with a WHERE clause for the active time range or to filter empty values, sometimes with `argMax` instead of `DISTINCT`. For OTel resource attributes the author also has to know the Map / `mapValues` / bracket syntax that most people don't carry in their head. Six variables = six little SQL writing exercises before the dashboard does anything useful.

The cost is dashboard author time, repeated across every dashboard a team ships. Time spent writing variable queries is time not spent on the panels and the narrative the dashboard is actually for.

What would help during authoring: a guided variable editor with presets for the common cases.

- "Values of a column": pick database, table, column, get distinct values.
- "Values of a Map key": pick database, table, Map column, key (e.g., `service.version` in `ResourceAttributes`), get distinct values without remembering Map syntax.
- "List databases" / "List tables" / "List columns" for navigation variables.

OTel resource attributes (`service.version`, `k8s.namespace`, `k8s.deployment`) become first-class variable sources without anyone needing the Map bracket syntax. Custom SQL stays available for everything the presets don't cover, and the generated SQL is editable for the cases that are 90% there but need one tweak.

The change is additive: lower the floor for the common authoring cases, keep the ceiling where it is for advanced ones.

Related: #1828 (cascading schema picker, the foundation this editor would build on).

alex-fedotyev:
## Summary

Adds a `CustomVariableSupport` editor that replaces the legacy free-form `metricFindQuery` text input with guided variable types. Builders pick a type and the SQL is generated; the generated SQL stays editable and the runtime resolver reads `rawSql` verbatim, so manual edits always win.

### Variable types

| Type | Behaviour |
| ---- | --------- |
| Custom SQL | Free-form SQL, identical to today's `metricFindQuery` text input. |
| List databases | `SELECT name FROM system.databases ORDER BY name`. |
| List tables | `SELECT name FROM system.tables WHERE database = '<db>' ORDER BY name`. |
| List columns | `SELECT name FROM system.columns WHERE database = '<db>' AND table = '<tbl>' ORDER BY name`. |
| Column values | `SELECT DISTINCT <col> AS value FROM <db>.<tbl> WHERE <col> IS NOT NULL ORDER BY value LIMIT 1000`. When the chosen column is `Map(String, String)`, bracket access is used: `<col>['<key>']`. |

### Built on #1828

This editor consumes the reusable `SchemaPicker` primitive from #1828, now merged. This branch is rebased onto `main`, so the diff here is just the variable editor; the `SchemaPicker` files are no longer duplicated in this PR.

## Changes

- `src/data/CHVariableSupport.tsx` (new): the editor component, the `generateVariableSql` pure function, and the `CHVariableSupport` class that extends `CustomVariableSupport<Datasource, CHVariableQuery>`. The class wraps `metricFindQuery` output into a single-field `DataFrame` with a `text` column so the Grafana variable resolver can read it.
- `src/data/CHDatasource.ts`: imports `CHVariableSupport`, wires `this.variables = new CHVariableSupport(this)` in the constructor, and exports the existing `escapeCHStringLiteral` helper so the generated SQL can reuse it.
- `src/data/CHVariableSupport.test.tsx` (new): 22 cases covering the SQL generator (every variable type, identifier and string-literal escaping, Map-key escape, default-database fallback, missing-fields short-circuit), the picker-level mapping, the editor's regenerate-on-type-change and persist-user-edits behaviour, and the query method's empty-frame and DataFrame-wrapping paths.

Net diff over `main`: 3 files, +567 / -1 lines.

## Visual parity on Grafana 13.0.1+security-01

Verified side by side with the official 4.17.0 plugin against the public ClickHouse demo (`sql-clickhouse.clickhouse.com/otel_v2`). The PR build is loaded as a second datasource (`grafana-clickhouse-datasource-custom`) so both editors render in the same Grafana instance.

### Default state, type picker open

Type picker open

### Column values + Map column auto-detected + Map Key picker

Column values cascade

### Map key selected, generated SQL uses bracket access, preview resolves against real data

Preview resolves to 19 services

The preview returns 19 distinct `ResourceAttributes['service.name']` values from the OTel demo (accounting, ad, cart, checkout, ...). The variable resolution path through `metricFindQuery` is unchanged.

## Test plan

- [x] \`npm run typecheck\` is green on \`@grafana/* @ 12.4.2\` (the pinned versions on \`main\` after merging \`feat/schema-picker\`).
- [x] \`npm run lint\` is green; no new warnings introduced. The pre-existing \`Select\` deprecation warnings on the existing editors remain (broader Combobox migration, out of scope).
- [x] \`npx jest --maxWorkers=2\` passes 69 suites / 856 tests with 22 new cases in \`CHVariableSupport.test.tsx\`.
- [x] Manual verification on Grafana 13.0.1+security-01 against the public ClickHouse demo: opened a dashboard, added a query variable, selected the PR-built datasource, picked Column values + \`otel_v2.otel_logs.ResourceAttributes['service.name']\`, confirmed generated SQL and that Preview returns 19 services.

- The editor intentionally has no schema-fetch logic of its own; everything cascades through \`SchemaPicker\`, including the existing \`useDatabases\` / \`useTables\` / \`useColumns\` / \`useUniqueMapKeys\` hooks. \`SchemaPicker\` reports whether the chosen column is a \`Map(...)\` type via \`isMapColumn\` on its emitted value, so \`onPickerChange\` records that flag without any extra fetch of its own.
- \`generateVariableSql\` is pure and exported so the editor and the tests agree. The class's \`query\` method only reads \`rawSql\`, never the picker state, so any manual edits to the SQL field always win at run time.
- No \`plugin.json\`, \`CODEOWNERS\`, \`CHANGELOG.md\`, \`package.json\`, schema, or backend changes.

Please add the \`changelog\` label so this surfaces in the next release notes.

Closes #1833.

### DIFF
--- a/src/data/CHVariableSupport.tsx
+++ b/src/data/CHVariableSupport.tsx
@@ -1,0 +1,269 @@
+import React, { useCallback, useMemo } from 'react';
+import {
+  CustomVariableSupport,
+  DataQueryRequest,
+  DataQueryResponse,
+  MetricFindValue,
+  QueryEditorProps,
+  toDataFrame,
+} from '@grafana/data';
+import { InlineFormLabel, Select, TextArea } from '@grafana/ui';
+import { Observable, from, of } from 'rxjs';
+import { SchemaPicker, SchemaPickerLevel, SchemaPickerValue } from 'components/queryBuilder/SchemaPicker';
+import { CHConfig } from 'types/config';
+import { CHQuery } from 'types/sql';
+import { Datasource } from './CHDatasource';
+
+/**
+ * Variable query types. Each one renders a different combination of pickers and
+ * generates a default SQL query that the user can edit before saving.
+ */
+export type CHVariableQueryType =
+  | 'sql'
+  | 'databases'
+  | 'tables'
+  | 'columns'
+  | 'columnValues'
+  | 'otelServices'
+  | 'otelLevels'
+  | 'otelOperations';
+
+/** Variable query model. Persisted as part of the dashboard JSON. */
+export interface CHVariableQuery {
+  refId: string;
+  queryType: CHVariableQueryType;
+  rawSql?: string;
+  database?: string;
+  table?: string;
+  column?: string;
+  /** Key inside a `Map(...)` column. Only meaningful when `columnIsMap` is true. */
+  mapKey?: string;
+  /** Captured at pick time so the runtime path doesn't need to refetch the column type. */
+  columnIsMap?: boolean;
+}
+
+const VARIABLE_TYPE_OPTIONS: Array<{ label: string; value: CHVariableQueryType; description?: string }> = [
+  { label: 'Custom SQL', value: 'sql', description: 'Write any SQL query, same as before' },
+  { label: 'List databases', value: 'databases', description: 'All databases on the server' },
+  { label: 'List tables', value: 'tables', description: 'Tables inside a database' },
+  { label: 'List columns', value: 'columns', description: 'Columns inside a table' },
+  { label: 'Column values', value: 'columnValues', description: 'Distinct values of a column or Map key' },
+  { label: 'OTel services', value: 'otelServices', description: 'ServiceName values from otel_logs' },
+  { label: 'OTel log levels', value: 'otelLevels', description: 'SeverityText values from otel_logs' },
+  { label: 'OTel operations', value: 'otelOperations', description: 'SpanName values from otel_traces' },
+];
+
+/** Returns the SchemaPicker depth for a query type, or null when no picker is needed. */
+export function pickerLevelFor(queryType: CHVariableQueryType): SchemaPickerLevel | null {
+  switch (queryType) {
+    case 'tables':
+      return 'database';
+    case 'columns':
+      return 'table';
+    case 'columnValues':
+      return 'mapKey';
+    default:
+      return null;
+  }
+}
+
+/**
+ * Generate the default SQL for a variable query. Pure function so the editor
+ * preview and the unit tests use the same source of truth. At run time the
+ * variable resolver only reads `rawSql`, so any manual edits to the SQL field
+ * always win.
+ */
+export function generateVariableSql(query: CHVariableQuery, defaultDatabase: string): string {
+  const db = query.database || defaultDatabase || '';
+  const otelDb = defaultDatabase || db || 'default';
+  switch (query.queryType) {
+    case 'databases':
+      return 'SELECT name FROM system.databases ORDER BY name';
+    case 'tables':
+      return db
+        ? `SELECT name FROM system.tables WHERE database = '${db}' ORDER BY name`
+        : 'SELECT name FROM system.tables ORDER BY name';
+    case 'columns':
+      if (!db || !query.table) {
+        return '';
+      }
+      return `SELECT name FROM system.columns WHERE database = '${db}' AND table = '${query.table}' ORDER BY name`;
+    case 'columnValues': {
+      if (!db || !query.table || !query.column) {
+        return '';
+      }
+      const target =
+        query.columnIsMap && query.mapKey ? `${query.column}['${query.mapKey}']` : query.column;
+      return `SELECT DISTINCT ${target} AS value FROM ${db}.${query.table} WHERE ${target} IS NOT NULL ORDER BY value LIMIT 1000`;
+    }
+    case 'otelServices':
+      return `SELECT DISTINCT ServiceName FROM ${otelDb}.otel_logs WHERE $__timeFilter(Timestamp) AND ServiceName != '' ORDER BY ServiceName`;
+    case 'otelLevels':
+      return `SELECT DISTINCT SeverityText FROM ${otelDb}.otel_logs WHERE $__timeFilter(Timestamp) AND SeverityText != '' ORDER BY SeverityText`;
+    case 'otelOperations':
+      return `SELECT DISTINCT SpanName FROM ${otelDb}.otel_traces WHERE $__timeFilter(Timestamp) AND SpanName != '' ORDER BY SpanName LIMIT 200`;
+    case 'sql':
+    default:
+      return query.rawSql || '';
+  }
+}
+
+type EditorProps = QueryEditorProps<Datasource, CHQuery, CHConfig, CHVariableQuery>;
+
+export const VariableQueryEditor = (props: EditorProps) => {
+  const { query, onChange, datasource } = props;
+  const safeQuery: CHVariableQuery = useMemo(
+    () => ({
+      refId: query?.refId || 'var',
+      queryType: query?.queryType || 'sql',
+      rawSql: query?.rawSql,
+      database: query?.database,
+      table: query?.table,
+      column: query?.column,
+      mapKey: query?.mapKey,
+      columnIsMap: query?.columnIsMap,
+    }),
+    [
+      query?.refId,
+      query?.queryType,
+      query?.rawSql,
+      query?.database,
+      query?.table,
+      query?.column,
+      query?.mapKey,
+      query?.columnIsMap,
+    ]
+  );
+
+  const defaultDatabase = datasource.getDefaultDatabase() || '';
+  const pickerLevel = pickerLevelFor(safeQuery.queryType);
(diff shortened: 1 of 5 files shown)


src/data/CHVariableSupport.tsx
```diff
@@ -0,0 +1,269 @@
...
+  }
+}
+
+/**
+ * Generate the default SQL for a variable query. Pure function so the editor
+ * preview and the unit tests use the same source of truth. At run time the
+ * variable resolver only reads `rawSql`, so any manual edits to the SQL field
+ * always win.
+ */
+export function generateVariableSql(query: CHVariableQuery, defaultDatabase: string): string {
```
mattvella07, line 71: Since allowCustomValue is enabled on the <Select> components in SchemaPicker, users can type anything into the database, table, column, and map key pickers. These values are interpolated directly into SQL without escaping.

mattvella07: Identifiers (lines 96–97, 100, 102, 104) — db, query.table, query.column, target, and otelDb used in FROM/SELECT/WHERE positions should be wrapped with escapeIdentifier from sqlGenerator.ts.Identifiers (lines 96–97, 100, 102, 104) — db, query.table, query.column, target, and otelDb used in FROM/SELECT/WHERE positions should be wrapped with escapeIdentifier from sqlGenerator.ts.

mattvella07: String literals (lines 84, 90, 96) — db and query.table in WHERE database = '...' AND table = '...' comparisons, and query.mapKey in column['${query.mapKey}'], need single quote escaping (s.replace(/'/g, "''")).String literals (lines 84, 90, 96) — db and query.table in WHERE database = '...' AND table = '...' comparisons, and query.mapKey in column['${query.mapKey}'], need single quote escaping (s.replace(/'/g, "''")).

src/data/CHVariableSupport.tsx
```diff
@@ -0,0 +1,269 @@
...
+  );
+
+  const onPickerChange = useCallback(
+    async (picked: SchemaPickerValue) => {
+      let columnIsMap = false;
+      if (safeQuery.queryType === 'columnValues' && picked.column && picked.database && picked.table) {
+        try {
+          const columns = await datasource.fetchColumns(picked.database, picked.table);
+          const sel = columns.find((c) => c.name === picked.column);
+          columnIsMap = sel ? sel.type.startsWith('Map(') : false;
```
mattvella07: I think this is calculated on SchemaPicker.tsx line 64 so you should be able to add it to `SchemaPickerValue` and then use `isMapColumn`

src/data/CHVariableSupport.tsx
alex-fedotyev: Good catch. Fixed in 661bb39:
- Identifiers (column, db, table in the column-values query) are now wrapped with escapeIdentifier.
- String literals (db and table in the system.tables / system.columns lookups, and the Map key) go through the existing escapeCHStringLiteral helper, which escapes backslash then single quote, so a trailing backslash cannot break out either.
- Added positive and negative tests covering quote and backslash payloads.

The otelDb usage and the OTel preset lines this referenced were removed in 8429c53, so that part is moot now.

src/data/CHVariableSupport.tsx
alex-fedotyev: This touches SchemaPickerValue in SchemaPicker.tsx, which lands in #1828 (this branch consumes a copy of it). I'll fold it in once #1828 merges so the change goes in the right place rather than diverging the copy here. Tracking it.

adamyeats (changes requested):
@alex-fedotyev Looks like this might depend on some changes we made in https://github.com/grafana/clickhouse-datasource/pull/1828, so I'll wait until you rebase before going ahead with the review of this branch!

alex-fedotyev:
@adamyeats rebased onto main now that #1828 is in. The duplicate SchemaPicker commit is dropped, so the diff here is just the variable editor sitting on top of the merged SchemaPicker. Ready for review when you have a moment.

src/data/CHVariableSupport.tsx
alex-fedotyev: Good call. Done in 9849df6: SchemaPicker now carries `isMapColumn` on its emitted value (computed from the columns it already fetched for the dropdown), and `onPickerChange` reads that instead of doing its own `fetchColumns`. Now that #1828 is merged this lands in SchemaPicker directly.

src/data/CHVariableSupport.tsx
```diff
@@ -0,0 +1,251 @@
...
+  constructor(private readonly datasource: Datasource) {
+    super();
+  }
+
+  editor = VariableQueryEditor;
+
+  query(request: DataQueryRequest<CHVariableQuery>): Observable<DataQueryResponse> {
+    const target = request.targets[0];
+    if (!target?.rawSql) {
+      return of({ data: [] });
```
adamyeats, line 246: Legacy string targets will resolve to empty on this check. Needs a `typeof target === 'string'` fallback (and `safeQuery.rawSql` likewise).

src/data/CHVariableSupport.tsx
```diff
@@ -0,0 +1,251 @@
...
+    }
+    // Pass rawSql as a string. metricFindQuery accepts (CHQuery | string) and
+    // wraps a string into a minimal SQL-mode CHQuery internally; that keeps
+    // pluginVersion / refId concerns where they already live.
+    const promise = this.datasource
+      .metricFindQuery(target.rawSql, { range: request.range })
+      .then((values: MetricFindValue[]) => ({
+        data: [toDataFrame({ fields: [{ name: 'text', values: values.map((v) => v.text) }] })],
```
adamyeats: `query()` emits only a text field, dropping `v.value`, so `SELECT id, label …` now substitutes the label instead of the id. Is that intentional?

src/data/CHVariableSupport.tsx
```diff
@@ -0,0 +1,251 @@
...
+
+type EditorProps = QueryEditorProps<Datasource, CHQuery, CHConfig, CHVariableQuery>;
+
+export const VariableQueryEditor = (props: EditorProps) => {
+  const { query, onChange, datasource } = props;
+  const safeQuery: CHVariableQuery = useMemo(
+    () => ({
+      refId: query?.refId || 'var',
+      queryType: query?.queryType || 'sql',
+      rawSql: query?.rawSql,
```
adamyeats:
```suggestion
      rawSql: typeof query === 'string' ? query : query?.rawSql,
```

adamyeats (changes requested):

src/data/CHVariableSupport.tsx
alex-fedotyev: Good catch. Fixed in bc14649: `query()` resolves the SQL via `typeof target === 'string' ? target : target?.rawSql`, and the editor's `safeQuery` does the same for the string case, so a legacy string query no longer resolves to empty.

src/data/CHVariableSupport.tsx
alex-fedotyev: Not intentional, thanks for catching it. Fixed in bc14649: `query()` now emits a `value` field alongside `text` (falling back to text when the query returns a single column), so `SELECT value, label` substitutes the value while showing the label.

src/data/CHVariableSupport.tsx
alex-fedotyev: Applied in bc14649. I pulled it into a derived `legacyRawSql` so the useMemo dependency list stays exhaustive, but it is the same effect as the suggestion.

adamyeats (approved):
Nice @alex-fedotyev, thank you! 🚢

PR #1885, opened 2026-06-10
fix(adhoc): bound Map-key probe to avoid full-column scans

### LINKED ISSUE #1843: fetchUniqueMapKeys probe scans full Map column on every Filter editor interaction — high ClickHouse CPU
**What happened**:

The Visual Query Builder's Filter editor invokes `useUniqueMapKeys` whenever a user adds or edits a filter on a `Map(...)` column (e.g. `LogAttributes`, `ResourceAttributes`, `ScopeAttributes`). The hook calls `Datasource.fetchUniqueMapKeys`, which issues an unbounded query against the source table:

https://github.com/grafana/clickhouse-datasource/blob/main/src/data/CHDatasource.ts#L766-L769

```typescript
// TODO: This query can be slow/expensive
async fetchUniqueMapKeys(mapColumn: string, db: string, table: string): Promise<string[]> {
  const rawSql = `SELECT DISTINCT arrayJoin(${mapColumn}.keys) as keys FROM "${db}"."${table}" LIMIT 1000`;
  return this.fetchData(rawSql);
}
```

The `LIMIT 1000` is applied after `DISTINCT arrayJoin(…keys)`. On low-cardinality Map columns the streaming `DISTINCT` will short-circuit once it has found 1000 keys, but on high-cardinality Map columns (e.g. user-supplied keys, IDs flattened into map keys) the optimizer has no choice but to scan the entire column. There is no time filter, no row sampling, and no `WHERE` predicate to bound the worst case.

On a large OTel logs table, **a single Filter-editor interaction scanned billions of rows fanned out across multiple replicas via parallel-replicas execution** — and that's the *good* case where `DISTINCT` saturated quickly. A less-friendly Map column on the same table would not short-circuit and would scan the full column. Even the short-circuited case shows up among the top CPU consumers cluster-wide, and every editor remount or change to `database`/`table`/`mapColumn` re-fires the hook — there is no cache or debounce.

This is the same class of bug as #1461 (JSON-paths probe), which was disabled in #1468. The Map-column probe was not addressed in that fix, and the open #1793 (ad-hoc Map filter support) routes additional call sites through `fetchUniqueMapKeys`, so the blast radius will grow once it merges. The PR description acknowledges the alignment:

> Aligned with #1461 — both this and the JSON-path discovery bug share a "sample map-like keys from a large table" pattern. […] if #1461's sampling strategy changes, this will inherit it for free.

A metadata probe used to populate a UI dropdown should not be capable of full-column scans on a large logs table. The dropdown is a hint, not a source of truth — bound the work, or let operators turn it off.

Possible directions:

1. **Time-bounded probe** — inject `WHERE <timeColumn> > now() - INTERVAL 1 HOUR` when a time column is configured (logs/traces config already has this). On a partitioned-by-day table this prunes to a handful of parts.
2. **Configurable on/off** in datasource settings, mirroring #1468's approach for JSON-paths.
3. **Cache + debounce in `useUniqueMapKeys`** so editor remounts don't re-fire the probe.

I tested a sample-then-distinct variant (push the `LIMIT` into a subquery) — it does *not* help: it still visits every part on the table to populate the subquery, so wall time and CPU were identical to the original (slightly worse, in fact).

Comparison on a large partitioned-by-day OTel logs table (single-replica execution, single-part-open cost dominates):

| Variant | Wall time | CPU time | Bytes read | Parts touched |
|---|---|---|---|---|
| Original `… LIMIT 1000` | 61.0 s | 1087 s | 14.5 GiB | 22,635 |
| Time-bounded `WHERE TimestampTime > now() - INTERVAL 1 HOUR` | **3.7 s** | **6.7 s** | 99 MiB | 3,659 |

That's a ~16× wall-time and ~160× CPU reduction from the partition prune alone, and the CPU win compounds when parallel-replicas execution is in play.

(1) + (2) feels like the minimum responsible fix. 

1. Connect the plugin to any ClickHouse table with a large `Map(String, String)` column (millions of rows).
2. Open Explore → Visual Query Builder.
3. Add a filter; pick a Map column for the filter key (e.g. `LogAttributes`).
4. Inspect `system.query_log` for `SELECT DISTINCT arrayJoin(LogAttributes.keys) … LIMIT 1000` entries — note `read_rows` and `read_bytes`.

The author already flagged this with a `// TODO: This query can be slow/expensive` comment at `CHDatasource.ts:764`.

**Environment**:

- Grafana version: 11.x
- Plugin version: 4.17.0
- ClickHouse: 25.x, large OTel logs table with `LogAttributes`, `ResourceAttributes`, `ScopeAttributes` Map columns

Happy to put up a PR if a direction is acceptable.

adamyeats:
## Type of Change

- [x] 🐛 Bug Fix

---

### What is the bug?

`Datasource.fetchUniqueMapKeys` issues an unbounded probe to populate the filter-key dropdown for `Map(...)` columns:

```sql
SELECT DISTINCT arrayJoin(<col>.keys) AS keys FROM <db>.<table> LIMIT 1000
```

`LIMIT 1000` is applied *after* `DISTINCT arrayJoin(…)`. On high-cardinality Map columns (user-supplied keys, IDs flattened into map keys) the `DISTINCT` cannot short-circuit and ClickHouse scans the entire column. There is no `WHERE` predicate, no row sampling, and no time bound — so the worst case on a large OTel logs table is a multi-billion-row scan fanned out across replicas.

#1793 (just landed) added a second call site (the adhoc tag-keys fan-out), so the blast radius is growing. Three open PRs add further call sites for the same probe (#1828 SchemaPicker, #1868 guided variable editor, #1841 compact filter popover) — all flow through the same `fetchUniqueMapKeys`, so a single-point fix benefits everything.

Reporter's measured win from a 1-hour time predicate on a large OTel logs table:

| Variant | Wall time | CPU time | Bytes read | Parts touched |
|---|---|---|---|---|
| Original \`… LIMIT 1000\` | 61.0 s | 1087 s | 14.5 GiB | 22,635 |
| Time-bounded | **3.7 s** | **6.7 s** | 99 MiB | 3,659 |

### How to reproduce

1. Connect the plugin to a ClickHouse table with a large `Map(String, String)` column (millions of rows).
2. Open Explore → Visual Query Builder.
3. Add a filter; pick a Map column as the filter key (e.g. `LogAttributes`).
4. Inspect `system.query_log` for `SELECT DISTINCT arrayJoin(LogAttributes.keys) … LIMIT 1000` and note `read_rows` / `read_bytes`.

### Related Issues

Refs #1843

(Leaving `Closes` off — the reporter proposed three directions and this PR ships the first two; an in-memory cache for editor remount churn is a reasonable follow-up.)

---

## What this changes

Two changes flow through the shared `fetchUniqueMapKeys`, so every current and upcoming call site inherits them automatically:

### 1. Time-bounded probe (when we know the time column)

When the probe target `(db, table)` matches the configured **OTel logs** or **OTel traces** table, inject `WHERE <timeColumn> >= now() - INTERVAL 6 HOUR`. The time column is sourced from `getDefaultLogsColumns()` / `getDefaultTraceColumns()` (already populated either from OTel canon or from the user-configured `logs.timeColumn` / `traces.startTimeColumn`).

On a partitioned-by-day MergeTree this prunes to a handful of parts. For **free-form tables** (no known time column), the bare-LIMIT shape is preserved — the plugin can't safely guess which column is the time column for arbitrary tables.

### 2. Opt-out setting

New `enableMapKeysDiscovery` flag in `CHConfig`, **default `true`** (preserves existing UX). When `false`:

- `fetchUniqueMapKeys` returns `[]` immediately, without issuing a query.
- The `getTagKeys()` Map-column fan-out short-circuits before issuing any probes.

Surfaced in both the v1 and v2 config UIs ([QuerySettingsConfig](src/components/configEditor/QuerySettingsConfig.tsx), [AdditionalSettingsSection](src/views/config-v2/AdditionalSettingsSection.tsx)) next to "Validate SQL".

### Why not also fix #1843 with caching?

Reporter's option #3 (cache + debounce in `useUniqueMapKeys`) is genuinely useful for editor remount churn but doesn't change the dashboard-load fan-out, which is where the largest CPU win lives. Kept out of this PR to keep the diff surgical; happy to open a follow-up.

### Why a 6-hour window and not configurable?

The 1-hour window the reporter measured already gave ~16× wall-time / ~160× CPU reductions; 6 hours is generous enough to catch keys on hourly/daily cadences while still partition-pruning aggressively. Made non-configurable for now — happy to expose it later if needed. YAGNI.

---

## Scope notes

- **Time predicate is gated to OTel logs/traces targets only.** For non-OTel/free-form tables the plugin emits the original bare-LIMIT shape — we don't risk picking the wrong column.
- **Default is unchanged behavior.** `enableMapKeysDiscovery` defaults to `true`; users only opt out if their workload is hitting the slow path.
- **Single-point fix.** Three open PRs add further call sites (#1828 SchemaPicker, #1868 guided variable editor, #1841 compact filter popover) — all consume `fetchUniqueMapKeys`, so they inherit both behaviors without needing to coordinate.

---

## Please check that:

### Tests

- Six new cases in [CHDatasource.test.ts](src/data/CHDatasource.test.ts) under `describe('fetchUniqueMapKeys probe (#1843)')`:
  - bare LIMIT for free-form tables
  - logs time-column injected when target matches logs config
  - traces start-time column injected when target matches traces config
  - no predicate when target doesn't match either
  - opt-out short-circuits to `[]` without querying
  - opt-out short-circuits the `getTagKeys` fan-out
- Updated [QuerySettingsConfig.test.tsx](src/components/configEditor/QuerySettingsConfig.test.tsx) to cover the new switch.

Net: 737/742 jest tests green (5 pre-existing skips), `npm run typecheck` and `npm run lint` clean, no new cspell hits.

- The time-column lookup uses `ColumnHint.FilterTime || ColumnHint.Time` for logs (matching how the rest of the logs pipeline picks a time column) and `ColumnHint.Time` for traces (which maps to `startTimeColumn`).
- The `getTagKeys` short-circuit is intentionally placed *before* the `Promise.all` fan-out, not inside `fetchUniqueMapKeys` alone — that way disabling the setting saves N round-trips on dashboard load, not just N×(one cheap empty fetch).
- I considered adding a per-instance Promise cache for `fetchUniqueMapKeys` (dedupe concurrent calls + short TTL for editor remount churn) but kept it out to minimize diff risk. Filing a follow-up issue.

### DIFF
--- a/src/components/configEditor/QuerySettingsConfig.test.tsx
+++ b/src/components/configEditor/QuerySettingsConfig.test.tsx
@@ -19,6 +19,7 @@ describe('QuerySettingsConfig', () => {
         onDialTimeoutChange={() => {}}
         onQueryTimeoutChange={() => {}}
         onValidateSqlChange={() => {}}
+        onEnableMapKeysDiscoveryChange={() => {}}
       />
     );
     expect(result.container.firstChild).not.toBeNull();
@@ -34,6 +35,7 @@ describe('QuerySettingsConfig', () => {
         onDialTimeoutChange={onDialTimeout}
         onQueryTimeoutChange={() => {}}
         onValidateSqlChange={() => {}}
+        onEnableMapKeysDiscoveryChange={() => {}}
       />
     );
     expect(result.container.firstChild).not.toBeNull();
@@ -56,6 +58,7 @@ describe('QuerySettingsConfig', () => {
         onDialTimeoutChange={() => {}}
         onQueryTimeoutChange={onQueryTimeout}
         onValidateSqlChange={() => {}}
+        onEnableMapKeysDiscoveryChange={() => {}}
       />
     );
     expect(result.container.firstChild).not.toBeNull();
@@ -78,15 +81,35 @@ describe('QuerySettingsConfig', () => {
         onDialTimeoutChange={() => {}}
         onQueryTimeoutChange={() => {}}
         onValidateSqlChange={onValidateSqlChange}
+        onEnableMapKeysDiscoveryChange={() => {}}
       />
     );
     expect(result.container.firstChild).not.toBeNull();
 
-    const input = result.getByRole('checkbox');
-    expect(input).toBeInTheDocument();
-    fireEvent.click(input);
+    // Two switches: validateSql first, enableMapKeysDiscovery second.
+    const inputs = result.getAllByRole('checkbox');
+    fireEvent.click(inputs[0]);
     expect(onValidateSqlChange).toHaveBeenCalledTimes(1);
     expect(onValidateSqlChange).toHaveBeenCalledWith(expect.any(Object));
+  });
+
+  it('should call onEnableMapKeysDiscoveryChange when changed', () => {
+    const onEnableMapKeysDiscoveryChange = jest.fn();
+    const result = render(
+      <QuerySettingsConfig
+        onConnMaxIdleConnsChange={() => {}}
+        onConnMaxLifetimeChange={() => {}}
+        onConnMaxOpenConnsChange={() => {}}
+        onDialTimeoutChange={() => {}}
+        onQueryTimeoutChange={() => {}}
+        onValidateSqlChange={() => {}}
+        onEnableMapKeysDiscoveryChange={onEnableMapKeysDiscoveryChange}
+      />
+    );
+    const inputs = result.getAllByRole('checkbox');
+    fireEvent.click(inputs[1]);
+    expect(onEnableMapKeysDiscoveryChange).toHaveBeenCalledTimes(1);
+    expect(onEnableMapKeysDiscoveryChange).toHaveBeenCalledWith(expect.any(Object));
   });
 
   it('should call onConnMaxIdleConnsChange when changed', () => {
@@ -99,6 +122,7 @@ describe('QuerySettingsConfig', () => {
         onDialTimeoutChange={() => {}}
         onQueryTimeoutChange={() => {}}
         onValidateSqlChange={() => {}}
+        onEnableMapKeysDiscoveryChange={() => {}}
       />
     );
     expect(result.container.firstChild).not.toBeNull();
@@ -121,6 +145,7 @@ describe('QuerySettingsConfig', () => {
         onDialTimeoutChange={() => {}}
         onQueryTimeoutChange={() => {}}
         onValidateSqlChange={() => {}}
+        onEnableMapKeysDiscoveryChange={() => {}}
       />
     );
     expect(result.container.firstChild).not.toBeNull();
@@ -145,6 +170,7 @@ describe('QuerySettingsConfig', () => {
         onDialTimeoutChange={() => {}}
         onQueryTimeoutChange={() => {}}
         onValidateSqlChange={() => {}}
+        onEnableMapKeysDiscoveryChange={() => {}}
       />
     );
     expect(result.container.firstChild).not.toBeNull();
--- a/src/components/configEditor/QuerySettingsConfig.tsx
+++ b/src/components/configEditor/QuerySettingsConfig.tsx
@@ -10,12 +10,14 @@ interface QuerySettingsConfigProps {
   maxOpenConns?: string;
   queryTimeout?: string;
   validateSql?: boolean;
+  enableMapKeysDiscovery?: boolean;
   onConnMaxIdleConnsChange: (e: FormEvent<HTMLInputElement>) => void;
   onConnMaxLifetimeChange: (e: FormEvent<HTMLInputElement>) => void;
   onConnMaxOpenConnsChange: (e: FormEvent<HTMLInputElement>) => void;
   onDialTimeoutChange: (e: FormEvent<HTMLInputElement>) => void;
   onQueryTimeoutChange: (e: FormEvent<HTMLInputElement>) => void;
   onValidateSqlChange: (e: FormEvent<HTMLInputElement>) => void;
+  onEnableMapKeysDiscoveryChange: (e: FormEvent<HTMLInputElement>) => void;
 }
 
 export const QuerySettingsConfig = (props: QuerySettingsConfigProps) => {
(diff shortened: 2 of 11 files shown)


alyssajoyner (changes requested):
Nice work 💯 A few questions/nits:

1. Could you add a test with `otelEnabled: true`? All the new tests use the manual config   path so the OTel column map path in getDefaultLogsColumns() is untested.  
2. Since `enableMapKeysDiscovery` was added to the jsonData - we could add this to the tracking event `trackClickhouseConfigV2QuerySettings` sent to rudderstack - wdyt?
3. `isMapKeysDiscoveryEnabled()` is only called within CHDatasource — should it be private?

adamyeats:
Thanks @alyssajoyner — all three addressed in 6b9997e:

1. **OTel test** — added [`bounds the probe via the OTel canon column map when otelEnabled is true`](https://github.com/grafana/clickhouse-datasource/blob/fix/map-keys-probe-1843/src/data/CHDatasource.test.ts). Sets `otelEnabled: true, otelVersion: '1.29.0'`, leaves manual `timeColumn` unset, and verifies the predicate uses `TimestampTime` (the `FilterTime` hint from the OTel 1.2.9 column map) — which would silently degrade to `Timestamp` or no predicate at all without that path covered.
2. **Tracking** — `enableMapKeysDiscovery?: boolean` added to both `trackClickhouseConfigV1QuerySettings` and `trackClickhouseConfigV2QuerySettings`. Both editor switches now emit the event alongside the existing `validateSql`/timeout interactions.
3. **`isMapKeysDiscoveryEnabled()` private** — done. Only used inside `CHDatasource` (`fetchUniqueMapKeys` + the `getTagKeys` short-circuit), no external callers.

762 jest tests still green, typecheck clean.

alyssajoyner (dismissed):
Amazing ❤️ LGTM

alyssajoyner (approved):

PR #1899, opened 2026-06-16
fix(otel): support collector-contrib clickhouseexporter v0.151.0 logs schema

### LINKED ISSUE #1882: Built-in OTel logs/traces queries break against schema produced by opentelemetry-collector-contrib clickhouseexporter ≥ v0.151.0
**What happened**:

The built-in OTel queries shipped by `grafana-clickhouse-datasource` (tested on v1.9.2, `otelVersion: latest`) fail against tables created by recent versions of the OpenTelemetry Collector Contrib `clickhouseexporter`. The exporter changed its default schema starting in **v0.151.0**, and the plugin's queries still reference columns/tables from the previous schema.

Would expect no errors.

1. Run `otel/opentelemetry-collector-contrib:0.152.1` (or any release ≥ 0.151.0) with the `clickhouse` exporter and `create_schema: true` against a fresh ClickHouse cluster.
2. Configure a Grafana data source with `grafana-clickhouse-datasource` v1.9.2, OTel integration enabled, `otelVersion: latest` (1.9.2).
3. Open Explore and run the default logs query.

The plugin-generated SQL references a `TimestampTime` column that does not exist in the table created by the current exporter — the actual column is `Timestamp`.

The OpenTelemetry Collector Contrib `clickhouseexporter` changed its default schema starting in [v0.151.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/cmd/codecovgen/v0.151.0). The relevant change is in commit [f7ece95](https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/f7ece9549f5241f99cccac685ab347679ed90147) (see also the discussion in [open-telemetry/opentelemetry-collector-contrib#48770](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48770)). Table and column names in the default logs/traces schema were changed, so the plugin's hardcoded queries no longer match what the exporter produces.

Since the exporter is the source of truth for the data, the plugin needs to be updated to match the new schema.

**Environment**:

- `grafana-clickhouse-datasource`: v1.9.2 (otelVersion: latest / 1.9.2)
- `opentelemetry-collector-contrib`: v0.152.1
- ClickHouse: created via clickhouse-operator 0.27.0
- Kubernetes 1.35.3

adamyeats:
## Type of Change

- [x] 🐛 Bug Fix

---

### What is the bug?

`opentelemetry-collector-contrib` clickhouseexporter [v0.151.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.151.0) ([PR #47720](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47720)) rewrote the default `otel_logs` schema and **removed the `TimestampTime` column**. The plugin's `1.2.9` OTel column map mapped `ColumnHint.FilterTime` → `TimestampTime`, so the default logs query (and any time-range filter / ORDER BY built from it) fails against tables created by the newer exporter with `Unknown identifier 'TimestampTime'`.

The schema diff in scope (only `otel_logs` changed; traces and metrics are untouched):

| | v0.150.x and earlier | v0.151.0+ |
| --- | --- | --- |
| `TimestampTime DateTime` column | present | **removed** |
| `PARTITION BY` | `toDate(TimestampTime)` | `toDate(Timestamp)` |
| `ORDER BY` | `(ServiceName, TimestampTime, Timestamp)` | `(toStartOfFiveMinutes(Timestamp), ServiceName, Timestamp)` |

### How to reproduce

1. Run `otel/opentelemetry-collector-contrib:0.152.1` (or any release ≥ 0.151.0) with the `clickhouse` exporter and `create_schema: true` against a fresh ClickHouse cluster.
2. Configure a Grafana data source with `grafana-clickhouse-datasource` v1.9.2, OTel integration enabled, `otelVersion: latest` (1.9.2).
3. Open Explore and run the default logs query — fails with `Unknown identifier 'TimestampTime'`.

### Fix

Add a `1.3.0` OTel schema entry that omits the `FilterTime` mapping. `sqlGenerator`'s existing fallback (`getFilters()` and `getOrderBy()` both fall back to `ColumnHint.Time` when `FilterTime` doesn't resolve) emits `Timestamp` everywhere `FilterTime` was referenced. `latest` now points at this entry; the old `1.2.9` entry is retained for users on pre-0.151 schemas.

The change is non-destructive: existing data sources keep their persisted `otelVersion` until manually updated, and `latest`-pinned data sources roll forward automatically.

### Related Issues

Closes #1882

---

## Please check that:

### Tests

- `src/otel.test.ts` (new) — locks in the new column map shape, the latest alias, and that the trace schema is unchanged.
- `src/data/sqlGenerator.test.ts` — adds two regression cases: one verifying the FilterTime → Time fallback emits `Timestamp` (and NOT `TimestampTime`) for the 1.3.0 schema, plus a control case locking in the pre-0.151 ORDER BY shape so re-pointing `latest` doesn't silently regress 1.2.9 users.
- `tests/e2e/otelLogsV151.spec.ts` + `tests/e2e/fixtures/otel_logs_v0151.sql` (new) — seeds a small `otel_logs`-shaped table matching the v0.151.0 layout, runs the SQL the plugin generates against both the 1.3.0 and 1.2.9 schemas, and asserts the 1.3.0 SQL succeeds (returns rows) while the 1.2.9 SQL fails with a `TimestampTime` error.

### Documentation

- `docs/sources/configure.md` — adds an "OTel schema versions" section explaining which exporter releases each entry tracks and the practical impact (Filter Time column left blank under 1.3.0). The Logs settings table now points to it.
- `CHANGELOG.md` — Unreleased / Features entry.

A couple of follow-ups intentionally out of scope:

- **Schema auto-detect** — probing `system.columns` to pick the right column map automatically would remove the version-selection step entirely. Filed separately.
- **Version-field naming cleanup** — the persisted `otelVersion` (`1.29.0`, `1.30.0`) uses OTel semconv numbering as a proxy for "exporter schema generation"; they aren't formally linked. A future refactor could switch to exporter-version IDs with a migration. Kept the existing convention here to avoid a config migration in a bug-fix PR.

### DIFF
--- a/src/data/sqlGenerator.test.ts
+++ b/src/data/sqlGenerator.test.ts
@@ -129,6 +129,89 @@ describe('SQL Generator', () => {
     expect(sql).toEqual(expectedSqlParts.join(' '));
   });
 
+  // Regression test for #1882: opentelemetry-collector-contrib clickhouseexporter v0.151.0
+  // dropped the TimestampTime column from otel_logs. The default WHERE filter and ORDER BY
+  // entries the logs query builder installs reference ColumnHint.FilterTime; when that hint
+  // isn't mapped (the 1.3.0 OTel schema), the generator must fall back to ColumnHint.Time and
+  // emit `Timestamp`, not `TimestampTime`.
+  it('falls back to Time column when FilterTime is unmapped (collector-contrib v0.151.0+ schema)', () => {
+    const opts: QueryBuilderOptions = {
+      database: 'otel',
+      table: 'otel_logs',
+      queryType: QueryType.Logs,
+      columns: [
+        { name: 'Timestamp', type: 'DateTime64(9)', hint: ColumnHint.Time },
+        { name: 'SeverityText', type: 'String', hint: ColumnHint.LogLevel },
+        { name: 'Body', type: 'String', hint: ColumnHint.LogMessage },
+      ],
+      limit: 1000,
+      filters: [
+        {
+          filterType: 'custom',
+          type: 'datetime',
+          key: '',
+          condition: 'AND',
+          hint: ColumnHint.FilterTime,
+          operator: FilterOperator.WithInGrafanaTimeRange,
+        },
+      ],
+      orderBy: [
+        { name: '', hint: ColumnHint.FilterTime, dir: OrderByDirection.DESC, default: true },
+        { name: '', hint: ColumnHint.Time, dir: OrderByDirection.DESC, default: true },
+      ],
+    };
+
+    const expectedSqlParts = [
+      'SELECT Timestamp as "timestamp", Body as "body", SeverityText as "level"',
+      'FROM "otel"."otel_logs"',
+      'WHERE ( timestamp >= $__fromTime AND timestamp <= $__toTime )',
+      'ORDER BY timestamp DESC LIMIT 1000',
+    ];
+
+    const sql = generateSql(opts);
+    expect(sql).toEqual(expectedSqlParts.join(' '));
+    expect(sql).not.toContain('TimestampTime');
+  });
+
+  // Companion to the v0.151.0 test above: when both FilterTime and Time are mapped (the older
+  // 1.2.9 schema where otel_logs has both `TimestampTime` and `Timestamp` columns), the grouped
+  // ORDER BY must still emit `(TimestampTime, Timestamp) DESC` and the WHERE filter must use
+  // `TimestampTime`. Locks in the pre-0.151 behavior so re-pointing `latest` to 1.3.0 doesn't
+  // silently regress users on the older schema.
+  it('emits TimestampTime when FilterTime is mapped (collector-contrib v0.150.x and earlier schema)', () => {
+    const opts: QueryBuilderOptions = {
+      database: 'otel',
+      table: 'otel_logs',
+      queryType: QueryType.Logs,
+      columns: [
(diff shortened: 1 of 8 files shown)


src/data/sqlGenerator.test.ts
```diff
@@ -129,6 +129,89 @@ describe('SQL Generator', () => {
...
+      ],
+      orderBy: [
+        { name: '', hint: ColumnHint.FilterTime, dir: OrderByDirection.DESC, default: true },
+        { name: '', hint: ColumnHint.Time, dir: OrderByDirection.DESC, default: true },
+      ],
+    };
+
+    const sql = generateSql(opts);
+    // The Time column is aliased to "timestamp" in the SELECT and the alias is reused in ORDER BY.
+    expect(sql).toContain('WHERE ( TimestampTime >=');
```
iwysiu: Nit: Is there a reason this one isn't just the full query (which would seem to be a stricter check?)

iwysiu (dismissed):

src/data/sqlGenerator.test.ts
adamyeats: Good call — no real reason for the asymmetry. Switched it to a full-query `toEqual` matching the 1.3.0 companion test (49ef993). That also picks up coverage the substring checks skipped, e.g. the upper `<= $__toTime` bound and the `LIMIT`.

dgiagio (approved):

PR #1919, opened 2026-06-17
fix(traces): preserve trace ID link optimization on cold cache

### LINKED ISSUE #1918: View trace by TraceID doesn't leverage the materialized view on the first query
**What happened**:

When clicking on a trace deeplink from `Trace Search`, the first query doesn't leverage the materialized view and eventually timeouts.
We have to cancel the query and then refresh so it uses the correct query this time.

When clicking on a trace deeplink from Trace Search, the first query should leverage the materialized view right away.

- Click on a trace deeplink from `Trace Search`
Image

- A new tab will be opened to search by TraceID. 
Image

- This first query will eventually timeout because it won't leverage the "trace_id to timestamp" materialized view
Image

- When clicking on "refresh", the query now uses the materialized view (see `otel_traces_trace_id_ts`) and it doesn't timeout.
Image

The bug is NOT present on `v4.13`.
The bug is present on both `v4.17` and `v4.18`

**Environment**:

- Grafana version: `13.0.1`
- Plugin version: `4.17` and `4.18`
- OS Grafana is installed on: EKS
- User OS & Browser: Chrome

adamyeats:
## Type of Change

- [x] 🐛 Bug Fix

---

### What is the bug?

The "View trace" data link generated by Trace Search ships with optimized `rawSql` (a `WITH … _trace_id_ts` clause that narrows the trace span scan to the time window for the trace ID) and `meta.hasTraceTimestampTable: true` baked in by `transformQueryResponseWithTraceAndLogLinks` ([src/data/utils.ts:397](https://github.com/grafana/clickhouse-datasource/blob/main/src/data/utils.ts#L397)). On a fresh page load — i.e. the deep-link opens in a new tab and the datasource's trace-timestamp-table cache is cold — three things race:

1. `useHasTraceTimestampTable` returns `false` on its initial render because `peekTraceTimestampTable()` is `undefined`.
2. The editor's dispatch effect mirrors that `false` into `builderOptions.meta.hasTraceTimestampTable`.
3. The next `builderOptions` effect calls `generateSql(builderOptions)`, which now emits the plain-`WHERE` fallback (no `_trace_id_ts` join), and onChange replaces the optimized query in Grafana's state.

The end result: the very first query Grafana runs after the link opens is the unoptimized one and times out on real traces. Hitting **refresh** regenerates SQL from the (by now resolved) `true` value, which is why a manual refresh "fixes" it.

### How to reproduce

1. Configure a ClickHouse datasource with OTel tracing and a real `<table>_trace_id_ts` companion table.
2. From a dashboard panel (so the link uses `targetBlank: true` and opens in a fresh tab with a cold datasource cache), run a Trace Search query and click **View trace** on a result row.
3. Observe the generated SQL: `WHERE traceID = '<id>'` with no `_trace_id_ts` lookup. The query eventually times out.
4. Click **refresh** — the SQL is regenerated with the `WITH trace_start/trace_end` shape and returns immediately.

### Related Issues

Closes #1918
Relates to #1705, #1786, #1853

---

## Please check that:

**Files changed**

`src/hooks/useHasTraceTimestampTable.ts` — Return type widens to `boolean | undefined`. The `useState` initializer drops the `?? false` fallback so the value stays `undefined` while the async lookup is in flight on a cold cache. Resolved values (true/false) and the empty-db/empty-table guard still set a concrete boolean.

`src/views/CHQueryEditor.tsx` — The dispatch effect that mirrors the hook into `meta.hasTraceTimestampTable` now bails out when the hook is `undefined`. This preserves the authoritative value that `transformQueryResponseWithTraceAndLogLinks` already baked into the link query, so the editor never regenerates `rawSql` with the unoptimized shape between mount and the hook's async resolution.

**Tests**

`src/hooks/useHasTraceTimestampTable.test.ts` — The cold-cache test is updated to assert the initial render is `undefined` (the regression guard for the cause), then resolves to the live value. A new case pins that a definitive `false` from the async check still propagates so the editor can still correct stale optimistic meta.

`src/views/CHQueryEditor.test.tsx` — New regression test mounts `CHQueryEditor` with a pre-baked trace ID link query against a cold-cache datasource (`peekTraceTimestampTable` returns `undefined`, `hasTraceTimestampTable` resolves `true`) and asserts that no `onChange` call ever ships `meta.hasTraceTimestampTable: false` or a `rawSql` missing the `_trace_id_ts` join. Verified that the test fails on the pre-fix code (the intermediate `onChange` carries `hasTraceTimestampTable: false` and the unoptimized SQL) and passes after.

### DIFF
--- a/src/hooks/useHasTraceTimestampTable.test.ts
+++ b/src/hooks/useHasTraceTimestampTable.test.ts
@@ -20,16 +20,35 @@ describe('useHasTraceTimestampTable', () => {
     expect(result.current).toBe(true);
   });
 
-  it('starts false on a cold cache then updates to the resolved value', async () => {
+  it('stays undefined on a cold cache then updates to the resolved value', async () => {
+    // Returning `undefined` (rather than a transient `false`) is what lets the
+    // CHQueryEditor effect skip dispatching while the async lookup is in
+    // flight, preserving any pre-set `meta.hasTraceTimestampTable: true` baked
+    // into trace ID deep-link queries by the response transform (#1918).
     const mockDs = {} as Datasource;
     mockDs.peekTraceTimestampTable = jest.fn(() => undefined);
     mockDs.hasTraceTimestampTable = jest.fn(() => Promise.resolve(true));
 
     const { result } = renderHook(() => useHasTraceTimestampTable(mockDs, 'otel', 'otel_traces'));
-    expect(result.current).toBe(false); // cold cache → initial render is false
+    expect(result.current).toBeUndefined(); // cold cache → "don't know yet"
 
     await act(async () => {}); // flush the async check
     expect(result.current).toBe(true); // updated once it resolves
+  });
+
+  it('resolves to false when the companion table does not exist', async () => {
+    // A definitive false from the async check must propagate so the editor can
+    // correct stale optimistic meta (e.g. a saved query from a table that has
+    // since lost its companion).
+    const mockDs = {} as Datasource;
+    mockDs.peekTraceTimestampTable = jest.fn(() => undefined);
+    mockDs.hasTraceTimestampTable = jest.fn(() => Promise.resolve(false));
+
+    const { result } = renderHook(() => useHasTraceTimestampTable(mockDs, 'otel', 'otel_traces'));
+    expect(result.current).toBeUndefined();
+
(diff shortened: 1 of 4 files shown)


zoltanbedi (approved):
LGTM

PR #1920, opened 2026-06-19
fix(traces): prevent crash on raw-SQL queries with a trace_id column

karl-power:
## Type of Change

- [ ] 🚀 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧹 Refactor / Chore

### What is the bug?

Any raw-SQL query (Explore mode, editorType: "sql") that returns a column named trace_id or traceid crashes with "Cannot read properties of undefined (reading 'meta')". Backend response is HTTP 200 with valid data but the frontend renders nothing.

### How to reproduce

1. Make sure `builderOptions` is really undefined: don't touch anything in query builder, don't configure a default table for traces.
2. Perform a raw sql query as above.
3. Observe error in console and empty response.

## Please check that:

### DIFF
--- a/src/data/utils.test.ts
+++ b/src/data/utils.test.ts
@@ -3,6 +3,7 @@ import {
   applyTraceSearchFieldConfig,
   columnLabelToPlaceholder,
   dataFrameHasLogLabelWithName,
+  getBuilderOptions,
   isBuilderOptionsRunnable,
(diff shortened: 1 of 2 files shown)


adamyeats (approved):
Great, thanks @karl-power ! 🚢

PR #1922, opened 2026-06-23
feat(annotations): preset-driven annotation editor with change detection

### LINKED ISSUE #1834: Friendlier dashboard annotations for authoring
A dashboard author building an OTel + ClickHouse dashboard usually wants two kinds of annotations on the panels:

- Mark version deployments when a version changes (a resource attribute like `service.version`, or a `deployment.environment` value rolling forward). The use case includes rollbacks: v1 → v2 → v1 should produce three annotations, not one.
- Mark deployments and rollouts from K8s events tables (the rows the K8s operator writes with `reason = Deployment` / `Rollout`).

Both should be a few clicks during dashboard authoring. Today they're not. Annotations against a ClickHouse table require writing the full SELECT yourself, knowing the exact alias names Grafana expects (`time`, `timeEnd`, `title`, `tags`, `text`), writing the time-filter clause manually so annotations honor the dashboard's time range, and reading opaque SQL failures when one of those steps is wrong.

What would help: annotation presets that cover the common cases.

- "Change detection" preset: pick a database, table, and column (or Map key, for OTel resource attributes). The annotation fires every time the watched value changes, rollback-aware.
- "K8s lifecycle events" preset: pick the K8s events table, get deployment / rollout annotations with sensible defaults for the title / tags / text columns.

Marking deployments on an OTel-or-K8s dashboard becomes a dropdown + a preview, not a SELECT-and-five-aliases ritual. Custom SQL stays for the cases the presets don't cover; generated SQL stays editable.

alex-fedotyev:
## Summary

The plugin ships no custom annotation editor today (`CHDatasource` sets `annotations = {}`), so authoring a dashboard annotation means hand-writing ClickHouse SQL with no guidance. This adds a preset-driven annotation editor with two options:

- **Custom SQL**: a plain SQL field (the previous behavior, now explicit).
- **Change Detection**: detect when a column value changes over time, such as a deployment (`service.version`), a config value, or a feature flag. Built on the cascading schema picker, it generates rollback-aware SQL with `lagInFrame()` over 30-second buckets, so a `v1.0` to `v1.1` back to `v1.0` rollback produces three annotations rather than two.

The generated SQL is always visible and editable, so you can start from the preset and refine it, or write a query from scratch. Legacy annotations stored as a top-level `rawQuery` string are migrated into the modern target shape on load.

Fixes #1834

Builds on the cascading schema picker from #1828, now merged. The branch is rebased onto main, so the diff here is annotation-only.

## Changes

- Add `src/data/CHAnnotationSupport.tsx`: preset types, the change-detection SQL generator, the editor component, and a `createAnnotationSupport(datasource)` factory returning Grafana's `AnnotationSupport`.
- Wire it into the datasource: `this.annotations = createAnnotationSupport(this)` in the `CHDatasource` constructor, mirroring how `this.variables` is set up.
- Generated identifiers go through `escapeIdentifier`; the watched map key is emitted as an escaped single-quoted string literal.
- Resolve the default database and table from the datasource's traces config, then the single-table config, then the general default with the conventional `otel_traces` name, so the editor adapts to both all-databases and single-table datasources.
- Document the editor in `docs/sources/annotations.md`.
- Add unit tests (`src/data/CHAnnotationSupport.test.tsx`) and a Playwright spec (`tests/e2e/annotations.spec.ts`).

## Why

Change detection on a deployment attribute (`ResourceAttributes['service.version']`) is the common case, and writing the windowed `lagInFrame()` query by hand is easy to get wrong. The builder produces correct rollback-aware SQL and keeps it editable for anything bespoke. The change-detection default matches the deployment annotation already embedded in the OOTB dashboards, so the two stay consistent.

## Local verification

Built the plugin and ran it in Grafana 13 against the public ClickHouse OTel demo, side by side with the official build. Walked the annotation settings: switching between Custom SQL and Change Detection, building a change-detection query through the schema picker (Database, Table, Watch Column, Map Key, Group By), and confirming the generated SQL runs against ClickHouse with no error. Verified in light and dark themes.

| | Custom SQL | Change Detection |
|---|---|---|
| **Light** | Custom SQL preset, light theme | Change Detection builder, light theme |
| **Dark** | Custom SQL preset, dark theme | Change Detection builder, dark theme |

The Change Detection builder above generated and ran this query (watching `ResourceAttributes['service.version']` grouped by `ServiceName`):

```sql
SELECT
  time,
  "ServiceName" AS tags,
  concat("ServiceName", ': ResourceAttributes.service.version changed to ', version,
    if(prev_version != '', concat(' (was ', prev_version, ')'), '')) AS text,
  'ResourceAttributes.service.version change' AS title
FROM (
  SELECT
    toStartOfInterval(Timestamp, INTERVAL 30 second) AS time,
    "ServiceName",
    topK(1)("ResourceAttributes"['service.version'])[1] AS version,
    lagInFrame(topK(1)("ResourceAttributes"['service.version'])[1])
      OVER (PARTITION BY "ServiceName" ORDER BY time) AS prev_version
  FROM "otel_v2"."otel_traces"
  WHERE $__timeFilter(Timestamp)
    AND "ResourceAttributes"['service.version'] != ''
  GROUP BY "ServiceName", time
  ORDER BY "ServiceName", time
)
WHERE prev_version != '' AND prev_version != version
ORDER BY time
```

Two details in that query matter against real OTel data, both verified on the public demo: the per-bucket value uses `topK(1)` (the dominant value), not `any()`, so a service running two versions at once during a rollout does not flap between them; and the `prev_version != ''` guard drops each group's first bucket, which `lagInFrame` would otherwise report as a change from an empty previous value. A service whose version is stable across the range produces no markers.

A logs-based "lifecycle events" preset (Kubernetes restarts, OOM kills, scaling) is intentionally left for a follow-up: getting readable annotations out of real OTel k8s events needs schema-aware extraction (the event detail is nested in the log body and varies by collector), which deserves its own focused PR.

[ui-states: allow] annotation editing is a form builder; empty / loading / error are states of Grafana's annotation results panel, not this editor.
[viewport: allow] the editor renders inside Grafana's dashboard settings form, where layout width is managed by Grafana.

## Test plan

- [x] `npx jest src/data/CHAnnotationSupport.test.tsx` (21 tests: change-detection SQL generation, map-key escaping, preset switching, legacy `rawQuery` migration, and default-schema resolution for all-databases / single-table / bare datasources)
- [x] `npx jest` (full suite green, no regressions from the datasource wiring)
- [x] `npx tsc --noEmit`
- [x] `npx eslint` (clean) and `npm run spellcheck`
- [x] `tests/e2e/annotations.spec.ts` (Playwright via `@grafana/plugin-e2e`: the change-detection builder renders the schema picker and generates SQL, and a custom-SQL query runs against ClickHouse)
- [x] Manual verification in local Grafana 13, light and dark, screenshots above

### DIFF
--- a/src/data/CHAnnotationSupport.tsx
+++ b/src/data/CHAnnotationSupport.tsx
@@ -1,0 +1,310 @@
+import React, { useCallback, useMemo } from 'react';
+import { AnnotationQuery, AnnotationSupport, GrafanaTheme2, QueryEditorProps } from '@grafana/data';
+import { InlineFormLabel, Select, TextArea, useStyles2 } from '@grafana/ui';
+import { css } from '@emotion/css';
+import { Datasource } from './CHDatasource';
+import { escapeIdentifier } from './sqlGenerator';
+import { CHQuery, CHSqlQuery, EditorType } from 'types/sql';
+import { CHConfig } from 'types/config';
+import { SchemaPicker, SchemaPickerValue } from 'components/queryBuilder/SchemaPicker';
+import useColumns from 'hooks/useColumns';
+
+/** Annotation preset types. */
+type AnnotationPreset = 'custom' | 'change_detection';
+
+/** Builder state for the change-detection preset: a schema selection plus a group-by column. */
+type ChangeDetectionState = SchemaPickerValue & { groupBy?: string };
+
+/** Annotation query extended with the builder state the editor needs to round-trip. */
+interface CHAnnotationQuery extends AnnotationQuery<CHQuery> {
+  preset?: AnnotationPreset;
+  changeDetection?: ChangeDetectionState;
+}
+
+/** Props Grafana passes to an annotation QueryEditor (the base props plus annotation-specific ones). */
+type AnnotationEditorProps = QueryEditorProps<Datasource, CHQuery, CHConfig> & {
+  annotation?: CHAnnotationQuery;
+  onAnnotationChange?: (annotation: CHAnnotationQuery) => void;
+};
+
+const ANNOTATION_REF_ID = 'annotation';
+
+/** Minimal defaults used only when Grafana mounts the editor without an existing annotation. */
+const DEFAULT_ANNOTATION: CHAnnotationQuery = { name: '', enable: true, iconColor: 'red' };
+
+/**
+ * Escape a value for use inside a single-quoted ClickHouse string literal.
+ * Backslashes and single quotes are the only characters that can break out of
+ * the literal. There is no exported string-literal escaper in sqlGenerator
+ * (escapeValue is unexported and bails out on special characters), so this
+ * stays local and small.
+ */
+const escapeStringContent = (value: string): string => value.replace(/\\/g, '\\\\').replace(/'/g, "\\'");
+
+/** Build the SQL annotation target, preserving the existing refId and plugin version. */
+const buildTarget = (existing: Partial<CHQuery> | undefined, rawSql: string): CHSqlQuery => ({
+  pluginVersion: existing?.pluginVersion ?? '',
+  editorType: EditorType.SQL,
+  rawSql,
+  refId: existing?.refId || ANNOTATION_REF_ID,
+});
+
+/**
+ * Resolve the default database and table for the change-detection preset.
+ * Prefers the traces OTel config, then the single-table config when it is
+ * configured for traces, then the general default database with the
+ * conventional otel_traces table. This adapts to both all-databases and
+ * single-table datasources.
+ */
+export function resolveTraceSchema(datasource: Datasource): { database: string; table: string } {
+  const { jsonData } = datasource.settings;
+  const singleTable =
+    jsonData.configMode === 'single-table' && jsonData.signalType === 'traces'
+      ? datasource.getDefaultTable()
+      : undefined;
+  return {
+    database: datasource.getDefaultTraceDatabase() || datasource.getDefaultDatabase(),
+    table: datasource.getDefaultTraceTable() || singleTable || 'otel_traces',
+  };
+}
+
+/**
+ * Generate the change-detection SQL from builder selections.
+ * Per 30-second bucket take the dominant value (topK(1)) so a service running
+ * two versions at once (canary, rolling deploy, multiple replicas) does not
+ * flap between them, then use lagInFrame() to detect transitions per group,
+ * including rollbacks (v1.0 -> v1.1 -> v1.0 produces three annotations). The
+ * outer prev_version != '' guard drops each group's first bucket in the window,
+ * where lagInFrame() returns '' and would otherwise mark every group at the
+ * left edge of the dashboard range.
+ * Identifiers are escaped; the watched map key is emitted as a quoted string literal.
+ */
+export function generateChangeDetectionSQL(opts: ChangeDetectionState): string {
+  const { database, table, column, mapKey, groupBy } = opts;
+  if (!table || !column) {
+    return '-- Select a table and column above to generate the change detection query';
+  }
+
+  const fullTable = database ? `${escapeIdentifier(database)}.${escapeIdentifier(table)}` : escapeIdentifier(table);
+  const columnExpr = mapKey
+    ? `${escapeIdentifier(column)}['${escapeStringContent(mapKey)}']`
+    : escapeIdentifier(column);
+  const groupByCol = escapeIdentifier(groupBy || 'ServiceName');
+  const displayLabel = escapeStringContent(mapKey ? `${column}.${mapKey}` : column);
+
+  return [
+    'SELECT',
+    '  time,',
+    `  ${groupByCol} AS tags,`,
+    `  concat(${groupByCol}, ': ${displayLabel} changed to ', version,`,
+    `    if(prev_version != '', concat(' (was ', prev_version, ')'), '')) AS text,`,
+    `  '${displayLabel} change' AS title`,
+    'FROM (',
+    '  SELECT',
+    '    toStartOfInterval(Timestamp, INTERVAL 30 second) AS time,',
+    `    ${groupByCol},`,
+    `    topK(1)(${columnExpr})[1] AS version,`,
+    `    lagInFrame(topK(1)(${columnExpr})[1])`,
+    `      OVER (PARTITION BY ${groupByCol} ORDER BY time) AS prev_version`,
+    `  FROM ${fullTable}`,
+    '  WHERE $__timeFilter(Timestamp)',
+    `    AND ${columnExpr} != ''`,
+    `  GROUP BY ${groupByCol}, time`,
+    `  ORDER BY ${groupByCol}, time`,
+    ')',
+    "WHERE prev_version != '' AND prev_version != version",
+    'ORDER BY time',
+  ].join('\n');
+}
+
+const PRESET_OPTIONS = [
+  { label: 'Custom SQL', value: 'custom' as AnnotationPreset, description: 'Write your own annotation query' },
+  {
+    label: 'Change Detection',
+    value: 'change_detection' as AnnotationPreset,
+    description: 'Detect when a column value changes (deployments, config changes, rollbacks)',
+    icon: 'rocket',
+  },
+];
+
+/** Annotation query editor with a preset selector and a change-detection builder. */
+const AnnotationQueryEditor = (props: AnnotationEditorProps) => {
+  const { annotation, onAnnotationChange, datasource } = props;
+  const anno: CHAnnotationQuery = annotation ?? DEFAULT_ANNOTATION;
+  const preset = anno.preset || 'custom';
+  const cd: ChangeDetectionState = useMemo(() => anno.changeDetection || {}, [anno.changeDetection]);
+  const styles = useStyles2(getStyles);
+
+  // Reuse the shared column hook for the group-by picker.
+  const columns = useColumns(datasource, cd.database || '', cd.table || '');
+
+  const updateChangeDetection = useCallback(
+    (updates: Partial<ChangeDetectionState>) => {
+      if (!onAnnotationChange) {
+        return;
+      }
+      const newCd = { ...cd, ...updates };
+      onAnnotationChange({
+        ...anno,
+        preset: 'change_detection',
+        changeDetection: newCd,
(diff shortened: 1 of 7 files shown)


src/data/CHAnnotationSupport.tsx
```diff
@@ -0,0 +1,310 @@
...
+ * flap between them, then use lagInFrame() to detect transitions per group,
+ * including rollbacks (v1.0 -> v1.1 -> v1.0 produces three annotations). The
+ * outer prev_version != '' guard drops each group's first bucket in the window,
+ * where lagInFrame() returns '' and would otherwise mark every group at the
+ * left edge of the dashboard range.
+ * Identifiers are escaped; the watched map key is emitted as a quoted string literal.
+ */
+export function generateChangeDetectionSQL(opts: ChangeDetectionState): string {
+  const { database, table, column, mapKey, groupBy } = opts;
+  if (!table || !column) {
```
adamyeats, line 95: A `Map` watch column with no Map Key yet falls through to the bare-identifier branch, so the query compares a `Map` to `''` and ClickHouse errors (`code 27: Cannot convert '' to Map`). When `column` is a `Map` type, require `mapKey` before generating, otherwise return the placeholder.

src/data/CHAnnotationSupport.tsx
```diff
@@ -0,0 +1,310 @@
...
+  const onPresetChange = useCallback(
+    (value: AnnotationPreset) => {
+      if (!onAnnotationChange) {
+        return;
+      }
+      const next: CHAnnotationQuery = { ...anno, preset: value };
+      if (value === 'change_detection') {
+        next.target = buildTarget(anno.target, generateChangeDetectionSQL(cd));
+      } else {
+        next.target = buildTarget(anno.target, anno.target?.rawSql || '');
```
adamyeats: Switching back to Custom rebuilds from `anno.target?.rawSql`, but that was already overwritten with generated SQL when the user entered Change Detection, so their original hand-written query is silently lost. Stash the custom SQL before overwriting it.

src/data/CHAnnotationSupport.tsx
```diff
@@ -0,0 +1,310 @@
...
+      if (legacyRawQuery && !json?.target?.rawSql) {
+        return {
+          ...json,
+          target: buildTarget(json.target, legacyRawQuery),
+        };
+      }
+      return json;
+    },
+
+    getDefaultQuery: (): Partial<CHQuery> => {
```
adamyeats: `getDefaultQuery` emits Change Detection SQL but never sets `preset`, so a fresh annotation opens in Custom SQL mode (`anno.preset || 'custom'`) showing a generated query with the builder hidden. The default query and default editor mode disagree.

src/data/CHAnnotationSupport.tsx
```diff
@@ -0,0 +1,310 @@
...
+  return [
+    'SELECT',
+    '  time,',
+    `  ${groupByCol} AS tags,`,
+    `  concat(${groupByCol}, ': ${displayLabel} changed to ', version,`,
+    `    if(prev_version != '', concat(' (was ', prev_version, ')'), '')) AS text,`,
+    `  '${displayLabel} change' AS title`,
+    'FROM (',
+    '  SELECT',
+    '    toStartOfInterval(Timestamp, INTERVAL 30 second) AS time,',
```
adamyeats, line 121: The time column is hardcoded as `Timestamp`, so the generated query fails ("Unknown identifier Timestamp") on any table whose time column is named differently. Consider a time-column picker or at least a doc note.

src/data/CHAnnotationSupport.tsx
```diff
@@ -0,0 +1,310 @@
...
+      onAnnotationChange({
+        ...anno,
+        target: buildTarget(anno.target, e.currentTarget.value),
+      });
+    },
+    [anno, onAnnotationChange]
+  );
+
+  const groupByOptions = [
+    { label: 'ServiceName', value: 'ServiceName' },
```
adamyeats: `ServiceName` is hardcoded as the first option while the column filter doesn't exclude it, so a table with a real `ServiceName` column (e.g. `otel_traces`) renders it twice. Exclude `'ServiceName'` from the filter.

src/data/CHAnnotationSupport.tsx
```diff
@@ -0,0 +1,310 @@
...
+ * left edge of the dashboard range.
+ * Identifiers are escaped; the watched map key is emitted as a quoted string literal.
+ */
+export function generateChangeDetectionSQL(opts: ChangeDetectionState): string {
+  const { database, table, column, mapKey, groupBy } = opts;
+  if (!table || !column) {
+    return '-- Select a table and column above to generate the change detection query';
+  }
+
+  const fullTable = database ? `${escapeIdentifier(database)}.${escapeIdentifier(table)}` : escapeIdentifier(table);
```
adamyeats: This duplicates `getTableIdentifier` in `sqlGenerator.ts:699` (which already handles the empty-database separator). Export and reuse it so quoting rules can't diverge.

src/data/CHAnnotationSupport.tsx
```diff
@@ -0,0 +1,310 @@
...
+/** Annotation query editor with a preset selector and a change-detection builder. */
+const AnnotationQueryEditor = (props: AnnotationEditorProps) => {
+  const { annotation, onAnnotationChange, datasource } = props;
+  const anno: CHAnnotationQuery = annotation ?? DEFAULT_ANNOTATION;
+  const preset = anno.preset || 'custom';
+  const cd: ChangeDetectionState = useMemo(() => anno.changeDetection || {}, [anno.changeDetection]);
+  const styles = useStyles2(getStyles);
+
+  // Reuse the shared column hook for the group-by picker.
+  const columns = useColumns(datasource, cd.database || '', cd.table || '');
```
adamyeats, line 184: `useColumns` runs here and again inside the rendered `SchemaPicker` with identical args, so `fetchColumns` is issued twice per editor with two loading states that can briefly disagree.

src/data/CHAnnotationSupport.tsx
```diff
@@ -0,0 +1,310 @@
...
+        target: buildTarget(anno.target, e.currentTarget.value),
+      });
+    },
+    [anno, onAnnotationChange]
+  );
+
+  const groupByOptions = [
+    { label: 'ServiceName', value: 'ServiceName' },
+    ...columns
+      .filter((c) => !c.type.startsWith('Map(') && c.name !== 'Timestamp' && c.name !== cd.column)
```
adamyeats: Unlike `SchemaPicker `(which appends an out-of-list controlled value), this filter drops a saved `groupBy` that's a `Map`/`Timestamp` column, so the Select renders blank while the SQL still references it.

adamyeats (changes requested):

src/data/CHAnnotationSupport.tsx
alex-fedotyev: Good catch. Added a guard: when the watch column is a `Map` and no key is chosen yet, `generateChangeDetectionSQL` returns the placeholder instead of emitting a query that compares the whole `Map` to `''`. The column type is threaded through from the schema picker so the guard knows it's a `Map`. 8b31d33

src/data/CHAnnotationSupport.tsx
alex-fedotyev: Right, the round trip was lossy. The editor now stashes the custom `rawSql` when you enter Change Detection and restores it when you switch back, so a hand-written query survives the toggle. 8b31d33

src/data/CHAnnotationSupport.tsx
alex-fedotyev: Fixed. `getDefaultQuery` now returns an empty Custom SQL query, so a fresh annotation opens in Custom mode with an empty box and the mode matches the query. The populated deployment query is seeded only when you actually pick Change Detection. 8b31d33

src/data/CHAnnotationSupport.tsx
alex-fedotyev: Went with the doc note for now, since the generated SQL stays editable. Documented that the builder uses `Timestamp` (the OTel convention for the traces and logs tables) and how to point the query at a differently named time column or drop to Custom SQL. A time-column picker is a reasonable follow-up if it comes up. 8b31d33

src/data/CHAnnotationSupport.tsx
alex-fedotyev: Good catch. The Group By options now exclude `ServiceName` (along with `Map` columns, the timestamp, and the watched column) from the column-derived list, with `ServiceName` kept as the single leading option, so a real `ServiceName` column no longer doubles up. 8b31d33

src/data/CHAnnotationSupport.tsx
alex-fedotyev: Done. Exported `getTableIdentifier` from `sqlGenerator.ts` and reused it here, so the quoting and empty-database separator rules stay in one place. 8b31d33

src/data/CHAnnotationSupport.tsx
alex-fedotyev: Fixed. `useColumns` now dedupes in-flight fetches keyed by datasource, database, and table, so the editor and the `SchemaPicker` it renders share a single `fetchColumns` request instead of racing two loading states. The entry is evicted once the request settles, so it never serves stale schema. 8b31d33

src/data/CHAnnotationSupport.tsx
alex-fedotyev: Good point, matched the `SchemaPicker` behavior. `buildGroupByOptions` now appends a saved `groupBy` that isn't in the filtered list (a `Map` or timestamp column from an older query), so the Select shows the real value instead of rendering blank while the SQL still references it. 8b31d33

alex-fedotyev:
Thanks for the thorough review. Pushed 8b31d33 with all eight addressed, replies inline on each thread.

On testing, I added unit coverage for the cases these comments surfaced: the `Map`-without-key guard and the database/table quoting in `generateChangeDetectionSQL`, a `buildGroupByOptions` suite (ServiceName de-duplication, the exclusion filter, and the saved-value append), and stash/restore coverage for the Custom SQL round trip. The annotation suite is at 31 tests and the full suite stays green.

adamyeats (approved):
LGTM! 🚢

PR #1924, opened 2026-06-24
feat(datasource): config macros for OTel logs/traces database and table

alex-fedotyev:
> **Status (2026-06-30): draft, reworking on top of #1802.** Per review, these macros currently expand in the frontend `applyTemplateVariables` hook, which the backend does not run for alerting or recorded queries, so the raw token would reach ClickHouse on those paths. Moving them to backend macros built on the macropro engine in #1802 so they expand server-side. The summary below describes the original frontend approach and will be updated once the rework lands.

---

## Summary

Adds four query macros that resolve to the database and table configured on the data source for each OpenTelemetry signal:

- `$__logsDatabase` / `$__logsTable`
- `$__tracesDatabase` / `$__tracesTable`

They expand to the data source's configured per-signal values (`logs.defaultDatabase` / `logs.defaultTable` / `traces.defaultDatabase` / `traces.defaultTable`), falling back to the connection default database and the conventional OTel table names (`otel_logs` / `otel_traces`) when a value is unset. Resolution happens in `applyTemplateVariables`, next to the existing `$__conditionalAll` handling, so it covers both panel queries and template-variable queries.

## Why

Queries and dashboards currently hard-code table names like `otel_logs`, which only works when the data source's default database happens to hold those tables. With these macros a query can reference the data source's configured schema instead:

- A single-table data source binds to its configured table automatically.
- A classic data source binds to its configured logs/traces defaults, and a dashboard can still override per panel.

This is the building block that lets the pre-built OTel dashboards follow whatever data source is selected instead of assuming a fixed schema.

## Test plan

- [x] Unit tests in `src/data/CHDatasource.test.ts`: macros resolve to the configured per-signal database/table, and fall back to the default database and `otel_logs` / `otel_traces` when per-signal config is unset.
- [x] `npm run typecheck`, `npm run lint`, `npm run test:ci`, `prettier --check`, and `cspell` all clean.
- [x] Docs: the four macros are documented in the query-editor Macros reference.

## Special notes for the reviewer

- Macro names (`$__logsDatabase` and friends) are open to your preference; happy to rename if a different convention fits better.
- They expand to bare identifiers, matching the existing `$clickhouse_adhoc_query` behavior. If you would prefer backtick-quoted identifiers for safety with non-standard names, that is a small change.
- No UI surface in this PR (query-engine + docs only). [ui-check: allow]

Please add the `changelog` label so it surfaces in the next release notes.

### DIFF
--- a/docs/sources/query-editor.md
+++ b/docs/sources/query-editor.md
@@ -285,8 +285,14 @@ WHERE $__timeFilter(date_time)
 | `$__timeInterval(columnName)`                | Interval from panel time range (seconds), for grouping.                                          | `toStartOfInterval(toDateTime(column), INTERVAL 20 second)`                                                                           |
 | `$__timeInterval_ms(columnName)`             | Interval from panel time range (milliseconds), for grouping.                                     | `toStartOfInterval(toDateTime64(column, 3), INTERVAL 20 millisecond)`                                                                 |
 | `$__conditionalAll(condition, $templateVar)` | Uses the first parameter when the template variable does not select all values; otherwise `1=1`. | `condition` or `1=1`                                                                                                                  |
+| `$__logsDatabase`                            | Configured OTel logs database for this data source (or the default database when unset).         | `otel`                                                                                                                                |
+| `$__logsTable`                               | Configured OTel logs table for this data source (or `otel_logs` when unset).                     | `otel_logs`                                                                                                                           |
+| `$__tracesDatabase`                          | Configured OTel traces database for this data source (or the default database when unset).       | `otel`                                                                                                                                |
+| `$__tracesTable`                             | Configured OTel traces table for this data source (or `otel_traces` when unset).                 | `otel_traces`                                                                                                                         |
 
 You can also use brace notation `{}` when the macro parameter must contain a query or other expression.
+
+The `$__logsDatabase` / `$__logsTable` / `$__tracesDatabase` / `$__tracesTable` macros resolve to the database and table configured on the data source for that signal. Use them instead of hard-coding table names so a dashboard follows whatever the selected data source is configured for.
 
 ## Next steps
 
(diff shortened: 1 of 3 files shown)


SpencerTorres (approved):

adamyeats (changes requested):
As written, this works for interactive dashboards but silently fails for alerting/recorded queries on those same dashboards: backend query evaluation never runs the frontend `applyTemplateVariables` hook where these expand, so the literal `$__logsTable` token reaches ClickHouse and errors.

Let's hold off and turn these into backend macros built on #1802. `sqlds`'s `MacroFunc` only receives the query (no datasource config), whereas #1802's `macropro` expands in `MutateQueryData` where the per-signal `jsonData` is in scope, so the macros can actually read `logs.defaultDatabase` / `traces.defaultTable`. Identifier quoting can come along for the ride. Wdyt?

alex-fedotyev:
Makes sense, and you're right about the failure mode. These expand in `applyTemplateVariables`, which the backend doesn't run for alerting or recorded queries, so the raw `$__logsTable` token would reach ClickHouse and error on those paths. That's exactly what the pre-built OTel dashboards will exercise, so frontend-only expansion is a trap here. I'll rework them as backend macros on top of #1802.

Two notes before I start:

- The Go `Settings` struct only parses `defaultDatabase` today; the per-signal `logs.defaultDatabase` / `traces.defaultTable` values exist only in the frontend config accessors. So this also needs `Settings` extended to parse the nested `logs`/`traces` jsonData before a handler can read them. Small, but flagging it.
- Identifier quoting: happy to fold that into the macropro handlers.

What's your read on #1802's timeline? If it's close I'll stack this on `refactor/use-macropro` so it's ready to merge right after; otherwise I'll rebase onto `main` once it lands. Converting to draft in the meantime.

adamyeats:
@alex-fedotyev https://github.com/grafana/clickhouse-datasource/pull/1802 has now landed in `main`!

PR #1927, opened 2026-06-24
fix: render SimpleAggregateFunction string columns correctly

### LINKED ISSUE #1926: SimpleAggregateFunction(any, String/Nullable(String)) columns render as NULL in Grafana
## Description

When querying a table with `SimpleAggregateFunction(any, Nullable(String))` or `SimpleAggregateFunction(any, String)` columns via the Grafana ClickHouse datasource plugin, the values display as `null` in the Grafana table panel. However, the data IS present — wrapping in `toString()` reveals the actual values.

Numeric types like `SimpleAggregateFunction(any, Nullable(UInt16))`, `Nullable(Float32)`, and `Bool` render correctly. Only String-based types are affected.

## Steps to Reproduce

1. Create an `AggregatingMergeTree` table with `SimpleAggregateFunction(any, Nullable(String))` columns:

```sql
CREATE TABLE test_db.test_table
(
    `id` String,
    `name` SimpleAggregateFunction(any, Nullable(String)),
    `status` SimpleAggregateFunction(any, Nullable(UInt16))
)
ENGINE = AggregatingMergeTree()
ORDER BY id;

INSERT INTO test_db.test_table VALUES ('abc', 'hello', 200);
```

2. Query via Grafana: `SELECT id, name, status FROM test_db.test_table LIMIT 1`
   - `id` → `abc` ✅
   - `name` → `null` ❌ (should be `hello`)
   - `status` → `200` ✅

3. Query via Grafana: `SELECT toString(name) FROM test_db.test_table LIMIT 1` → shows `hello` ✅

4. Query via ClickHouse client/console: `SELECT name FROM test_db.test_table LIMIT 1` → shows `hello` ✅

## Verification

```sql
-- Displays as null in Grafana, but data exists:
SELECT 
  name,                          -- shows null in Grafana
  toString(name) AS name_str,    -- shows 'hello'
  isNull(name) AS is_null,       -- shows 0 (NOT null!)
  length(name) AS len,           -- shows 5
  toTypeName(name) AS col_type   -- shows 'SimpleAggregateFunction(any, Nullable(String))'
FROM test_db.test_table LIMIT 1
```

## Expected Behavior

`SimpleAggregateFunction(any, String)` and `SimpleAggregateFunction(any, Nullable(String))` columns should render their values the same way numeric `SimpleAggregateFunction` columns do.

## Root Cause Analysis

The `SimpleAggregateFunction()` converter in `pkg/converters/converters.go` uses a generic `jsonConverter` with `scanType: reflect.TypeOf((*interface{})(nil)).Elem()` for ALL `SimpleAggregateFunction` types. For numeric types, `json.Marshal` produces valid JSON numbers. For String/Nullable(String), the Go driver's scan into `interface{}` produces a value that `jsonConverter` cannot properly serialize.

## Suggested Fix

Follow the same pattern as `LowCardinality()` handling — extract the inner type from `SimpleAggregateFunction(func_name, <inner_type>)` and delegate to the existing converter for that inner type:

```go
// Similar to how LowCardinality is already handled:
if innerType, ok := extractSimpleAggregateFunctionType(columnType); ok {
    return GetConverter(innerType)
}
```

## Environment

- Grafana ClickHouse datasource plugin version: 4.17.0
- ClickHouse server version: 26.2.1.453 (ClickHouse Cloud)
- Table engine: AggregatingMergeTree / SharedAggregatingMergeTree

## Workaround

Wrap string columns in `toString()`, or create a VIEW that casts all `SimpleAggregateFunction` string columns.

SaiPrasanna9:
## Summary

`SimpleAggregateFunction(*, String)` and `SimpleAggregateFunction(*, Nullable(String))` columns render as blank/null in Table view. This PR fixes the issue by adding specific regex matchers for SAF string types that use native scan types, placed before the catch-all.

Also adds `extractSimpleAggregateFunctionType()` for the `GetConverter()` path, following the same unwrap-and-delegate pattern used by `LowCardinality` and the ClickHouse JDBC driver (clickhouse-java#2022).

## Problem

The existing catch-all regex converter scans all SAF types into `interface{}` and uses `jsonConverter`. This works for numerics, bools, and arrays (which `json.Marshal` serializes correctly), but for strings it produces `json.RawMessage("hello")` which is invalid JSON and renders blank.

## Fix

- Add regex matchers for `SimpleAggregateFunction(*, String)` and `SimpleAggregateFunction(*, Nullable(String))` before the catch-all, with native `*string` / `**string` scan types
- Add `extractSimpleAggregateFunctionType()` helper for the `GetConverter()` path
- Non-string SAF types continue to use the existing `jsonConverter` catch-all (which already works)

## Tests

- Tests exercise both the production regex path (`findConverterWithRegex`) and the `GetConverter` delegation path
- Covers: String, Nullable(String), Nullable(String) nil, numeric→JSON (production), array→JSON (production), GetConverter delegation, and `extractSimpleAggregateFunctionType` parsing

Fixes #1926

### DIFF
--- a/pkg/converters/converters.go
+++ b/pkg/converters/converters.go
@@ -30,28 +30,30 @@ type Converter struct {
 // For example `^Date` and `^DateTime` could conflict when matching `DateTime64`
 var matchRegexes = map[string]*regexp.Regexp{
 	// for complex Arrays e.g. Array(Tuple)
-	"Array()":                   regexp.MustCompile(`^Array\(.*\)`),
-	"Date":                      regexp.MustCompile(`^Date\(?`),
-	"Decimal":                   regexp.MustCompile(`^Decimal`),
-	"FixedString()":             regexp.MustCompile(`^Nullable\(FixedString\(.*\)\)`),
-	"IP":                        regexp.MustCompile(`^IPv[4,6]`),
-	"LowCardinality()":          regexp.MustCompile(`^LowCardinality\(([^)]*)\)`),
-	"LowCardinality(Nullable)":  regexp.MustCompile(`^LowCardinality\(Nullable([^)]*)\)`),
-	"Map()":                     regexp.MustCompile(`^Map\(.*\)`),
-	"Nested()":                  regexp.MustCompile(`^Nested\(.*\)`),
-	"Nullable(Date)":            regexp.MustCompile(`^Nullable\(Date\(?`),
-	"Nullable(Decimal)":         regexp.MustCompile(`^Nullable\(Decimal`),
-	"Nullable(IP)":              regexp.MustCompile(`^Nullable\(IP`),
-	"Nullable(String)":          regexp.MustCompile(`^Nullable\(String`),
-	"Point":                     regexp.MustCompile(`^Point`),
-	"SimpleAggregateFunction()": regexp.MustCompile(`^SimpleAggregateFunction\(.*\)`),
-	"Tuple()":                   regexp.MustCompile(`^Tuple\(.*\)`),
-	"Variant":                   regexp.MustCompile(`^Variant`),
-	"Dynamic":                   regexp.MustCompile(`^Dynamic`),
-	"JSON":                      regexp.MustCompile(`^JSON`),
-	"Nullable(JSON)":            regexp.MustCompile(`^Nullable\(JSON`),
-	"Enum":                      regexp.MustCompile(`^Enum(8|16)\(.*\)`),
-	"Nullable(Enum)":            regexp.MustCompile(`^Nullable\(Enum(8|16)\(.*\)\)`),
+	"Array()":                         regexp.MustCompile(`^Array\(.*\)`),
+	"Date":                            regexp.MustCompile(`^Date\(?`),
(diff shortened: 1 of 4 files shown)


SpencerTorres (approved):

adamyeats:
@SaiPrasanna9 Thanks for your contribution! Before I review, could you please rebase and GPG sign your commits? Sorry for the trouble here, this is a new requirement on all of our repos, I'm afraid.

SaiPrasanna9:
@adamyeats Done — rebased onto main and commits are GPG signed now. Thanks!

adamyeats (changes requested):
Looks good! I've tested this locally and can confirm it seems to be working as expected. No comments on the code. For the sake of completeness, can I request that you add an E2E test exercising `SimpleAggregateFunction(any, String)` / `Nullable(String)`? A small addition to the fixture data + a table-panel assertion would lock the fix in.

adamyeats (approved):
Thanks for addressing everything @SaiPrasanna9 , LGTM! Merging! 🚢

PR #1928, opened 2026-06-24
feat: native type support for SimpleAggregateFunction numeric/bool/date columns

### LINKED ISSUE #1926: SimpleAggregateFunction(any, String/Nullable(String)) columns render as NULL in Grafana

SaiPrasanna9:
## Summary

- Resolves `SimpleAggregateFunction`-wrapped numeric, boolean, and datetime columns to their native Grafana field types instead of falling through to the JSON catch-all
- Enables numeric sorting, graphing, and threshold evaluation for SAF columns in Table/TimeSeries panels
- Uses `interface{}` scan type with type-asserting converter functions to work around the ClickHouse Go driver's non-standard scan types

## Context

Builds on #1927 which fixed SAF String/Nullable(String) rendering. This PR includes all changes from #1927 — merging this one alone is sufficient; #1927 can be closed without merging.

The ClickHouse Go driver transparently unwraps SAF to the inner type at the scan level, but sends non-standard Go types (`*uint16`, `*float32`, etc.) that `database/sql.Scan` cannot handle with typed destinations. The solution scans into `interface{}` and type-asserts in the converter function.

### Types now natively supported:
| Inner Type | Grafana FieldType |
|---|---|
| UInt8/16/32/64, Int8/16/32/64 | number (respective width) |
| Float32/64 | number (float) |
| Bool | boolean |
| Date, Date32, DateTime, DateTime64 (with optional timezone) | time |
| Nullable variants of all above | nullable equivalents |
| Array, Map, Tuple (unchanged) | JSON |

## Test plan
- [x] Unit tests for type resolution and value conversion (including null handling and bare-value driver inconsistency)
- [x] `mage test` passes (all Go tests)
- [x] `gofmt` clean
- [x] Verified in local Grafana against ClickHouse with SAF table containing all types
- [ ] Maintainer CI approval needed (external contributor)

Fixes #1926

### DIFF
--- a/pkg/converters/converters.go
+++ b/pkg/converters/converters.go
@@ -548,6 +548,9 @@ func safConvertNullableUint8(in interface{}) (interface{}, error) {
 	if val, ok := v.(*uint8); ok {
 		return val, nil
 	}
+	if val, ok := v.(uint8); ok {
+		return &val, nil
+	}
 	return jsonConverter(in)
 }
 
@@ -558,6 +561,9 @@ func safConvertNullableUint16(in interface{}) (interface{}, error) {
 	}
 	if val, ok := v.(*uint16); ok {
 		return val, nil
+	}
+	if val, ok := v.(uint16); ok {
+		return &val, nil
 	}
 	return jsonConverter(in)
 }
@@ -570,6 +576,9 @@ func safConvertNullableUint32(in interface{}) (interface{}, error) {
 	if val, ok := v.(*uint32); ok {
 		return val, nil
 	}
+	if val, ok := v.(uint32); ok {
+		return &val, nil
+	}
 	return jsonConverter(in)
 }
 
@@ -580,6 +589,9 @@ func safConvertNullableUint64(in interface{}) (interface{}, error) {
 	}
 	if val, ok := v.(*uint64); ok {
 		return val, nil
+	}
+	if val, ok := v.(uint64); ok {
+		return &val, nil
 	}
 	return jsonConverter(in)
 }
@@ -592,6 +604,9 @@ func safConvertNullableInt8(in interface{}) (interface{}, error) {
 	if val, ok := v.(*int8); ok {
 		return val, nil
 	}
+	if val, ok := v.(int8); ok {
+		return &val, nil
+	}
 	return jsonConverter(in)
 }
 
@@ -602,6 +617,9 @@ func safConvertNullableInt16(in interface{}) (interface{}, error) {
 	}
 	if val, ok := v.(*int16); ok {
 		return val, nil
+	}
+	if val, ok := v.(int16); ok {
+		return &val, nil
 	}
 	return jsonConverter(in)
 }
@@ -614,6 +632,9 @@ func safConvertNullableInt32(in interface{}) (interface{}, error) {
 	if val, ok := v.(*int32); ok {
 		return val, nil
 	}
+	if val, ok := v.(int32); ok {
+		return &val, nil
+	}
 	return jsonConverter(in)
 }
 
@@ -624,6 +645,9 @@ func safConvertNullableInt64(in interface{}) (interface{}, error) {
 	}
 	if val, ok := v.(*int64); ok {
 		return val, nil
+	}
+	if val, ok := v.(int64); ok {
+		return &val, nil
 	}
 	return jsonConverter(in)
 }
@@ -636,6 +660,9 @@ func safConvertNullableFloat32(in interface{}) (interface{}, error) {
 	if val, ok := v.(*float32); ok {
 		return val, nil
 	}
+	if val, ok := v.(float32); ok {
+		return &val, nil
+	}
 	return jsonConverter(in)
 }
 
@@ -647,6 +674,9 @@ func safConvertNullableFloat64(in interface{}) (interface{}, error) {
 	if val, ok := v.(*float64); ok {
 		return val, nil
 	}
+	if val, ok := v.(float64); ok {
+		return &val, nil
+	}
 	return jsonConverter(in)
 }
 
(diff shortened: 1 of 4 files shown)


SpencerTorres (approved):
this works for now, but I'm wondering if we need to consider a driver level change to support these types better. Thanks for the PR!

pkg/converters/converters.go
```diff
@@ -389,9 +382,339 @@ func ClickHouseConverters() []sqlutil.Converter {
...
+		})
+	}
+	return converters
+}
+
+// SAF converter functions. The ClickHouse driver scans SAF values into interface{},
+// and the underlying value is the native Go type matching the inner column type.
+
+func safConvertUint8(in interface{}) (interface{}, error) {
+	if v, ok := (*(in.(*interface{}))).(uint8); ok {
```
adamyeats: The non-nullable converters only accept the bare value, but every Nullable converter also handles the pointer form, because (per your own `"bare value (driver inconsistency)"` test) the driver is inconsistent about bare-vs-pointer. 

Worth handling both forms here too: these `unexpected type` errors abort the whole frame (`sqlutil`'s `Append` returns on the first converter error), so any representation you don't cover turns into a failed panel rather than a single bad cell. Returning the error is the right call as a guard. The goal is just to make it unreachable for inputs the driver actually produces.

pkg/converters/converters.go
```diff
@@ -389,9 +382,339 @@ func ClickHouseConverters() []sqlutil.Converter {
...
+
+func generateSAFConverters() []sqlutil.Converter {
+	var converters []sqlutil.Converter
+	for _, entry := range safTypeMappings {
+		pattern := `^SimpleAggregateFunction\([^,]+,\s*` + entry.innerPattern + `\)$`
+		converters = append(converters, sqlutil.Converter{
+			Name:           "SimpleAggregateFunction(" + entry.innerPattern + ")",
+			InputScanType:  reflect.TypeOf((*interface{})(nil)).Elem(),
+			InputTypeRegex: regexp.MustCompile(pattern),
+			InputTypeName:  "SimpleAggregateFunction(" + entry.innerPattern + ")",
```
adamyeats: Since `entry.innerPattern` is a regex, `InputTypeName` ends up containing literal backslashes (e.g. `SimpleAggregateFunction(Nullable\(UInt8\)))`, which reads oddly in logs/debugging. Consider a separate human-readable label distinct from the regex pattern.

adamyeats (changes requested):
LGTM and has been tested locally. Just a couple of points below.

SaiPrasanna9:
@adamyeats Addressed both — non-nullable converters now handle the pointer form (with unit tests covering the driver inconsistency), and `InputTypeName` uses a human-readable label instead of the raw regex pattern.

SaiPrasanna9:
@adamyeats also just wondering when the next grafana release is, so that we can use the fix on our instance

adamyeats (approved):
@SaiPrasanna9 LGTM, please rebase again and we'll merge this. I'm not able to comment about when the next release is, but it should be soon.

SaiPrasanna9:
@adamyeats Rebased and pushed.

PR #1935, opened 2026-06-27
adds dsconfig schema

jcolladokuri:
## Type of Change

- [ ] 🚀 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧹 Refactor / Chore

---

We are investing in creating uniform datasource config schemas across datasources to primarily have a way for agents to read and understand the fields that are necessary for configuring a datasource and guide users via a chat.

This PR adds support for dsconfig schema:

1. creates the new `dsconfig.json`
2. Sets up generated schema files base from `dsconfig.json` for different downstream dependencies to read from it.

This is to have a way for LLMs (for now, potentially expandable to other areas later on) to be able to successfully manage datasoruces irrespective of how different all the fields may be across each of them


pkg/schema/models/settings.go
yesoreyeram, line 1: can we just use https://github.com/grafana/clickhouse-datasource/blob/3ffd8f41edb3788f2f70c8af04004b68b5d4c7ff/pkg/plugin/settings.go#L19 instead?

pkg/schema/dsconfig.json
yesoreyeram, line 1: seems `dsconfig.json` still using `.` convention for `id` fields and some validations are also missing. Can we use https://raw.githubusercontent.com/grafana/dsconfig/refs/heads/schema-discovery/registry/grafana-clickhouse-datasource/dsconfig.json ( this json is the one used in storybook )

Issue #958, opened 2024-08-22
SQL parsing into an AST fails on INTERVAL and SAMPLE statements


eKrajnak:
**What happened**:
Clickhouse query which contains:
- "INTERVAL 1 MINUTE" in where clause 
or
- "FROM table **SAMPLE 1/10**"
fails at SQL parsing into an AST when adhoc filter is applied

**What you expected to happen**:
Apply adhoc filter and perform query

**How to reproduce it (as minimally and precisely as possible)**:

Apply any adhoc filter and run the query:
```
SELECT *
FROM data
SAMPLE 1/10
WHERE time > NOW() - INTERVAL 1 MINUTE
```
and watch javascript console log:
```
Unexpected int token: "1". Instead, I was expecting to see one of the following:

    - A "lparen" token
    - A "kw_cross" token
    - A "kw_left" token
    - A "kw_right" token
    - A "kw_full" token
    - A "comma" token
    - A "kw_inner" token
    - A "kw_where" token
    - A "kw_join" token
    - A "kw_group" token
    - A "kw_order" token
    - A "kw_offset" token
    - A "kw_for" token
    - A "kw_limit" token
    - A "kw_fetch" token
    - A "kw_union" token
    - A "semicolon" token
```

**Environment**:

- Grafana version: 11.1.4 (2355de00c6)
- Plugin version: 4.3.2
- OS Grafana is installed on: Rocky Linux release 8.10 (Green Obsidian)
- User OS & Browser: macOS 14.6.1 and Chrome 127.0.6533.120

SpencerTorres:
The AST is notorious for having SQL syntax problems. I'll try to reproduce this locally and see if it can be patched.

@bossinc We might need to start exploring other options for the AST. I believe one of its main purposes is for extracting the table name (at least for adhoc filters). Maybe we can add a field to the SQL UI where people can manually write the table name. Or we can find a good Go module on the backend to handle it and use that to get an AST via HTTP request to the plugin.

eKrajnak:
Thank you.
For INTERVAL 1 MINUTE I can use workaround "toIntervalMinute(1)" already now.
But for SAMPLE I have no solution. I took a fast look on AST. If I'm correct it's AST plugin for Postgresql and it knows "TABLESAMPLE" with the very similar meaning. Maybe it could be just renamed.

SpencerTorres:
[...]

PR #1936, opened 2026-06-30
fix(adhoc): resolve ad-hoc filter table with the ClickHouse parser

alex-fedotyev:
## Summary

Ad-hoc filters silently stop applying on queries that use ClickHouse-specific syntax. `getTable()` resolved the target table with `pgsql-ast-parser`, and that Postgres grammar throws on valid ClickHouse SQL (`SAMPLE`, `INTERVAL`, lambdas, backtick-quoted names). On a throw it returned `''`, so `AdHocFilter.apply()` became a no-op and the filter was dropped with no error shown to the user (#958).

`getTable()` now resolves the table with the in-repo ClickHouse parser (`src/ch-parser`) that already backs SQL autocomplete, through one shared `parseSelect`. It resolves the outer `FROM` at each select level, descending into that `FROM`'s own subquery (and only it) or resolving a `FROM` that references a CTE to the CTE's underlying table. Table functions and top-level `UNION`s resolve to no table.

Refs #958, #714

## Changes

- `src/ch-parser/parser.ts`:
  - `node.from` is set only at paren depth 0, so a `FROM` nested in a function call (`EXTRACT`/`trim(... FROM ...)`) does not shadow the real table;
  - keyword-named tables resolve; the clause-starting keywords are refused only in the unqualified first `FROM` segment, so `default.order` resolves while `FROM  WHERE ...` does not take `WHERE` as the table;
  - a Grafana `${var}` resolves in the `FROM` position on either side of a dot, adjacency-bounded so an unterminated `${` while typing does not consume the rest of the query;
  - table functions are flagged and a top-level `UNION` is flagged;
  - CTE definitions are recorded so a `FROM` that names a CTE resolves to its body;
  - a semicolon ends the statement; `parseSelect` is exported and reused by autocomplete.
- `src/data/ast.ts`: `getTable()` resolves via `parseSelect` with subquery descent, CTE resolution, and no table for a table function or top-level `UNION`.
- `src/components/suggestions.ts`: `getSuggestions` reuses `parseSelect` and offers variable completions when the `FROM` target is a `${var}`.
- `src/data/adHocFilter.ts`: `apply()` resolves the table per call rather than caching it on the shared instance (so panels on different tables each filter); the presence re-check runs only for an explicit tag-source target; a trailing semicolon is stripped and the settings clause is appended on a new line so a trailing line comment cannot swallow it.

## Behavior changes

The ad-hoc parity suite in `src/data/adHocFilter.test.ts` still passes, so queries the old code filtered still filter the same way. Beyond the ClickHouse syntax that now resolves, a few shapes the Postgres grammar previously accepted or rejected also change, all user-facing:

- `WITH ... SELECT` and `UNION` statements: the old code returned `''` for these (`type !== 'select'`), so ad-hoc filters were dropped. A `WITH` now resolves through the CTE; a top-level `UNION` still resolves to no table (per-branch keying is tracked in #2172).
- `db.<keyword>` table names (e.g. `default.order`) now resolve.
- An unqualified table named like a clause keyword (e.g. `FROM order`) needs backtick quoting (`` FROM `order` ``) to resolve, so `FROM  WHERE ...` is not misread mid-edit.
- With one shared `AdHocFilter` across a dashboard, each panel now resolves its own table instead of the first-rendered panel deciding the table for all.

## Local verification

Built the plugin and ran it in Grafana 13.2.1 next to the official 4.21.2 build, both pointed at the public `otel_v2` demo dataset. A table panel groups by `ServiceName` over an `INTERVAL`-based query, with an ad-hoc filter `ServiceName = frontend`.

| | Official 4.21.2 | This PR |
| --- | --- | --- |
| Rows returned | all 15 services (filter dropped) | 1 (`frontend`) |
| SQL sent to ClickHouse | `... GROUP BY ServiceName ...` (no filter) | `... settings additional_table_filters={'otel_v2.otel_logs' : ' ServiceName = \'frontend\' '}` |

| Before (official 4.21.2) | After (this PR) |
| --- | --- |
| before | after |

Autocomplete still resolves columns through the shared parser (a regression check for the parser change):

autocomplete

## Test plan

- [x] `npx jest` (full suite green; the ad-hoc parity suite and the autocomplete suite pass)
- [x] `npx tsc --noEmit`
- [x] `npx eslint .`
- [x] `npx prettier --check`
- [x] `npm run spellcheck`
- [x] Manual verification in Grafana 13.2.1, official build vs this build, against real ClickHouse

Tests cover a `getTable` matrix (shapes the old parser dropped; no-regression rows pinned to its output; CTE resolution; table functions; UNION; `db.<keyword>`; subquery-descent and Grafana-variable cases), `src/ch-parser/parser.test.ts` for `node.from`, CTE capture and the UNION flag, `apply()` cases including per-call resolution, the trailing-comment newline and a spaced-dot name, and autocomplete for keyword-named tables and `${var}` in FROM.

### DIFF
--- a/src/data/ast.test.ts
+++ b/src/data/ast.test.ts
@@ -1,4 +1,4 @@
-import { getFields, sqlToStatement } from './ast';
+import { getFields, getTable, sqlToStatement } from './ast';
 import { toSql } from 'pgsql-ast-parser';
 
 describe('ast', () => {
@@ -41,6 +41,127 @@ describe('ast', () => {
       expect(errLog).toHaveBeenCalledTimes(0);
       expect(stm).not.toEqual({});
       expect(astSql).not.toBeFalsy();
+    });
+  });
+
+  describe('getTable', () => {
+    // Each case asserts the physical table an ad-hoc filter targets. The
+    // ClickHouse-specific shapes (and UNION / CTE) returned '' on the previous
+    // pgsql-ast-parser implementation: it threw on valid ClickHouse syntax, so
+    // the ad-hoc filter was silently dropped. See
+    // grafana/clickhouse-datasource#958 and #714.
+    const cases: Array<{ group: string; label: string; sql: string; expected: string }> = [
+      // Baseline: the previous implementation handled these too.
+      {
+        group: 'baseline',
+        label: 'simple select',
+        sql: 'SELECT ServiceName FROM otel_logs LIMIT 10',
+        expected: 'otel_logs',
+      },
+      {
+        group: 'baseline',
+        label: 'database-qualified',
+        sql: 'SELECT * FROM default.otel_logs',
+        expected: 'default.otel_logs',
+      },
+      {
+        group: 'baseline',
+        label: 'double-quoted identifiers',
+        sql: 'SELECT * FROM "default"."otel_logs"',
+        expected: 'default.otel_logs',
+      },
+      { group: 'baseline', label: 'preserves table casing', sql: 'SELECT * FROM fooTable', expected: 'fooTable' },
+
+      // ClickHouse-specific syntax: pgsql-ast-parser threw on each of these.
+      {
+        group: 'ch-syntax',
+        label: 'SAMPLE clause',
+        sql: 'SELECT ServiceName FROM otel_traces SAMPLE 0.1',
+        expected: 'otel_traces',
+      },
+      { group: 'ch-syntax', label: 'FINAL modifier', sql: 'SELECT * FROM otel_logs FINAL', expected: 'otel_logs' },
+      {
+        group: 'ch-syntax',
+        label: 'INTERVAL literal',
+        sql: 'SELECT * FROM otel_logs WHERE Timestamp >= now() - INTERVAL 1 HOUR',
+        expected: 'otel_logs',
+      },
+      {
+        group: 'ch-syntax',
+        label: 'Map access in SELECT',
+        sql: `SELECT ResourceAttributes['service.name'] FROM otel_logs`,
+        expected: 'otel_logs',
+      },
+      {
+        group: 'ch-syntax',
+        label: 'lambda expression',
+        sql: 'SELECT arrayMap(x -> x * 2, [1, 2, 3]) FROM otel_logs',
+        expected: 'otel_logs',
+      },
+      {
+        group: 'ch-syntax',
+        label: 'existing SETTINGS clause',
+        sql: 'SELECT * FROM otel_logs SETTINGS max_threads = 4',
+        expected: 'otel_logs',
+      },
+
+      // Multi-table: the contract is the first physical table.
+      {
+        group: 'multi-table',
+        label: 'JOIN returns the first table',
+        sql: 'SELECT l.Body, t.SpanName FROM otel_logs AS l JOIN otel_traces AS t ON l.TraceId = t.TraceId',
+        expected: 'otel_logs',
+      },
+      {
+        group: 'multi-table',
+        label: 'subquery in FROM descends to the physical table',
+        sql: 'SELECT * FROM (SELECT ServiceName FROM otel_logs) AS sub',
+        expected: 'otel_logs',
+      },
+      {
+        group: 'multi-table',
+        label: 'UNION ALL returns the first table',
+        sql: 'SELECT ServiceName FROM otel_logs UNION ALL SELECT ServiceName FROM otel_traces',
+        expected: 'otel_logs',
+      },
+      {
+        group: 'multi-table',
+        label: 'CTE resolves to the underlying physical table',
+        sql: 'WITH recent AS (SELECT * FROM otel_logs) SELECT * FROM recent',
+        expected: 'otel_logs',
+      },
+
+      // Grafana macros embedded (the real dashboard-panel shape).
+      {
+        group: 'grafana-macros',
+        label: 'timeFilter + timeInterval',
+        sql: 'SELECT $__timeInterval(Timestamp) AS t, count() FROM otel_logs WHERE $__timeFilter(Timestamp) GROUP BY t',
+        expected: 'otel_logs',
+      },
+    ];
+
+    it.each(cases)('[$group] $label', ({ sql, expected }) => {
+      expect(getTable(sql)).toBe(expected);
+    });
+
+    it('returns an empty string when the query has no table', () => {
+      expect(getTable('SELECT 1')).toBe('');
+    });
+
+    it('returns an empty string when the input cannot be parsed as a select', () => {
+      expect(getTable('select not sql')).toBe('');
+    });
+
+    // Regression guard for the silent-drop bug: each of these returned '' on
+    // the previous implementation, which made AdHocFilter.apply() a no-op.
+    it.each([
+      ['SAMPLE', 'SELECT ServiceName FROM otel_traces SAMPLE 0.1'],
+      ['INTERVAL', 'SELECT * FROM otel_logs WHERE Timestamp >= now() - INTERVAL 1 HOUR'],
+      ['lambda', 'SELECT arrayMap(x -> x * 2, [1, 2, 3]) FROM otel_logs'],
+      ['UNION', 'SELECT ServiceName FROM otel_logs UNION ALL SELECT ServiceName FROM otel_traces'],
+      ['CTE', 'WITH recent AS (SELECT * FROM otel_logs) SELECT * FROM recent'],
+    ])('now resolves a table for the previously-dropped %s shape', (_label, sql) => {
+      expect(getTable(sql)).not.toBe('');
     });
   });
 });
--- a/src/data/ast.ts
+++ b/src/data/ast.ts
@@ -1,4 +1,14 @@
-import { parseFirst, Statement, SelectFromStatement, astMapper, toSql, ExprRef } from 'pgsql-ast-parser';
+import { parseFirst, Statement, SelectFromStatement, astMapper, ExprRef } from 'pgsql-ast-parser';
+import { Lexer } from 'ch-parser/lexer';
+import { Token } from 'ch-parser/types';
+import {
+  FromQueryNode,
+  parseSelectQueryNode,
+  QueryNode,
+  QueryNodeParser,
+  QueryNodeType,
+  SelectQueryNode,
+} from 'ch-parser/parser';
 
 interface ReplacePart {
   startIndex: number;
@@ -123,26 +133,92 @@ export function sqlToStatement(rawSql: string): Statement {
   return mapper.statement(ast)!;
 }
 
+/**
+ * Tokenizes `sql` with the in-repo ClickHouse lexer and builds the shallow
+ * select-query node tree. Returns null when the input is not a SELECT (or
+ * WITH ... SELECT) statement.
+ */
+function parseSelect(sql: string): SelectQueryNode | null {
+  const lexer = new Lexer(sql);
+  const tokens: Token[] = [];
+  for (let token = lexer.nextToken(); !token.isEnd(); token = lexer.nextToken()) {
+    if (token.isSignificant()) {
+      tokens.push(token);
+    }
+  }
+  return parseSelectQueryNode(new QueryNodeParser(tokens));
+}
+
+// Strip backtick / double-quote identifier quoting so the name matches the
+// physical table name `additional_table_filters` keys on. `apply()` already
+// strips double quotes from the query before matching (see adHocFilter.ts).
+function unquoteIdentifier(name: string): string {
+  return name.replace(/^["`]|["`]$/g, '');
+}
+
+function qualifiedTableName(node: FromQueryNode): string | undefined {
+  if (!node.table) {
+    return undefined;
+  }
+  const table = unquoteIdentifier(node.table);
+  return node.database ? `${unquoteIdentifier(node.database)}.${table}` : table;
+}
+
+function isFromNode(node: QueryNode): node is FromQueryNode {
+  return node.type === QueryNodeType.From;
+}
+
+function isSelectNode(node: QueryNode): node is SelectQueryNode {
+  return node.type === QueryNodeType.Select;
+}
+
+/**
+ * Walks the select-query node tree in document order and returns the first
+ * physical table referenced by a FROM / JOIN clause, descending into
+ * subqueries and CTE bodies. We resolve to the underlying physical table
+ * rather than a subquery/CTE alias because that is what
+ * `additional_table_filters` keys on. Returns undefined when no physical
+ * table is present.
+ */
+function firstPhysicalTable(node: SelectQueryNode): string | undefined {
+  if (!node.children) {
+    return undefined;
+  }
+  for (const child of node.children) {
+    if (isFromNode(child)) {
+      const name = qualifiedTableName(child);
+      if (name) {
+        return name;
+      }
+    } else if (isSelectNode(child)) {
+      const nested = firstPhysicalTable(child);
+      if (nested) {
+        return nested;
+      }
+    }
+  }
+  return undefined;
+}
+
+/**
+ * Returns the physical table an ad-hoc filter should target for `sql`, or ''
+ * when none can be found.
+ *
+ * Previously this parsed with `pgsql-ast-parser`, a Postgres grammar that
+ * throws on valid ClickHouse syntax (SAMPLE, INTERVAL, lambdas, an existing
+ * SETTINGS clause, ...). On a throw it returned '', so the ad-hoc filter was
+ * silently dropped (grafana/clickhouse-datasource#958, #714). This uses the
+ * in-repo ClickHouse lexer and parser already shipping for query
+ * autocomplete, so ClickHouse syntax tokenizes instead of failing. The
+ * contract is unchanged: the first FROM table (descending into a leading
+ * subquery or CTE), preserving its original casing, and '' when none exists.
+ */
 export function getTable(sql: string): string {
-  const stm = sqlToStatement(sql);
-  if (stm.type !== 'select' || !stm.from?.length || stm.from?.length <= 0) {
+  const root = parseSelect(sql);
+  if (!root) {
     return '';
   }
-  switch (stm.from![0].type) {
-    case 'table': {
-      const table = stm.from![0];
-      const tableName = `${table.name.schema ? `${table.name.schema}.` : ''}${table.name.name}`;
-      // clickhouse table names are case-sensitive and pgsql parser removes casing,
-      // so we need to get the casing from the raw sql
-      const s = new RegExp(`\\b${tableName}\\b`, 'gi').exec(sql);
-      return s ? s[0] : tableName;
-    }
-    case 'statement': {
-      const table = stm.from![0];
-      return getTable(toSql.statement(table.statement));
-    }
-  }
-  return '';
+  return firstPhysicalTable(root) ?? '';
 }
 
 export function getFields(sql: string): string[] {
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -25,7 +25,7 @@ permissions: {}
 jobs:
   cd:
     name: CD
-    uses: grafana/plugin-ci-workflows/.github/workflows/cd.yml@ci-cd-workflows/v9.0.0
+    uses: grafana/plugin-ci-workflows/.github/workflows/cd.yml@ci-cd-workflows/v10.1.0
     permissions:
       contents: write
       id-token: write
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -24,7 +24,7 @@ jobs:
   ci:
     name: CI
 
-    uses: grafana/plugin-ci-workflows/.github/workflows/ci.yml@ci-cd-workflows/v9.0.0
+    uses: grafana/plugin-ci-workflows/.github/workflows/ci.yml@ci-cd-workflows/v10.1.0
 
     # Only run CI job for PRs / non-main refs.
     # Main publishing is handled by the CD workflow below.
--- a/cspell.config.json
+++ b/cspell.config.json
@@ -54,6 +54,8 @@
     "jsonencode",
     "listbox",
     "lowcardinality",
+    "macropro",
+    "datepart",
     "magefile",
     "multiquery",
     "mgbench",
--- a/docs/sources/configure.md
+++ b/docs/sources/configure.md
@@ -174,7 +174,7 @@ Use **Configuration mode** to choose how the data source is used by the query bu
 | **All databases** | Allows queries against any database and table the ClickHouse user can read. Use this mode for general exploration and dashboards that query multiple tables. |
 | **Single source** | Focuses the data source on one logs or traces table. Choose a **Signal type**, then configure the database, table, and column mappings for that source. |
 
-Use **Single source** when the data source is dedicated to one logs or traces table, such as an OpenTelemetry table. This keeps the logs or traces schema settings with the selected source and avoids reconfiguring column mappings in each query.
+Use **Single source** when the data source is dedicated to one logs or traces table, such as an OpenTelemetry table. This keeps the logs or traces schema settings with the selected source and avoids reconfiguring column mappings in each query. Single source data sources use the [compact query mode](/docs/plugins/grafana-clickhouse-datasource/<CLICKHOUSE_PLUGIN_VERSION>/query-editor/#compact-query-mode) in the query editor.
 
 ### Additional settings
 
--- a/docs/sources/query-editor.md
+++ b/docs/sources/query-editor.md
@@ -48,6 +48,18 @@ The query editor appears in [Explore](https://grafana.com/docs/grafana/latest/vi
 
 If the data source is configured in **Single source** mode, the query builder uses the configured logs or traces source as its default database, table, and column mapping.
 
+### Compact query mode
+
+When a data source is configured in **Single source** mode for logs or traces, the query builder uses a compact editor focused on the configured source. The compact editor omits database, table, and query type controls because those values come from the data source configuration.
+
+For logs, use the search field to filter log message text. Use **Add filter** to add column filters, and use the advanced options to adjust sorting and the result limit.
+
+Compact filters use Grafana's selected time range automatically. The configured time fields are not shown in the filter field list because the query builder manages the time range filter for you.
+
+Filter operators are tailored to the selected field type. Numeric fields offer comparison operators such as `>`, `<`, `>=`, `<=`, `=`, and `!=`. String fields offer text matching and equality operators such as **contains**, **does not contain**, `=`, and `!=`.
+
+The compact editor also includes a generated SQL preview. Use **Copy** to copy the generated query, or **Open in SQL editor** to switch to the SQL editor for manual customization.
+
 ## Build queries
 
 You can build queries using the **SQL editor** (raw SQL) or the **Query builder**. Queries can include macros for dynamic parts such as time range filters.
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/ClickHouse/clickhouse-go/v2 v2.45.0
 	github.com/docker/go-units v0.5.0
 	github.com/grafana/grafana-plugin-sdk-go v0.292.1
+	github.com/grafana/macropro v1.0.0
 	github.com/grafana/sqlds/v5 v5.1.1
 	github.com/moby/moby/api v1.54.2
 	github.com/paulmach/orb v0.13.0
--- a/go.sum
+++ b/go.sum
@@ -128,6 +128,8 @@ github.com/grafana/dataplane/sdata v0.0.9 h1:AGL1LZnCUG4MnQtnWpBPbQ8ZpptaZs14w6k
 github.com/grafana/dataplane/sdata v0.0.9/go.mod h1:Jvs5ddpGmn6vcxT7tCTWAZ1mgi4sbcdFt9utQx5uMAU=
 github.com/grafana/grafana-plugin-sdk-go v0.292.1 h1:8wvKUqIOtbHC43WIr6kheIGdV6q4SMyWU9+jxcUA6mE=
 github.com/grafana/grafana-plugin-sdk-go v0.292.1/go.mod h1:RM/Ku+hoyicIO/UVsGeFJfed/3/iXaCT3opiwE69THY=
+github.com/grafana/macropro v1.0.0 h1:R4/CgSgRrNQXpqk1JCHolg1/8Fz5g5CVW/C86AMz9Zw=
+github.com/grafana/macropro v1.0.0/go.mod h1:ID2n/OCKz4v4sqj+ckSNT7S9aRabod7G6Eg5S+sS4LU=
 github.com/grafana/otel-profiling-go v0.5.1 h1:stVPKAFZSa7eGiqbYuG25VcqYksR6iWvF3YH66t4qL8=
 github.com/grafana/otel-profiling-go v0.5.1/go.mod h1:ftN/t5A/4gQI19/8MoWurBEtC6gFw8Dns1sJZ9W4Tls=
 github.com/grafana/pyroscope-go/godeltaprof v0.1.9 h1:c1Us8i6eSmkW+Ez05d3co8kasnuOY813tbMN8i/a3Og=
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -14,3 +14,11 @@ const mockIntersectionObserver = jest.fn().mockImplementation((arg) => ({
 }));
 
 global.IntersectionObserver = mockIntersectionObserver;
+
+Object.defineProperty(HTMLCanvasElement.prototype, 'getContext', {
+  configurable: true,
+  writable: true,
+  value: jest.fn(() => ({
+    measureText: (text = '') => ({ width: String(text).length * 8 }),
+  })),
+});
--- a/pkg/converters/converters.go
+++ b/pkg/converters/converters.go
@@ -541,7 +541,7 @@ func defaultConvert(in interface{}) (interface{}, error) {
 	}
 
 	// handle pointers and dereference if possible
-	if val.Kind() == reflect.Ptr {
+	if val.Kind() == reflect.Pointer {
 		if val.IsNil() {
 			return nil, errors.New("nil pointer cannot be dereferenced in defaultConvert")
 		}
--- a/pkg/macros/macros.go
+++ b/pkg/macros/macros.go
@@ -6,128 +6,223 @@ import (
 	"strings"
 	"time"
 
-	"github.com/grafana/sqlds/v5"
-
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
 	"github.com/grafana/grafana-plugin-sdk-go/data/sqlutil"
+	"github.com/grafana/macropro"
 )
 
-// Converts a time.Time to a Date
+// badArgsErr wraps a bad-argument-count error so that downstream callers
+// (sqlds/grafana) classify it as ErrorSourceDownstream rather than a plugin
+// bug. macropro's MacroFunc signature forbids a backend.ErrorWithSource
+// return type, so we wrap via backend.DownstreamError and let errors.As
+// unwrap it at the sqlds boundary.
+func badArgsErr(macro string, want, got int) error {
+	return backend.DownstreamError(fmt.Errorf("%w: %s expected %d argument(s), received %d", sqlutil.ErrorBadArgumentCount, macro, want, got))
+}
+
+// timeToDate converts a time.Time to a ClickHouse Date literal.
 func timeToDate(t time.Time) string {
 	return fmt.Sprintf("toDate('%s')", t.Format("2006-01-02"))
 }
 
-// Converts a time.Time to a UTC DateTime with seconds precision
+// timeToDateTime converts a time.Time to a ClickHouse DateTime (seconds precision).
 func timeToDateTime(t time.Time) string {
 	return fmt.Sprintf("toDateTime(%d)", t.Unix())
 }
 
-// Converts a time.Time to a UTC DateTime64 with milliseconds precision
+// timeToDateTime64 converts a time.Time to a ClickHouse DateTime64 (milliseconds precision).
 func timeToDateTime64(t time.Time) string {
 	return fmt.Sprintf("fromUnixTimestamp64Milli(%d)", t.UnixMilli())
 }
 
-// FromTimeFilter returns a time filter expression based on grafana's timepicker's "from" time in seconds
-func FromTimeFilter(query *sqlutil.Query, args []string) (string, error) {
-	return timeToDateTime(query.TimeRange.From), nil
-}
-
-// ToTimeFilter returns a time filter expression based on grafana's timepicker's "to" time in seconds
-func ToTimeFilter(query *sqlutil.Query, args []string) (string, error) {
-	return timeToDateTime(query.TimeRange.To), nil
-}
-
-// FromTimeFilterMs returns a time filter expression based on grafana's timepicker's "from" time in milliseconds
-func FromTimeFilterMs(query *sqlutil.Query, args []string) (string, error) {
-	return timeToDateTime64(query.TimeRange.From), nil
-}
-
-// ToTimeFilterMs returns a time filter expression based on grafana's timepicker's "to" time in milliseconds
-func ToTimeFilterMs(query *sqlutil.Query, args []string) (string, error) {
-	return timeToDateTime64(query.TimeRange.To), nil
-}
-
-func TimeFilter(query *sqlutil.Query, args []string) (string, error) {
-	if len(args) != 1 {
-		return "", backend.DownstreamError(fmt.Errorf("%w: expected 1 argument, received %d", sqlutil.ErrorBadArgumentCount, len(args)))
+// contextFrom builds a macropro QueryContext from a sqlds Query.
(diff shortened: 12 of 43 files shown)


SpencerTorres (approved):
We may need a more elaborate fix in the future that resolves the table schema for better column matching, but this is still a good improvement. Thanks!

src/data/ast.ts
```diff
@@ -123,26 +133,92 @@ export function sqlToStatement(rawSql: string): Statement {
...
+    if (isFromNode(child)) {
+      const name = qualifiedTableName(child);
+      if (name) {
+        return name;
+      }
+    } else if (isSelectNode(child)) {
+      const nested = firstPhysicalTable(child);
+      if (nested) {
+        return nested;
+      }
```
adamyeats: `firstPhysicalTable()` walks in document order, so a CTE or scalar subquery appearing before the outer FROM wins: `WITH lookup AS (SELECT id FROM dim_services) SELECT * FROM otel_logs JOIN lookup ...` returns `dim_services`, and `SELECT (SELECT max(ts) FROM other_table) AS m, col FROM main_table` returns `other_table`, a regression (the old implementation returned `main_table`). The parser already tracks the real outer FROM in `SelectQueryNode.from` (`src/ch-parser/parser.ts:121-123`), so prefer that before descending into nested selects.

src/data/ast.ts
```diff
@@ -123,26 +133,92 @@ export function sqlToStatement(rawSql: string): Statement {
...
+ * rather than a subquery/CTE alias because that is what
+ * `additional_table_filters` keys on. Returns undefined when no physical
+ * table is present.
+ */
+function firstPhysicalTable(node: SelectQueryNode): string | undefined {
+  if (!node.children) {
+    return undefined;
+  }
+  for (const child of node.children) {
+    if (isFromNode(child)) {
```
adamyeats: The `ch-parser` emits a From node for SQL-standard function syntax too, so `SELECT EXTRACT(YEAR FROM Timestamp) AS y, count() FROM otel_logs` resolves to `Timestamp` (the old parser correctly returned `otel_logs`) and `trim(BOTH ' ' FROM col)` resolves to `col`. The filter is then keyed on a nonexistent table and ClickHouse silently ignores it.

src/data/ast.ts
```diff
@@ -123,26 +133,92 @@ export function sqlToStatement(rawSql: string): Statement {
...
+ * autocomplete, so ClickHouse syntax tokenizes instead of failing. The
+ * contract is unchanged: the first FROM table (descending into a leading
+ * subquery or CTE), preserving its original casing, and '' when none exists.
+ */
+export function getTable(sql: string): string {
+  const root = parseSelect(sql);
+  if (!root) {
+    return '';
+  }
+  return firstPhysicalTable(root) ?? '';
```
adamyeats, line 184: `getTable('SELECT * FROM ${table}')` now returns `''` (previously `'${table}'`), so `setTargetTableFromQuery()` in `src/data/adHocFilter.ts:16` throws `Failed to get table from adhoc query.` for `$clickhouse_adhoc_query` sources that reference a dashboard variable in FROM, a shape that worked before this PR.

src/data/ast.ts
```diff
@@ -123,26 +133,92 @@ export function sqlToStatement(rawSql: string): Statement {
...
+ * throws on valid ClickHouse syntax (SAMPLE, INTERVAL, lambdas, an existing
+ * SETTINGS clause, ...). On a throw it returned '', so the ad-hoc filter was
+ * silently dropped (grafana/clickhouse-datasource#958, #714). This uses the
+ * in-repo ClickHouse lexer and parser already shipping for query
+ * autocomplete, so ClickHouse syntax tokenizes instead of failing. The
+ * contract is unchanged: the first FROM table (descending into a leading
+ * subquery or CTE), preserving its original casing, and '' when none exists.
+ */
+export function getTable(sql: string): string {
+  const root = parseSelect(sql);
```
adamyeats, line 180: The `FROM` handler rejects barewords in the ch-parser keyword set, so tables named like keywords now return `''`: `SELECT * FROM default.values` and `SELECT * FROM sample` resolved to `default.values` / `sample` before this PR and now silently drop the filter (or throw in the tag-fetch path).

src/data/ast.ts
```diff
@@ -123,26 +133,92 @@ export function sqlToStatement(rawSql: string): Statement {
...
-      return getTable(toSql.statement(table.statement));
+  }
+  return parseSelectQueryNode(new QueryNodeParser(tokens));
+}
+
+// Strip backtick / double-quote identifier quoting so the name matches the
+// physical table name `additional_table_filters` keys on. `apply()` already
+// strips double quotes from the query before matching (see adHocFilter.ts).
+function unquoteIdentifier(name: string): string {
+  return name.replace(/^["`]|["`]$/g, '');
```
adamyeats, line 128: Stripping backticks here strands the new capability at `apply()`: the guard at `src/data/adHocFilter.ts:78` only strips double quotes, so ``FROM `my-db`.`my-table` `` resolves a table but the settings clause is never appended. That is the same silent-drop class this PR fixes for double quotes, and a name containing regex metacharacters (e.g. `` `a[b` ``) also throws an uncaught `SyntaxError` when interpolated into `new RegExp` there.

src/data/ast.ts
```diff
@@ -123,26 +133,92 @@ export function sqlToStatement(rawSql: string): Statement {
...
+ * `additional_table_filters` keys on. Returns undefined when no physical
+ * table is present.
+ */
+function firstPhysicalTable(node: SelectQueryNode): string | undefined {
+  if (!node.children) {
+    return undefined;
+  }
+  for (const child of node.children) {
+    if (isFromNode(child)) {
+      const name = qualifiedTableName(child);
```
adamyeats: Table functions are reported as physical tables: `SELECT * FROM merge('default', '^otel')` returns `merge` (previously `''` plus a loud error), so the query ships with `additional_table_filters={'merge': ...}` that ClickHouse silently ignores. The user sees unfiltered data that looks filtered.

src/data/ast.ts
```diff
@@ -123,26 +133,92 @@ export function sqlToStatement(rawSql: string): Statement {
...
-      return s ? s[0] : tableName;
+/**
+ * Tokenizes `sql` with the in-repo ClickHouse lexer and builds the shallow
+ * select-query node tree. Returns null when the input is not a SELECT (or
+ * WITH ... SELECT) statement.
+ */
+function parseSelect(sql: string): SelectQueryNode | null {
+  const lexer = new Lexer(sql);
+  const tokens: Token[] = [];
+  for (let token = lexer.nextToken(); !token.isEnd(); token = lexer.nextToken()) {
```
adamyeats: Semicolons are consumed as ordinary tokens, so the stream crosses statement boundaries: `SELECT 1; SELECT * FROM second_table` returns `second_table` (previously `''`), keying the filter on a statement the panel query didn't target.

src/data/ast.ts
```diff
@@ -123,26 +133,92 @@ export function sqlToStatement(rawSql: string): Statement {
...
-      // clickhouse table names are case-sensitive and pgsql parser removes casing,
-      // so we need to get the casing from the raw sql
-      const s = new RegExp(`\\b${tableName}\\b`, 'gi').exec(sql);
-      return s ? s[0] : tableName;
+/**
+ * Tokenizes `sql` with the in-repo ClickHouse lexer and builds the shallow
+ * select-query node tree. Returns null when the input is not a SELECT (or
+ * WITH ... SELECT) statement.
+ */
+function parseSelect(sql: string): SelectQueryNode | null {
```
adamyeats: This is a third hand-rolled copy of the lex, filter significant tokens, `QueryNodeParser` pipeline (`src/components/suggestions.ts:118-134` has the same loop). Consider exporting a single `parseSelect(sql)` from `src/ch-parser` and reusing it in both places so autocomplete and ad-hoc table detection cannot diverge.

adamyeats (changes requested):

alex-fedotyev:
Thanks for the detailed review. I reworked the table resolution around your feedback, and it changed the approach enough that I force-pushed a single clean commit (b39c40f).

The core change: instead of a separate document-order walk over the tokens, `getTable()` now resolves through `SelectQueryNode.from` (the outer FROM the parser already tracks), and I corrected the parser so `node.from` is right for the shapes you flagged. `parseSelect()` is exported from `src/ch-parser` and shared with autocomplete, so the two paths cannot diverge.

Point by point:

1. Document-order walk (CTE / scalar subquery): now uses `node.from` with descent into a leading subquery or CTE. `WITH lookup AS (...) SELECT ... FROM otel_logs JOIN lookup` resolves `otel_logs`; `SELECT (SELECT max(ts) FROM other) ... FROM main` resolves `main`.
2. SQL-standard `FROM` inside functions: `node.from` is set only at paren depth 0, so `EXTRACT(YEAR FROM Timestamp) ... FROM otel_logs` resolves `otel_logs` and `trim(BOTH ' ' FROM col) FROM otel_logs` resolves `otel_logs`.
3. `${var}` in FROM: captured again, so `SELECT * FROM ${table}` resolves `${table}`.
4. Keyword-named tables: accepted in the FROM position, so `default.values` and `sample` resolve.
5. Table functions: flagged and resolved to no table, so `merge('default', '^otel')` (and combined with `SAMPLE`) resolves to `''` instead of a key ClickHouse ignores.
6. Semicolons: a semicolon ends the statement, so `SELECT 1; SELECT * FROM second_table` resolves `''` (the first statement has no table).
7. Backtick stranding and regex metacharacters at `apply()`: `apply()` now strips backticks as well as double quotes and regex-escapes the target table, so backtick-quoted names get the settings clause and metacharacter names no longer throw.
8. Shared pipeline: `parseSelect(sql)` is exported and reused by both `getSuggestions` and `getTable`.

Each point has a test (the `getTable` matrix plus no-regression cases in `ast.test.ts`, `node.from` in `parser.test.ts`, the `apply()` cases in `adHocFilter.test.ts`, and an autocomplete case for keyword-named tables). Verified end to end in Grafana 13.2.1 against the official 4.21.2 build; before/after and the emitted SQL are in the description. Ready for another look when you have time.

src/data/ast.ts
adamyeats: The outer-FROM preference fixes both shapes I quoted, and `SELECT (SELECT max(ts) FROM other_table) AS m, col FROM main_table` returns `main_table` again. The descent still runs in document order whenever `node.from` yields no table, so a subquery FROM re-opens the same hole:

| query | main | this PR |
|---|---|---|
| `SELECT (SELECT max(ts) FROM other_table) AS m FROM (SELECT * FROM main_table) x` | `main_table` | `other_table` |
| `WITH x AS (SELECT a FROM cte_table) SELECT * FROM (SELECT * FROM inner_t) y` | `''` | `cte_table` |
| `WITH a AS (SELECT 1 FROM t_a), b AS (SELECT 2 FROM t_b) SELECT * FROM (SELECT * FROM real_t) x` | `''` | `t_a` |

The first row is a regression. `apply()` emits `additional_table_filters={'other_table' : ...}`, so the panel's own table stays unfiltered while an unrelated table is narrowed.

`FROM <cte-alias>` also resolves to the alias. `WITH lookup AS (SELECT id FROM dim_services) SELECT * FROM lookup` returns `lookup`, and ClickHouse keys `additional_table_filters` on tables it reads, so that entry matches nothing.

Consider descending only into the select child that belongs to `node.from`, rather than the first select child in document order. Resolving a FROM target that matches a CTE name in the same level's WITH list would cover the alias case.

src/data/ast.ts
adamyeats: `SELECT * FROM ${table}` resolves again, and `setTargetTableFromQuery` no longer throws for it. The qualified forms still fail, and one of them fails the same way the unqualified form used to:

| query | main | this PR |
|---|---|---|
| `SELECT * FROM mydb.${table}` | `mydb.${table}` | `''`, then `setTargetTableFromQuery` throws |
| `SELECT * FROM ${db}.mytable` | `${db}.mytable` | `${db}` |

`SELECT * FROM ${db}.${table}` also yields `${db}`. The loop at `parser.ts:147-159` breaks on the first `ClosingCurlyBrace` and never checks for a following `Dot`, so the table half is dropped.

The loop is unbounded when no closing brace follows, which is the state on every keystroke while someone types a variable into FROM. It consumes the rest of the token stream, so no clause node after FROM is created. Measured with a stub schema, `getSuggestions('SELECT * FROM ${table WHERE ', ...)`:

- main: 19 suggestions, macros first
- this PR: 2 suggestions, `['default', 'logs']`

Autocomplete offers database and table names instead of columns and macros for the rest of the query. Bounding the loop at a clause keyword, a `Semicolon` or a `ClosingRoundBracket` would cover this and the nested case, where `SELECT * FROM (SELECT * FROM ${t) WHERE x=1` currently swallows the subquery's closing bracket.

src/data/ast.ts
adamyeats: `SELECT * FROM merge('default', '^otel')` returns `''` now, and the same holds for `cluster`, `remote`, `numbers`, `s3`, `url` and `file`. Whitespace before the bracket is handled too.

The descent added alongside it re-introduces the symptom for a narrower shape. When the outer FROM is a table function, `firstPhysicalTable` falls through to the children and picks up an unrelated table:

| query | main | this PR |
|---|---|---|
| `SELECT * FROM merge('default','^otel') WHERE id IN (SELECT id FROM inner_t)` | `''` | `inner_t` |
| `WITH lookup AS (SELECT id FROM dim_services) SELECT * FROM merge('default','^otel')` | `''` | `dim_services` |
| `SELECT * FROM numbers(10) WHERE id IN (SELECT id FROM users)` | `''` | `users` |

This lands worse than the original finding rather than equivalent. `merge` was a key ClickHouse ignored, so the filter was dropped. `inner_t` is a real table, so ClickHouse honours the clause and narrows a table the user did not target.

Guarding the descent on `node.from` being absent or a subquery, rather than on it yielding no name, would keep the table-function case at `''`.

src/data/adHocFilter.ts
```diff
@@ -74,8 +74,14 @@ export class AdHocFilter {
...
 
-    // sql can contain a query with double quotes around the database and table name, e.g. "default"."table", so we remove those
-    if (this._targetTable !== '' && !sql.replace(/"/g, '').match(new RegExp(`.*\\b${this._targetTable}\\b.*`, 'gi'))) {
+    // sql can contain a query that quotes the database and table name, e.g.
+    // "default"."table" or `default`.`table`, so we remove the quoting before
+    // matching. The table name is regex-escaped so a name containing regex
+    // metacharacters (e.g. a backtick-quoted `a[b`) neither throws nor mismatches.
+    if (
+      this._targetTable !== '' &&
+      !sql.replace(/["`]/g, '').match(new RegExp(`.*\\b${escapeRegExp(this._targetTable)}\\b.*`, 'gi'))
```
adamyeats, line 83: `_targetTable` is resolved once and then cached for the life of the instance (`adHocFilter.ts:88-90`), and `CHDatasource` holds a single `AdHocFilter` (`CHDatasource.ts:243`). `setTargetTableFromQuery` only runs on the tag-source path, so on a normal dashboard the first panel that resolves a table wins, and this guard rejects every later panel on a different table.

The caching predates this PR. Widening `getTable` widens the blast radius to the shape the query builder itself emits:

```
panel1 = SELECT toStartOfInterval("ts", INTERVAL 1 second) AS "time", count() FROM "default"."tbl_a" GROUP BY time
panel2 = SELECT count() FROM "default"."tbl_b"
```

Run through one `AdHocFilter`, in that order:

- main: `getTable(panel1)` is `''`, nothing is cached, and panel2 is filtered
- this PR: panel1 is filtered, `_targetTable` caches `default.tbl_a`, and panel2 comes back unchanged

On a multi-table dashboard the first panel to render decides which table gets ad-hoc filters for the session. Every other panel shows unfiltered data with no error.

Resolving the table per `apply()` call would fix it, and the tag-source path can keep its explicit target. It might be worth doing that here rather than in a follow-up, because this PR is what makes time-series panels reach the cache.

src/data/ast.ts
```diff
@@ -123,26 +124,64 @@ export function sqlToStatement(rawSql: string): Statement {
...
+ * `parseSelect`), not `pgsql-ast-parser`. The Postgres grammar throws on valid
+ * ClickHouse syntax (SAMPLE, INTERVAL, lambdas, an existing SETTINGS clause),
+ * and on a throw the old implementation returned '', so the ad-hoc filter was
+ * silently dropped (grafana/clickhouse-datasource#958). The ClickHouse parser
+ * tokenizes that syntax instead of failing, and resolves through subqueries,
+ * CTEs, keyword-named tables, and Grafana variables while rejecting table
+ * functions and honoring statement boundaries.
+ */
+export function getTable(sql: string): string {
+  const root = parseSelect(sql);
```
adamyeats, line 180: Queries that used to return `''` here now resolve, which makes three existing `apply()` hazards reachable for the first time. Each one predates this PR, and each one was out of reach for ClickHouse-only syntax because the Postgres grammar rejected it.

`sql.replace(';', '')` at `adHocFilter.ts:102` has no `/g` and no position check, so it deletes the first semicolon anywhere in the query:

```
in:  SELECT splitByChar(';', col) FROM default.tbl WHERE ts > now() - INTERVAL 1 DAY
out: SELECT splitByChar('', col) FROM default.tbl WHERE ts > now() - INTERVAL 1 DAY settings additional_table_filters={...}
```

ClickHouse rejects `splitByChar('', ...)`, so that panel fails. Anchoring the replace as `/;\s*$/` would keep literals intact.

A trailing line comment swallows the appended clause. I checked on a server, and `SELECT 1 -- c SETTINGS max_block_size=100` returns `1`, so the setting never applies. For a query ending in `-- TODO revisit`, the emitted SQL puts `settings additional_table_filters={...}` inside that comment. The panel runs and looks healthy while the filter is inert. Appending after a newline would avoid it.

For `SELECT a FROM default.tbl_a UNION ALL SELECT a FROM default.tbl_b`, only the first branch is keyed, so half the result set obeys the filter.

These are arguably out of scope. I am flagging them because this PR is what exposes them, and a half-applied filter reads as a worse failure than the old visible no-op.

src/ch-parser/parser.ts
```diff
@@ -86,6 +90,13 @@ export interface SelectQueryNode extends QueryNode {
   from?: FromQueryNode;
 }
 
+// A table/database name in a FROM clause. Unlike the general identifier
+// positions, keywords are accepted here because ClickHouse allows keyword-named
+// tables (e.g. `default.values`, `sample`).
+function isTableNameToken(token: Token): boolean {
```
adamyeats, line 96: Dropping the `!isKeyword()` guard fixes `default.values` and `sample`, and it also lets a clause keyword be taken as the table name when the name is missing. `SELECT * FROM  WHERE ` parses with `from.table` set to `WHERE`, and no `Where` clause node is created.

That state occurs while someone deletes a table name to retype it. With a stub schema:

- main: 19 suggestions, macros first
- this PR: 2 suggestions, `['default', 'logs']`

`getTable('SELECT * FROM WHERE Timestamp > now()')` also returns `WHERE`, where main returned `''`.

Keyword acceptance is only needed for a real table name. Refusing the clause-starting keywords (`WHERE`, `GROUP`, `ORDER`, `HAVING`, `LIMIT`, `SETTINGS`, `PREWHERE`, `JOIN`, `FORMAT`) in that position would keep both behaviours.

src/data/ast.test.ts
```diff
@@ -43,4 +43,98 @@ describe('ast', () => {
...
+      ['lambda', 'SELECT arrayMap(x -> x + 1, arr) AS a FROM otel_logs', 'otel_logs'],
+      ['qualified name + SAMPLE', 'SELECT * FROM db.otel_logs SAMPLE 1/10 WHERE x > 1', 'db.otel_logs'],
+      ['subquery with ClickHouse syntax', 'SELECT * FROM (SELECT * FROM physical SAMPLE 0.1) sub', 'physical'],
+      ['backtick-quoted name', 'SELECT * FROM `my-db`.`my-table`', 'my-db.my-table'],
+      [
+        'CTE + JOIN',
+        'WITH lookup AS (SELECT id FROM dim_services) SELECT * FROM otel_logs JOIN lookup ON otel_logs.id = lookup.id',
+        'otel_logs',
+      ],
+      ['existing SETTINGS clause', 'SELECT * FROM otel_logs SETTINGS max_threads = 1', 'otel_logs'],
```
adamyeats, line 62: nit: this row passes with the whole fix reverted. `replaceMacroFunctions` truncates the SQL at the first `settings` keyword before `parseFirst` runs, so the pre-PR `getTable` already returned `otel_logs` here. I checked both spacings.

It sits under a comment saying the previous parser threw on this syntax, which does not hold for this row. Moving it to the no-regression block below would keep the block comment accurate. The same claim appears in the PR body and the commit message, so those are worth a look too.

src/data/ast.test.ts
```diff
@@ -43,4 +43,98 @@ describe('ast', () => {
...
+        'the outer table, not a scalar subquery in the SELECT list',
+        'SELECT (SELECT max(ts) FROM other_table) AS m, col FROM main_table',
+        'main_table',
+      ],
+      [
+        'the outer table, not the EXTRACT operand',
+        'SELECT EXTRACT(YEAR FROM Timestamp) AS y FROM otel_logs',
+        'otel_logs',
+      ],
+      ['the outer table, not the trim operand', "SELECT trim(BOTH ' ' FROM col) FROM otel_logs", 'otel_logs'],
```
adamyeats, line 84: nit: the pre-PR `getTable` returns `''` for this input rather than `otel_logs`, so this row belongs in the first block. The block comment says the expected values match the previous parser's output, which holds for the other eight rows.

src/ch-parser/parser.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  it('captures a Grafana variable in the FROM position', () => {
+    expect(parseSelect('SELECT * FROM ${table}')?.from?.table).toBe('${table}');
+  });
+
+  it('stops at a statement boundary so a later statement does not set node.from', () => {
+    expect(parseSelect('SELECT 1; SELECT * FROM second_table')?.from).toBeUndefined();
+  });
+
+  it('leaves node.from unset for a leading subquery (descent is the caller step)', () => {
```
adamyeats, line 48: nit: `node.from` is set for this input, and only `node.from.table` is undefined, so the title reads as stronger than the assertion. The optional chaining also passes if `parseSelect` ever returns `null`, and the same applies to the statement-boundary test above. Adding `expect(node).not.toBeNull()` would close that off.

adamyeats (changes requested):
Two mechanical items before merge, both outside the diff.

The PR title and the commit subject differ:

- title: `fix: detect ad-hoc filter table with the ClickHouse parser`
- commit: `fix(adhoc): resolve ad-hoc filter table with the ClickHouse parser`

The repo squash-merges with `squash_title=PR_TITLE` and `squash_message=COMMIT_MESSAGES`, so the squash commit carries both lines and both parse as Conventional Commits. release-please then writes two entries under 🐛 Bug Fixes for one commit. Making the two strings identical avoids a manual CHANGELOG fix afterwards. `fix(adhoc): ...` reads better, as the scope matches the area.

nit: the `changelog` label is not needed. No workflow reads it since release-please was adopted, and the `fix` type already makes the change changelog-visible. The instruction in `pull_request_template.md` is stale and worth a separate cleanup.

alex-fedotyev:
Thanks, this was a thorough pass and every case reproduced. Reworked in `1fc6bf8`.

Table resolution now goes through `node.from`, and when that FROM is a subquery it descends into that subquery and only that one, so a scalar subquery in the SELECT list or a sibling CTE body can no longer win.

- **Descent picking an unrelated table.** Fixed. `SELECT (SELECT max(ts) FROM other_table) AS m, col FROM (SELECT * FROM main_table) x` resolves `main_table`, and the CTE-plus-subquery shapes resolve the outer FROM's subquery.
- **Table-function descent.** Fixed. A table-function FROM resolves to `''` and its argument subqueries are not followed, so `merge(...) WHERE id IN (SELECT id FROM inner_t)` and `numbers(10) WHERE id IN (...)` stay `''`.
- **Qualified variables / unbounded loop.** Fixed. `${db}.mytable`, `mydb.${table}` and `${db}.${table}` all round-trip, and the variable capture stops at a clause keyword, a semicolon or a bracket, so an unterminated `${` while typing no longer swallows the rest of the query.
- **Clause keyword as table.** Fixed. `WHERE`, `GROUP`, `ORDER`, `HAVING`, `LIMIT`, `SETTINGS`, `PREWHERE`, `JOIN`, `FORMAT` and `UNION` are refused in the table position, so `FROM  WHERE ...` leaves the table unset and keeps the `WHERE` clause node.
- **`_targetTable` cached across panels.** Fixed here rather than deferred, since this change is what makes the query-builder's time-series shape reach the cache. `apply()` now resolves the table per call and no longer writes it back to the shared instance; the tag-source path keeps its explicit target.
- **Trailing semicolon.** Fixed. The strip is anchored to `/;\s*$/`, so `splitByChar(';', col)` is left intact.
- **Test rows in the wrong block, and the same claim in the body and commit.** Fixed. The SETTINGS row moved to the no-regression block, the `trim(... FROM ...)` row moved to the block for shapes the Postgres grammar rejected, and the body and commit message no longer claim SETTINGS threw.
- **Optional chaining hiding a null node.** Fixed. The `node.from`-is-undefined tests now assert the node is non-null first.

The trailing-line-comment case, keying every UNION branch, and resolving a `FROM` that references a CTE are pre-existing behaviors that are no-ops today (the CTE case) or the known multi-table-application limitation (UNION). I split them into #2172 to keep this diff focused.

On the mechanical items: the title now matches the commit subject, and I dropped the `changelog` label line from the description (noted the stale template instruction in #2172).

Verified again in Grafana 13.2.1 against the official 4.21.2 build; details in the description.

src/data/ast.ts
adamyeats: The qualified forms all resolve now, and `${db}.${table}` is better than main, which leaked a random replacement token into the key. The bounded capture also does what it says for clause detection: `SELECT * FROM ${table WHERE ` gets a `Where` node and macro suggestions again.

One new regression in the same code. Typing a variable as the FROM target no longer offers variable completions. With `table` and `database` defined, cursor at the end:

| input | main | this PR |
|---|---|---|
| `SELECT * FROM ${` | 21 items incl. `${table}`, `${database}` | 2 items: `default`, `otel_logs` |
| `SELECT * FROM ${ta` | 20 items incl. `${table}` | 2 items, same |
| `SELECT * FROM default.${ta` | 20 items incl. `${table}` | 1 item: `otel_logs` |
| `SELECT * FROM otel_logs WHERE ${` | 22 incl. variables | 22 incl. variables |

On main the `$` and `{` tokens were `Identifier` children, so the cursor mapped to the column context, which calls `getVariableSuggestions`. Now `readVariableSegment` makes `${ta` the table, the clause stays `From`, and that context never asks for variables. The shipped OTel service dashboard uses `FROM ${database}.otel_traces`, so this is the authoring flow the PR sets out to support.

The smallest fix is in `suggestions.ts`: when the FROM target text starts with `$`, add the variable suggestions to the database/table list. Alternatively the parser could leave `from.table` unset for an unterminated `${` and record the prefix instead.

nit: the capture only stops at `;`, `(`, `)` and the reserved words, so `${table AS t WHERE x=1` yields `${tableASt` and `${table FINAL SAMPLE 0.1` yields `${tableFINALSAMPLE0.1`. Mid-edit only, and the comment above the loop reads as a stronger bound than that.

src/ch-parser/parser.ts
adamyeats: `SELECT * FROM  WHERE ` gets 19 macro suggestions and no `DESC` again, and `getTable` returns `''` for it. That part is fixed.

The guard costs something main had. ClickHouse accepts every one of the ten words as an unquoted table name in FROM position. On 25.8 and 26.8, `CREATE TABLE order (x UInt8)` then `SELECT x FROM d1.order` returns the row, and the public 26.9 server reports `UNKNOWN_TABLE` rather than a syntax error for all ten. main resolved them:

| query | main | this PR |
|---|---|---|
| `SELECT * FROM default.order` (also `.where`, `.group`, `.limit`, `.format`, `.union`, `.join`, `.having`, `.prewhere`) | `default.order` | `''` |
| `SELECT * FROM format` (also `prewhere`) | `format` | `''` |
| `` SELECT * FROM `order` `` | `''` | `order` |

So a dashboard on a table literally named `order` or `format` silently loses ad-hoc filtering, the #958 failure class. The segment after a dot can never start a clause, so the guard has no mid-edit value there. Consider applying `RESERVED_FROM_KEYWORDS` only to the first segment. That restores every `db.<word>` case. For the unqualified first segment the heuristic is a fair trade, and quoting is the escape hatch, which now works where it did not on main. It would be worth a line in the PR body as user-facing impact.

nit: the comment on line 97 says these keywords are never a table name in the FROM position, which the server contradicts, and `sample` is an unlucky example on line 114 because `SAMPLE` legally follows a table, so `SELECT * FROM  SAMPLE 0.1` mid-edit takes it as the table. The set also misses the join modifiers (`LEFT`, `INNER`, `ARRAY`, `ASOF`, `GLOBAL`...) and `FINAL`, which main rejected via `isKeyword()`. The visible list is unchanged for those, the cost is one failing `DESC TABLE "default"."LEFT"` per name per session.

src/data/ast.ts
```diff
@@ -123,26 +124,58 @@ export function sqlToStatement(rawSql: string): Statement {
...
+  }
+  const table = unquoteIdentifier(node.table);
+  return node.database ? `${unquoteIdentifier(node.database)}.${table}` : table;
+}
+
+// Resolves the table the outer FROM reads from at this select level. When the
+// FROM is a subquery it descends into that subquery, and only that subquery, so
+// a scalar subquery in the SELECT list or a sibling CTE body is never mistaken
+// for the target. A table-function FROM resolves to undefined.
+function firstPhysicalTable(node: SelectQueryNode): string | undefined {
```
adamyeats, line 147: The CTE-alias deferral holds for the panel path and not for the tag source. With `$clickhouse_adhoc_query` set to `WITH t AS (SELECT * FROM default.tbl_a) SELECT DISTINCT k FROM t`, then a panel `SELECT count() FROM default.otel_logs`:

- main: `setTargetTableFromQuery` threw `Failed to get table from adhoc query.`, `_targetTable` stayed empty, and the panel still resolved per query and was filtered
- this PR: no throw, `_targetTable` is pinned to `t`, and every panel whose SQL does not contain the word `t` comes back unfiltered

Keys load, the filter chips look active, nothing filters. #2172 item 1 says it is a no-op today, which is true only for the panel `apply()` path.

Two small options. Return `undefined` here when `from.table` matches a CTE name declared in the same statement's WITH list, which restores the throw. Or resolve through that CTE's body to `default.tbl_a`, which makes the filter work. Either is a handful of lines. If it stays deferred, the "exactly as it was before" wording in the test and the issue needs the tag-source caveat.

src/data/adHocFilter.ts
```diff
@@ -74,16 +74,22 @@ export class AdHocFilter {
...
-    if (this._targetTable === '') {
-      this._targetTable = getTable(sql);
-    }
-
-    if (this._targetTable === '') {
+    // sql can contain a query that quotes the database and table name, e.g.
+    // "default"."table" or `default`.`table`, so we remove the quoting before
+    // matching. The table name is regex-escaped so a name containing regex
+    // metacharacters (e.g. a backtick-quoted `a[b`) neither throws nor mismatches.
+    if (!sql.replace(/["`]/g, '').match(new RegExp(`.*\\b${escapeRegExp(targetTable)}\\b.*`, 'gi'))) {
```
adamyeats: When `targetTable` came from `getTable(sql)` two lines up, this guard re-checks the same string against a normalised spelling of a name that was just parsed out of it. It cannot produce a true positive there, only false negatives, and it does:

| query | main | this PR |
|---|---|---|
| `SELECT count() FROM "日志"` (also `` `логи` ``) | filtered on the first render, dropped on every refresh | never filtered |
| `SELECT count() FROM default . otel_logs` | same | never filtered |

JavaScript `\b` is ASCII-only, so a space before a Cyrillic or CJK letter is not a boundary, and the spaced dot does not match `default\.otel_logs`. main was already broken after the first render, so this is not much of a regression, but the per-call change removes the one render that worked.

Consider running the guard only when `_targetTable` was set explicitly by the tag-source path, where it is the thing that gives the explicit target its meaning. That fixes both rows above and the non-word-edge names (`-weird`, `tbl-`) without touching the regex.

src/data/ast.ts
```diff
@@ -123,26 +124,58 @@ export function sqlToStatement(rawSql: string): Statement {
...
+  return undefined;
+}
+
+/**
+ * Returns the physical table an ad-hoc filter should target for `sql`, or ''
+ * when none can be found.
+ *
+ * Parses with the in-repo ClickHouse parser (shared with SQL autocomplete via
+ * `parseSelect`), not `pgsql-ast-parser`. The Postgres grammar throws on valid
+ * ClickHouse syntax (SAMPLE, INTERVAL, lambdas, an existing SETTINGS clause),
```
adamyeats: nit: the docblock still lists "an existing SETTINGS clause" among the syntax the Postgres grammar threw on. The pre-PR `getTable` returned `otel_logs` for `SELECT * FROM otel_logs SETTINGS max_threads = 1` with no throw, because `replaceMacroFunctions` stripped the clause before `parseFirst`. The commit message and PR body were corrected to "backtick-quoted names" and this line was not.

Line 132 above `qualifiedTableName` says a Grafana-variable target resolves to `undefined`, but `getTable('SELECT * FROM ${table}')` returns `${table}` verbatim, which is what the docblock six lines below says. One of the two comments has to give.

src/data/ast.test.ts
```diff
@@ -43,4 +43,139 @@ describe('ast', () => {
...
+      ],
+      [
+        'the outer table, not the EXTRACT operand',
+        'SELECT EXTRACT(YEAR FROM Timestamp) AS y FROM otel_logs',
+        'otel_logs',
+      ],
+      ['a Grafana variable used as the table', 'SELECT * FROM ${table}', '${table}'],
+      ['a Grafana variable as the qualified table', 'SELECT * FROM my_db.${table}', 'my_db.${table}'],
+      ['a Grafana variable as the database', 'SELECT * FROM ${db}.my_table', '${db}.my_table'],
+      ['Grafana variables on both sides of the dot', 'SELECT * FROM ${db}.${table}', '${db}.${table}'],
```
adamyeats: nit: this row sits under "the expected values match the previous parser's output", but the pre-PR `getTable` returned `${db}.v<random>` for `SELECT * FROM ${db}.${table}`, a leaked replacement token that changed on every call. It is a fix, so it belongs in the first block.

Three neighbours have the same shape of problem. Line 58 `CTE + JOIN` sits under "the parser threw", but pgsql parsed it as statement type `with` and the old code returned `''` through the `type !== 'select'` check, without throwing. Line 95's label "stripped before parsing" describes the old pgsql path, not what `getTable` does now. Line 176's "exactly as it was before this change" holds for the panel path only, per the `firstPhysicalTable` comment.

src/data/adHocFilter.test.ts
```diff
@@ -186,6 +214,48 @@ describe('AdHocManager', () => {
...
+      { key: 'k', operator: '=', value: 'v' },
+    ] as AdHocVariableFilter[]);
+    const p2 = ahm.apply('SELECT count() FROM default.tbl_b', [
+      { key: 'k', operator: '=', value: 'v' },
+    ] as AdHocVariableFilter[]);
+    expect(p1).toContain("additional_table_filters={'default.tbl_a'");
+    expect(p2).toContain("additional_table_filters={'default.tbl_b'");
+  });
+
+  it('strips only a trailing semicolon, not one inside a string literal', () => {
```
adamyeats: nit: the SQL here has no trailing semicolon, so the "strips only a trailing semicolon" half of the title is unasserted. `'SELECT * FROM default.tbl;'` alongside it would cover both halves.

Line 89's regex-metachar test asserts only `not.toThrow`. With `getTable` stubbed to a wrong non-empty name it still passes while `apply()` returns the SQL unchanged. Head produces `additional_table_filters={'a[b' : ...}` for it, so asserting that clause, as the backtick test next to it does, would pin the behaviour the comment in `apply()` claims.

src/ch-parser/parser.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    expect(node).not.toBeNull();
+    expect(node?.from?.table).toBeUndefined();
+    expect(node?.from?.subquery?.from?.table).toBe('inner_t');
+  });
+
+  it('returns null for a non-select statement', () => {
+    expect(parseSelect('INSERT INTO foo VALUES (1)')).toBeNull();
+  });
+
+  it('does not take a clause keyword as the table when the table is missing', () => {
```
adamyeats, line 61: nit: same optional-chain shape as the two tests renamed above. Both assertions pass if `parseSelect` returns `null`, so the test would survive the parser rejecting these mid-edit inputs entirely, which is what the guard exists to prevent. `expect(node).not.toBeNull()` plus a check that a `Where` clause child exists would pin it.

adamyeats (changes requested):
Thanks for the thorough revision. Seven threads are resolved, the title now matches the commit subject, and the rebase, CI and local gates are all clean. What is left is two regressions the fixes introduced, each a few lines to correct, one deferral whose wording needs a caveat, and nits. Details are inline.

nit: the semicolon boundary I asked for has an autocomplete side effect. With more than one statement in the editor, the cursor in the second one gets FROM-position suggestions where main gave column context. ClickHouse rejects multi-statement queries, so this is scratch editing only. Fine to leave as a known trade-off.

nit: three wording points in #2172. Item 1's "no-op today" holds for the panel path only, as the inline comment on `firstPhysicalTable` shows. Item 2's first-branch keying is new at head, since main applied nothing to a UNION. Item 3's newline fix still leaves a stray `;` for a query ending in `; -- comment`, so the strip wants to cover that too.

nit: the body's "the only behavior difference is that queries the Postgres grammar rejected now resolve a table" undersells the change. WITH and UNION statements that the grammar accepted also changed, because the old `type !== 'select'` check rejected them, and `apply()` now resolves per call. Worth listing, and the `db.<keyword>` impact from the reserved-keyword thread belongs there as user-facing.

alex-fedotyev:
Thanks again, all of these reproduced. Fixed in `06a8cc0`. To stop chasing these one round at a time, I built a differential harness (run `getTable` and `getSuggestions` over an enumerated corpus on `main` and on this branch, diff, and classify every delta), so I could confirm the whole main-to-HEAD delta rather than the lines quoted. It caught one regression I introduced mid-fix (below), which is exactly the point.

- **Descent into a table-function's argument subquery / greedy descent.** Fixed. `merge(...) WHERE id IN (SELECT id FROM inner_t)` and `numbers(10) WHERE id IN (...)` stay `''`; descent only follows the FROM's own subquery.
- **CTE-alias pinning the tag-source target.** Fixed by resolving rather than deferring: a `FROM <cte>` now resolves to the CTE body's table (`WITH lookup AS (SELECT id FROM dim_services) SELECT * FROM lookup` resolves `dim_services`), so `setTargetTableFromQuery` no longer pins an alias. The CTE work and the settings-after-newline fix from the follow-up are folded into this PR; #2172 is rescoped to per-branch UNION keying.
- **Reserved-keyword guard too broad.** Fixed. The guard applies only to the unqualified first segment, so `default.order` (and `.where`, `.format`, ...) resolve again. An unqualified keyword table now needs backtick quoting (`` `order` `` resolves); noted as user-facing in the description.
- **`${` losing variable completions.** Fixed. The variable capture is now adjacency-bounded (it keeps a keyword variable name like `${table}` and stops an unterminated `${` at whitespace or a bracket), and `getSuggestions` offers variables when the FROM target starts with `$`. `${` and `default.${` list `${table}` / `${database}` again.
- **`apply()` guard false negatives.** Fixed. The presence re-check runs only for an explicit tag-source target; a target parsed from the query is not re-checked, so non-ASCII and spaced-dot names filter.
- **Trailing semicolon inside a literal.** The strip stays anchored to a trailing `;`; the settings clause now appends on a new line so a trailing `-- comment` cannot swallow it.
- **UNION.** A top-level UNION now resolves to no table, matching main, rather than keying only the first branch. Per-branch keying is the remaining #2172 item.
- **Test-label and doc nits.** The SETTINGS and `${db}.${table}` rows moved to the correct blocks, the "threw" block comment now says "threw or resolved to ''", the SETTINGS docblock line and the `qualifiedTableName` comment are corrected, and the optional-chain tests now assert `not.toBeNull()` plus a clause child.

On the multi-statement autocomplete side effect of the semicolon boundary: agreed it is scratch editing only (ClickHouse rejects multi-statement), leaving it as the known trade-off.

Description updated with the user-facing behavior changes (WITH/UNION statements the grammar previously rejected, `db.<keyword>`, unqualified keyword tables needing quotes). Re-verified in Grafana 13.2.1.

alex-fedotyev:
@adamyeats - please take a look, I think everything is addressed now and just rebased it as well.

adamyeats (approved):
Approving to land the #958 fix. Every review thread is resolved and verified against main: subquery and CTE descent, table functions, per-call table resolution, the first-segment keyword rule, variable completions in FROM, and the guard and semicolon handling in `apply()`.

Three small items remain and go to #2172 as follow-ups rather than another round:

- a `;` followed by a trailing comment now produces `Multi-statements are not allowed` when a filter is active, where main ran the query
- unqualified `format` and `prewhere` tables no longer resolve (the two of the ten reserved words pgsql did not reserve)
- typing `${` as a JOIN target offers no variable completions, since `typingVariable` reads `root.from`

Thanks for the thorough work across the rounds.

PR #1937, opened 2026-06-30
fix(traces): prevent crash when running Traces queries on older Grafana

### LINKED ISSUE #2034: Prevent crash when running Traces queries on older Grafana
Traces queries crashed the plugin on older Grafana versions that lack newer SDK capabilities. This adds version-appropriate handling so traces degrade gracefully instead of crashing.

Created retroactively to record effort attribution for https://github.com/grafana/clickhouse-datasource/pull/1937.

adamyeats:
## Type of Change

- [x] 🐛 Bug Fix

---

### What is the bug?

Running a **Traces** query in Explore crashes with `RangeError: Maximum call stack size exceeded` and blanks the results panel. It reproduces on Grafana versions below 13.x and works on 13.x.

Root cause: trace results attach **View trace** / **View logs** data links whose `internal.query` embedded the **live `Datasource` instance**. That instance became circular (`datasource.variables.datasource === datasource`) once the guided variable editor wired up `CustomVariableSupport`. Grafana's Explore link scanner (`getStringsFromObject` in `explore/utils/links.ts`) walks the internal query recursively with no cycle guard, so on Grafana without the upstream guard it recurses until the stack overflows.

### How to reproduce

1. On Grafana < 13 (e.g. 12.2), open Explore with the ClickHouse data source.
2. Switch **Query Type** to **Traces** and run a query against a table containing trace data.
3. The results panel is blank and the browser console shows `RangeError: Maximum call stack size exceeded`.

- **Grafana version**: reproduces below 13.x (verified on 12.2.0); plugin floor is `>=11.6.0`
- **Plugin version**: 4.18.0 / `main`

---

## The fix

Embed a plain `{ uid, type }` datasource ref via `getDataSourceRef(datasource)` in the link's internal query instead of the live instance. `datasourceUid` / `datasourceName` are already passed separately, so the links behave identically. Fixing this plugin-side covers every Grafana version rather than relying on the newer-Grafana guard.

---

## Please check that:

- **Unit** (`src/data/utils.test.ts`): asserts the link's internal query carries a serializable `{ uid, type }` ref (not the live instance) and is JSON-serializable. It fails on the pre-fix code with the circular `variables.datasource` reference.
- **E2E** (`tests/e2e/traceQueryNoCrash.spec.ts`): loads a Traces query in Explore and fails on any uncaught/fatal client error — matching the stack-overflow class **and** the React "maximum update depth" render-loop class — so future trace-render regressions are caught too, not just this one. Its value is the version matrix: the reusable Playwright CI runs down to the `>=11.6` floor where this crash manifests (a latest-only run would miss it).
- Verified locally on Grafana 12.2: the E2E fails on the pre-fix bundle (`console.error: RangeError: Maximum call stack size exceeded`) and passes after the fix; full unit suite green (69 suites / 855 tests).

Closes #2034

### DIFF
--- a/cspell.config.json
+++ b/cspell.config.json
@@ -72,6 +72,7 @@
     "OFREP",
     "otel",
     "otelcol",
+    "pageerror",
     "semconv",
     "OUTFILE",
     "paulmach",
--- a/src/data/utils.test.ts
+++ b/src/data/utils.test.ts
@@ -465,6 +465,39 @@ describe('transformQueryResponseWithTraceAndLogLinks', () => {
     expect(out.data[0].fields[0].config.links ?? []).toHaveLength(0);
   });
 
+  it('embeds a serializable datasource ref (not the live instance) in trace/log link queries', async () => {
+    // The live Datasource instance contains a circular reference (datasource.variables.datasource
+    // === datasource, added with CustomVariableSupport). Embedding the instance in a data link's
+    // internal query makes Grafana's link scanner (getStringsFromObject) recurse infinitely on
+    // older Grafana, throwing "Maximum call stack size exceeded". The link must carry a plain
+    // { uid, type } ref instead.
+    const mockDatasource = newMockDatasource();
+    configureOtelLogs(mockDatasource);
+    configureOtelTraces(mockDatasource);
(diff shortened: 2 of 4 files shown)


zoltanbedi (approved):
lgtm

PR #1987, opened 2026-07-01
feat(auth): add Forward OAuth Identity authentication

benjamin-awd:
## Type of Change

- [x] 🚀 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧹 Refactor / Chore

---

### What is this feature?

Adds an opt-in **Forward OAuth Identity** option to the data source configuration. When enabled, the plugin forwards the logged-in Grafana user's OAuth access token to ClickHouse as a JWT, using clickhouse-go's native `GetJWT` callback, instead of authenticating with the configured username and password.

When the toggle is on:
- Static username/password are suppressed on per-query connections, so the JWT is the sole credential.
- Health checks (e.g. Save & test) and alert rule evaluation fall back to the configured username/password, since no user token is available outside a query request.
- Header forwarding and per-user connection pooling are enabled automatically (`ForwardHeaders` is OR'd with `oauthPassThru`).
- A secure (TLS) connection is required; the connection is rejected otherwise.

Today the plugin authenticates every query with static credentials, so all queries reach ClickHouse as a single shared account. This makes it impossible to attribute queries to the actual Grafana user or to apply ClickHouse-side access rights, row policies, and quotas per user. `Forward Grafana HTTP headers` carries user *metadata* (e.g. `X-Grafana-User`) but does not establish the user as an authenticated principal. Forwarding the user's OAuth token as a JWT closes that gap.

### Who is this feature for?

Operators connecting Grafana to **ClickHouse Cloud** (where JWT authentication is available) who want per-user identity end-to-end — for query-log attribution, row policies, or quotas keyed on the real user rather than a shared service account.

### How to test this feature

Automated:
1. Backend: `go test ./pkg/plugin/...` (see `TestResolveJWTAuth`, `TestBuildClickHouseOptionsJWT*`, `TestSettingsForwardHeadersWithJWT`, `TestLoadSettingsOAuthPassThru`).
2. Frontend: `npm run test` (see the `oauthPassThru` cases in `CHConfigEditor` and `DatabaseCredentialsSection`).

Manual (requires ClickHouse Cloud with JWT auth configured and Grafana set up to forward the user's OAuth token):
1. Add a ClickHouse data source, enable **Secure connection**, and set a username/password (used for health checks).
2. Enable **Forward OAuth Identity** under **Database credentials**, then **Save & test** (succeeds via the fallback credentials).
3. Run a query from a dashboard and confirm in `system.query_log` that it is attributed to the logged-in user rather than the service account.
4. Disable **Secure connection** with the toggle on and confirm the connection is rejected with a TLS-required error.

---

## Screenshots / Videos

**With Forward OAuth Identity**
image

**Query result with JWT auth**
Screenshot 2026-07-01 at 1 34 58 PM

---

## Please check that:

- **Availability:** JWT authentication is currently [ClickHouse Cloud only](https://clickhouse.com/docs/en/operations/external-authenticators/jwt) server-side. This is called out in the UI tooltip and the docs.
- **Refactor:** `Connect()` is split to extract `buildClickHouseOptions()` so option assembly and JWT/header resolution can be unit-tested in isolation. This accounts for most of the non-additive diff in `driver.go`.
- **No behavior change when the toggle is off** — existing username/password and header-forwarding paths are untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features
 
-- Add Forward OAuth Identity authentication: forward the signed-in Grafana user's OAuth token to ClickHouse Cloud as a JWT so queries are attributed to the real user instead of a shared service account. Requires a verified TLS connection. Alert and other backend queries have no signed-in user and are blocked by default; enable **Allow service account fallback** to let them run with the configured username/password
+- Add Forward OAuth Identity authentication: forward the signed-in Grafana user's OAuth token to ClickHouse Cloud as a JWT so queries are attributed to the real user instead of a shared service account. Requires a verified TLS connection. Alert and other backend queries have no signed-in user and are blocked by default; enable **Allow service account fallback** to let them run with the configured username/password (#1987)
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Add Forward OAuth Identity authentication: forward the signed-in Grafana user's OAuth token to ClickHouse Cloud as a JWT so queries are attributed to the real user instead of a shared service account. Requires a verified TLS connection. Alert and other backend queries have no signed-in user and are blocked by default; enable **Allow service account fallback** to let them run with the configured username/password
+
 ### Fixes
 
 - Apply the log message search to the logs volume and logs sample queries, so the volume histogram matches the filtered log list (#2092)
--- a/docs/sources/configure.md
+++ b/docs/sources/configure.md
@@ -298,7 +298,7 @@ To enable it, turn on **Forward OAuth Identity** in the **Database credentials**
 
 ### Requirements and behavior
 
-- **A secure (TLS) connection is required.** Enable **Secure connection** and set the **Port** to a TLS-enabled port. The plugin rejects the connection if JWT authentication is enabled without TLS.
+- **A secure (TLS) connection is required.** Enable **Secure connection** and set the **Port** to a TLS-enabled port. The plugin rejects the connection if JWT authentication is enabled without TLS. It also rejects the connection when **Skip TLS Verify** is enabled, because forwarding a real user's token over an unverified connection exposes it to interception.
 - **Credentials are suppressed on query connections.** When enabled, the configured username and password are not sent on per-query connections; the forwarded token is the sole credential.
 - **Health checks fall back to username and password.** **Save & test** and other health checks run outside a user request, where no user token is available, so they use the configured username and password. Keep valid credentials configured so connection tests can succeed.
 - **Alerting is blocked by default.** Alert rule evaluation also runs outside a user session, so there is no identity to forward. By default these queries are **rejected** rather than run as a shared account. To keep alerting working, enable **Allow service account fallback** in the **Database credentials** section.
--- a/docs/sources/configure.md
+++ b/docs/sources/configure.md
@@ -130,6 +130,7 @@ After adding the data source, configure the following settings.
 | **Secure connection** | Enable when your ClickHouse server uses TLS. When enabled, update the **Port** to a TLS-enabled port and configure [TLS settings](#tls-settings) below. |
 | **Username**          | ClickHouse user name. Use a [read-only user](#clickhouse-user-and-permissions).                                                                         |
 | **Password**          | ClickHouse user password.                                                                                                                               |
+| **Forward OAuth Identity** | Forward the logged-in Grafana user's OAuth token to ClickHouse as a JWT instead of authenticating with the configured username and password. ClickHouse Cloud only; requires a [secure (TLS) connection](#tls-settings). See [JWT authentication](#jwt-authentication). |
 | **Default database**  | The database the query builder uses when no database is selected. If left blank, the plugin defaults to `default`.                                      |
 | **Default table**     | The default table used by the query builder.                                                                                                            |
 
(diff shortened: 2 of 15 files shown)


ktw4071 (commented):
## Review: align this with the "Forward OAuth Identity" feature

Thanks for this — the refactor into `buildClickHouseOptions`/`resolveJWTAuth` is clean and the backend test coverage of the token-swap is solid. A few things below, framed around the fact that this is the ClickHouse instance of the same Grafana feature (`oauthPassThru` / **Forward OAuth Identity**, mirroring grafana core #124617/#124618) that was just landed for Google Cloud Monitoring in [grafana-cloudmonitoring-datasource#84](https://github.com/grafana/grafana-cloudmonitoring-datasource/pull/84). Worth keeping the two consistent where it matters.

### 1. Naming — call it "Forward OAuth Identity", not "JWT authentication"

From Grafana's perspective this is the **Forward OAuth Identity** auth method (`oauthPassThru`). JWT is only *how ClickHouse ingests the forwarded token* (via `GetJWT`), not the auth type. Calling it "JWT authentication" reads like configuring a static JWT credential, which this isn't — and the PR's own UI label already says **Forward OAuth Identity**, so the title contradicts the code.

Suggest: `feat(auth): add Forward OAuth Identity authentication`, and mention JWT in the body as the ClickHouse-side delivery mechanism. This also matches the sibling cloudmonitoring PR and the docs.

### 2. Alerting runs silently as the service account (please confirm this is intended)

The cloudmonitoring PR **blocks** alerting under Forward OAuth Identity — backend returns a `DownstreamError` and the query editor shows an explanatory alert — because there's no signed-in user to attribute the query to.

This PR instead **silently falls back** to the configured username/password for alert evaluation. The whole point of the feature is per-user identity (row policies, quotas, query-log attribution), so silently evaluating alerts as the shared service account means alerts run with *broader* privileges than the user would have. That's a surprising and security-relevant divergence.

It's fine to choose fallback over fail-fast, but I'd like it to be a deliberate, documented decision — and ideally we mirror cloudmonitoring's block-or-warn so operators aren't caught out. At minimum, call out the privilege implication in the docs.

### 3. Making username optional under JWT contradicts the fallback design

The v2 editor relaxes username to `required={!jsonData.oauthPassThru}` and skips the required-username validation when the toggle is on. But health checks (**Save & test**) *and* alert evaluation depend on those static credentials — the PR body itself says "keep valid credentials configured." So a user can enable the toggle, leave username blank, pass validation, and then hit a failing **Save & test**.

The cloudmonitoring PR went the other way and *added* validation. Suggest keeping username required (or showing an inline "still needed for health checks/alerting" note) rather than relaxing it.

### 4. TLS check allows `InsecureSkipVerify`

The guard `settings.OAuthPassThru && tlsConfig == nil` ensures a TLS config exists, but `tlsConfig` is still non-nil when **Skip TLS Verify** is on. That means a user-identity bearer token can be forwarded over a connection whose server cert isn't verified (MITM-capturable). Consider rejecting — or at least prominently warning — when `OAuthPassThru && InsecureSkipVerify`. Forwarding a real user's token is a higher bar than a shared service credential.

### 5. Missing unit test for the TLS-required rejection

`buildClickHouseOptions` was extracted specifically to be testable, but the `"JWT authentication requires a secure (TLS) connection"` path is only manually verified. A `Secure: false, OAuthPassThru: true` case asserting the error is a two-line add and guards a security-relevant invariant.

### 6. Parity nits (lower priority)

- **Friendly auth errors:** cloudmonitoring surfaces guidance on 401 (session expired → sign out/in) and 403 (missing scope/role). A ClickHouse JWT rejection currently surfaces as a raw driver error — a friendlier message would help.
- **CHANGELOG:** cloudmonitoring updated its changelog; this PR doesn't touch one (if the repo tracks it).
- **`.(bool)` type assertion** in `LoadSettings` panics on a non-string/non-bool value. It matches the existing pattern in the file, so not a blocker, but noting it.

---

Net: the core design and per-user connection isolation check out (sqlds keys the cached connection by the forwarded headers, including the token, so there's no cross-user reuse). The main asks before merge are **(1) the rename**, **(2)/(3) the alerting + required-username decisions** so behavior matches the documented intent and the sibling plugin, and **(4)/(5) the TLS hardening + test**.

ktw4071 (changes requested):

benjamin-awd:
Hey @ktw4071, thanks for your review! I've made changes following the numbers above:

1. Merged existing commits into a single commit `feat(auth): add Forward OAuth Identity authentication` bc980b1

2. Alerting is now managed by a `Allow service account fallback` toggle so that queries with a JWT are always blocked by default. df11ea7

3. Reverted the `required={!oauthPassThru}` relaxation; username is always required (backs health checks + fallback). Regression test added. f3802e7

4. Decision: reject hard. `OAuthPassThru && InsecureSkipVerify` → DownstreamError. Docs + test. 911df27

5. Added a test to assert a non-secure OAuthPassThru config is rejected `07dc71e`

6. Resolved nits: Friendly auth hints (401/403 + native codes 516/497/164) via `authErrorHint`; CHANGELOG entry. `.(bool)` assertion left as-is. abf2980

AndreZiviani:
hi, I suggest updating `clickhouse-go` to version `v2.48.0` due to this bug
https://github.com/ClickHouse/clickhouse-go/pull/1914

benjamin-awd:
Hey @AndreZiviani, thanks for the heads up -- this PR fortunately doesn't go through that path since it uses `Options.GetJWT` rather than the per-query `WithJWT`. I'm running a fork of this at the moment and haven't observed the regression

I think it makes sense to bump to 2.48.0, but probably as a separate PR

ktw4071 (dismissed):
Thanks for the thorough turnaround here — this addresses almost everything from the last pass, and a couple of points went further than I asked.

Looks great:

- Alerting (my main concern): blocking tokenless data queries by default, with the opt-in oauthPassThruAllowFallback toggle + a backend warn log when fallback is exercised + the docs warning admonition, is a clean resolution. Keying on token-presence rather than a FromAlert header is arguably better — it also covers non-OAuth users, not just alerts. 👍
- Username stays required (with the regression test), Skip TLS Verify now rejected, the TLS-required test, the 401/403 authErrorHint, and the CHANGELOG entry — all resolved.
- One thing left over — the rename is only half done. The labels, changelog, and commit messages now say Forward OAuth Identity, but the two most visible spots still say "JWT authentication":

Minor changes

- The PR title (feat(auth): add JWT authentication > feat(auth): add Forward OAuth Identity authentication) — GitHub doesn't pick this up from commit messages, so it needs a manual edit.
- Section heading ## JWT authentication → ## Forward OAuth Identity (renaming the heading changes the anchor from #jwt-authentication to #forward-oauth-identity).
- The config-table entry currently ends See [JWT authentication](#jwt-authentication). → update the link target to (#forward-oauth-identity) so it doesn't break.

benjamin-awd:
Hey @ktw4071, updated the PR and updated the headings 54b20e0

ktw4071 (approved):
LGTM!

PR #1995, opened 2026-07-02
fix(config): restore inline required-field validation for host/port on default installs

### LINKED ISSUE #2028: Restore inline required-field validation for host and port on default installs
The config editor stopped showing inline required-field validation for host and port on fresh installs. This restores the validation so misconfiguration is caught before save and test.

Created retroactively to record effort attribution for https://github.com/grafana/clickhouse-datasource/pull/1995.

adamyeats:
## Type of Change

- [x] 🐛 Bug Fix

---

### What is the bug?

Adding config validation (#1769) changed the required Server address / Port `Field`s from a structural check to feature-toggle-gated `fieldErrors`:

```diff
- invalid={!jsonData.host}   error={labels.serverAddress.error}
+ invalid={!!fieldErrors.host}   error={fieldErrors.host}
```

`fieldErrors` comes from `validation?.getErrors()`, and `validation` is only defined when the **Grafana-core** `clickHouseConfigValidation` feature toggle is enabled. That toggle is not shipped or defaulted-on by this plugin, so on the **default install** (across Grafana 11.6–12.x) `validation` is `undefined`, `fieldErrors` is always `{}`, and the inline "required" indicator for an empty Server address / Port — present in v4.17.0 — no longer appears. The field looks valid until the backend **Save & Test** round-trip fails with a less specific error. (CI/e2e didn't catch this because the dev/e2e environment force-enables the toggle.)

This is a UX regression only — the field still functions and Save & Test still blocks a broken config — but it removes the immediate inline feedback users had before.

### How to reproduce

1. Default Grafana install (no `clickHouseConfigValidation` feature toggle).
2. Open the ClickHouse datasource config, leave **Server address** (or **Port**) empty.
3. **Before:** no inline required error (only the static `*`). **After:** the field shows the inline "Server address required" / "Port is required" error immediately, as in v4.17.0.

### Related Issues

Closes #

- **ClickHouse version**: n/a (config UI)
- **Grafana version**: 11.6–12.x default install (no `clickHouseConfigValidation` toggle)
- **Plugin version**: 4.18.0

---

## Please check that:

The fix falls back to the structural empty check (`!jsonData.host` / `!jsonData.port`) only when `validation` is unavailable; when the toggle **is** on, `fieldErrors` remains authoritative and behaviour is unchanged. Two tests were added (empty host shows the error; filled host hides it) — these run with the toggle off, which is the default in the test env.

Closes #2028

### DIFF
--- a/src/views/CHConfigEditor.test.tsx
+++ b/src/views/CHConfigEditor.test.tsx
@@ -3,7 +3,7 @@ import { fireEvent, render, screen } from '@testing-library/react';
 import { ConfigEditor } from './CHConfigEditor';
 import { mockConfigEditorProps } from '__mocks__/ConfigEditor';
 import '@testing-library/jest-dom';
-import { Protocol } from 'types/config';
+import { CHConfig, Protocol } from 'types/config';
 import allLabels from 'labels';
 
 jest.mock('@grafana/runtime', () => {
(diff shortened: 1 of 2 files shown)


alyssajoyner (approved):
Great catch 💯

PR #2002, opened 2026-07-04
fix(converters): handle NaN/Inf in Tuple, Map and Nested JSON conversion

### LINKED ISSUE #1049: Unable to render Tuple(Float64) if holding NaN or Inf
**What happened**:

When querying dataset containing column of Float64/32 type nested in a Tuple, the query fails in case the value is NaN/Inf with the error:

```
json: unsupported value: NaN: Could not process SQL results
json: unsupported value: +Inf: Could not process SQL results
```

**How to reproduce it**:

Reproducible query within the datasource:
```
SELECT tuple(NaN)
SELECT tuple(Inf)
SELECT tuple(-Inf)
```

It works as expected when not in a tuple:
```
SELECT NaN, Inf, -Inf
```

Tarasusrus:
## What this PR does

Fixes #1049.

`Tuple`, `Map` and `Nested` columns are converted to JSON via `json.Marshal` in `jsonConverter`. Go's `encoding/json` rejects `NaN` and `±Inf` floats, so a query such as:

```sql
SELECT tuple(NaN)
SELECT tuple(Inf)
SELECT tuple(-Inf)
```

failed with:

```
json: unsupported value: NaN: Could not process SQL results
```

even though the same values work fine as top-level columns (`SELECT NaN, Inf, -Inf`).

As @SpencerTorres noted in the issue, the tuple is marshaled as JSON and fails to convert in the Go backend.

## Approach

When `json.Marshal` fails, retry after replacing every `NaN`/`±Inf` float (at any nesting depth) with `null` via a new `sanitizeJSONFloats` helper. Key points:

- The retry only runs **on failure**, so the common (finite-value) path is completely unchanged — no added overhead.
- Sanitizing walks maps, slices/arrays, pointers and interfaces recursively.
- Byte slices are left untouched so they keep marshaling as base64 strings rather than being exploded into arrays.

`null` matches how JSON represents "no value", and is the natural encoding for a float that has no JSON representation.

## Tests

Added `TestTupleWithNaNAndInf` covering: unnamed tuple with `NaN`, named tuple with `+Inf`/`-Inf`, nested array of floats with `NaN`, map value with `NaN`, pointer-to-`NaN`, byte-slice preservation, and a finite-value regression check. `go test ./pkg/converters/... ./pkg/plugin/...` passes.

### DIFF
--- a/cspell.config.json
+++ b/cspell.config.json
@@ -19,6 +19,7 @@
     "ASOF",
     "apikey",
     "buildinfo",
+    "chcol",
     "CHSQL",
     "ClickHouse",
     "clickhouseexporter",
--- a/pkg/converters/values.go
+++ b/pkg/converters/values.go
@@ -1,13 +1,17 @@
 package converters
 
 import (
+	"encoding"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math"
 	"math/big"
 	"net"
 	"reflect"
+	"strconv"
 
+	"github.com/ClickHouse/clickhouse-go/v2/lib/chcol"
 	"github.com/paulmach/orb"
 	"github.com/shopspring/decimal"
 )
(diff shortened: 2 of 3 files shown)


pkg/converters/converters.go
```diff
@@ -895,13 +896,65 @@ func jsonConverter(in any) (any, error) {
...
+		}
+		return out
+	case reflect.Map:
+		out := make(map[string]any, rv.Len())
+		iter := rv.MapRange()
+		for iter.Next() {
+			out[fmt.Sprint(iter.Key().Interface())] = sanitizeJSONFloats(iter.Value().Interface())
+		}
+		return out
+	default:
```
adamyeats: Variant, Dynamic and JSON columns also route through `jsonConverter`, but they scan into `chcol.Variant`/`chcol.JSON` structs that land in this default case and come back unchanged, so a NaN inside them still fails on the retry. It might be worth handling those types (or recursing into structs generally) so they get the same fix, for example `SELECT CAST(nan AS Variant(Float64))` still reproduces #1049.

pkg/converters/converters.go
```diff
@@ -895,13 +896,65 @@ func jsonConverter(in any) (any, error) {
...
+		out := make([]any, rv.Len())
+		for i := 0; i < rv.Len(); i++ {
+			out[i] = sanitizeJSONFloats(rv.Index(i).Interface())
+		}
+		return out
+	case reflect.Map:
+		out := make(map[string]any, rv.Len())
+		iter := rv.MapRange()
+		for iter.Next() {
+			out[fmt.Sprint(iter.Key().Interface())] = sanitizeJSONFloats(iter.Value().Interface())
```
adamyeats: For non-string keys `fmt.Sprint` produces a different key format from `encoding/json`, so a `Map(DateTime, Float64)` row containing NaN renders keys as `2024-01-01 00:00:00 +0000 UTC` while clean rows in the same column render `2024-01-01T00:00:00Z`. Consider encoding keys the way `encoding/json` does (via `encoding.TextMarshaler` when implemented) so sanitised and clean rows stay consistent.

pkg/converters/converters.go
```diff
@@ -895,13 +896,65 @@ func jsonConverter(in any) (any, error) {
 	}
 
 	jBytes, err := json.Marshal(in)
+	if err != nil {
```
adamyeats: This retries on every marshal error, not just NaN/Inf, so unrelated failures pay a full recursive walk plus a second marshal, and a cyclic value would overflow the stack rather than surface the cycle error (the driver cannot produce cycles today, so this is hardening rather than a live bug). Consider gating the retry on a `*json.UnsupportedValueError` check with `errors.As` to keep it targeted.

adamyeats (changes requested):
Looks good so far @Tarasusrus! Tested locally and it seems to work for me. Some comments below.

Tarasusrus:
Thanks for the thorough review @adamyeats! Rebased onto latest `main` (the converters refactor) and addressed all three points in `pkg/converters/values.go`:

1. **Variant / Dynamic / JSON.** You're right — those scan into `chcol.Variant`/`chcol.JSON`, which marshal over unexported fields, so reflection never reached the floats and they fell through unchanged. `sanitizeJSONFloats` now handles them explicitly via their public accessors (`Variant.Any()`, `JSON.NestedMap()`), recursing into the value they actually marshal. `CAST(nan AS Variant(Float64))` now renders `null` (covered by a new test).

2. **Map keys.** Replaced `fmt.Sprint` with a `jsonMapKey` helper that mirrors `encoding/json`: string keys pass through, `encoding.TextMarshaler` keys (e.g. `time.Time`) use `MarshalText`, integers are formatted numerically. A `Map(DateTime, Float64)` now keeps RFC3339 keys consistent with clean rows — test added.

3. **Targeted retry.** The retry is now gated on `*json.UnsupportedValueError` via `errors.As`, so unrelated marshal failures surface immediately instead of paying a full recursive walk. Also added a depth bound so a pathological/cyclic value can't overflow the stack.

Added tests for nested slice/map, string- and time-keyed maps, Variant, JSON, and the non-retried error path. `go test ./pkg/converters/`, `go vet`, `gofmt` and `cspell` all pass locally.

adamyeats (approved):
Great stuff @Tarasusrus , thank you! LGTM!

PR #2008, opened 2026-07-08
refactor(macros): expand macros via the sqlds Interpolator extension point

adamyeats:
## Type of Change

- [x] 🧹 Refactor / Chore

## What is this change?

Follow-up to #1802. sqlds v5.2.0 added a pluggable `Interpolator` extension point (grafana/sqlds#269) that lets a plugin replace the default `sqlutil.Interpolate` pipeline. This PR bumps sqlds to v5.3.0 and installs a macropro-backed `Interpolator`, replacing the `MutateQueryData` pre-expansion workaround that #1802 introduced.

- `NewDatasource` now assigns `ds.Interpolator = interpolateMacros`, a small adapter that calls `macros.Interpolate` on the parsed query.
- The `expandMacros`, `expandMacrosInQuery` and `macroErrorQuery` plumbing in `driver.go` is removed. It only existed because `MutateQueryData` cannot return an error, so macro failures had to be rewritten into a `SELECT throwIf(...)` query that failed at execution time.

Two behaviour improvements fall out of this:

- Macro errors are returned directly on the query response, classified as downstream, without a database round trip. The user-visible message prefix changes from `macro expansion failed:` to sqlds's `Could not apply macros:`.
- `$__table` and `$__column` now expand correctly. The interpolator receives the fully parsed `sqlutil.Query` including `Table` and `Column`, which the old pre-expansion path never carried.

### How to test

1. `docker compose up -d clickhouse-server`
2. `go test -tags=integration ./pkg/plugin/ -run TestQueryDataMacroExpansion -v`

## Please check that:

- [x] Tests for this change have been added/updated.
- [ ] Documentation has been added/updated (where applicable).

## Special notes for your reviewer

The driver's empty `Macros()` map is retained. With a custom `Interpolator` installed sqlds never consults it on the query path, but it remains part of the `sqlds.Driver` interface.

Closes #2026

PR #2014, opened 2026-07-08
feat: add optional Row Capacity Hint setting for large query responses

### LINKED ISSUE #2025: Add optional Row Capacity Hint setting for large query responses
Large query responses caused repeated slice reallocation during row-to-frame conversion. This adds an opt-in Row Capacity Hint data source setting so frames can be pre-sized, building on the row capacity support in sqlds v5.3.0. The setting defaults to existing behaviour.

Created retroactively to record effort attribution for https://github.com/grafana/clickhouse-datasource/pull/2014.

adamyeats:
## Type of Change

- [x] 🚀 Feature

---

### What is this feature?

Adds an optional **Row Capacity Hint** setting to the ClickHouse datasource. It wires the sqlds `DriverSettings.RowCapacityHint` knob (grafana/sqlds#240, released in sqlds v5.3.0) through to the query path. sqlds uses the hint to pre-allocate each `data.Frame`'s fields to the expected row count before scanning rows.

Without a hint, sqlds grows each column's slice as rows arrive, which triggers repeated reallocation and copying on large responses. Supplying an expected row count lets the frame fields be allocated once, cutting allocation churn on the hot path for datasources that reliably return large results.

The value is static per datasource and applied to every query, so it is opt-in and defaults to `0` (disabled). Operators who know their workload can set it. A value larger than the typical result would waste memory, so the tooltip calls this out and the default keeps existing behaviour unchanged.

### Who is this feature for?

Operators running dashboards or alerting against ClickHouse datasources whose queries consistently return a large, similar number of rows and who want to reduce backend allocation overhead.

### How to test this feature

1. Open a ClickHouse datasource config page and expand **Query settings**.
2. Confirm a new numeric **Row Capacity Hint** field appears (default empty/0).
3. Set it to a value such as `50000`, save, and run a query. Behaviour is unchanged from the user's perspective; the hint only affects backend frame allocation.
4. Leave it unset to confirm the historical behaviour (no pre-allocation) still applies.

Backend and frontend unit tests cover the settings parsing (number, string, negative-clamped, absent) and the new config field.

---

## Notes

- Based on #2008, which bumps sqlds to v5.3.0 (the release containing grafana/sqlds#240). This PR targets `feat/sqlds-interpolator` and should merge after #2008.
- The field lives in the shared `QuerySettingsConfig` component, so it renders in both the current config editor and the v2 design.

## Please check that:

- [x] The change follows the existing settings-parsing and config-editor patterns.

Closes #2025

### DIFF
--- a/src/views/config-v2/AdditionalSettingsSection.tsx
+++ b/src/views/config-v2/AdditionalSettingsSection.tsx
@@ -189,6 +189,7 @@ export const AdditionalSettingsSection = (props: Props) => {
               maxIdleConns={jsonData.maxIdleConns}
               maxOpenConns={jsonData.maxOpenConns}
               queryTimeout={jsonData.queryTimeout}
+              rowCapacityHint={jsonData.rowCapacityHint}
               validateSql={jsonData.validateSql}
               enableMapKeysDiscovery={jsonData.enableMapKeysDiscovery}
               onDialTimeoutChange={(e) => {
@@ -198,6 +199,10 @@ export const AdditionalSettingsSection = (props: Props) => {
               onQueryTimeoutChange={(e) => {
                 trackClickhouseConfigV2QuerySettings({ queryTimeout: Number(e.currentTarget.value) });
                 onUpdateDatasourceJsonDataOption(props, 'queryTimeout')(e);
+              }}
+              onRowCapacityHintChange={(e) => {
+                trackClickhouseConfigV2QuerySettings({ rowCapacityHint: Number(e.currentTarget.value) });
+                onUpdateDatasourceJsonDataOption(props, 'rowCapacityHint')(e);
               }}
               onConnMaxLifetimeChange={(e) => {
                 trackClickhouseConfigV2QuerySettings({ connMaxLifetime: Number(e.currentTarget.value) });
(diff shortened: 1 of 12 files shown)


src/views/config-v2/AdditionalSettingsSection.tsx
```diff
@@ -189,6 +189,7 @@ export const AdditionalSettingsSection = (props: Props) => {
               maxIdleConns={jsonData.maxIdleConns}
               maxOpenConns={jsonData.maxOpenConns}
               queryTimeout={jsonData.queryTimeout}
+              rowCapacityHint={jsonData.rowCapacityHint}
```
ryanmorris708, line 193: Should we add `!!rowCapacityHint` to the conditional for `shouldBeOpen`?

adamyeats: Good call! `shouldBeOpen` already checks the other query settings fields, so `rowCapacityHint` belongs there too. Added alongside `queryTimeout` inside the non-single-table branch, since that's where `QuerySettingsConfig` renders.

ryanmorris708 (approved):
LGTM 😄

PR #2015, opened 2026-07-10
ci: enforce conventional commit pr titles and adopt release-please

adamyeats:
> [!WARNING]
> **The `Plugins - CI` run fails at startup (zero jobs), not on a defect.** `push.yml` and `publish.yml` are thin callers of the shared `datasource-cicd.yml` / `datasource-publish.yml` reusable workflows, which aren't on `data-sources-ci-workflows` main yet (they land in grafana/data-sources-ci-workflows#16). GitHub can't resolve a reusable workflow that isn't on the referenced ref, so the run fails before any job starts. It goes green once #16 is merged to main. Nothing to fix on this PR.

## Type of Change

- [x] 🧹 Refactor / Chore

## Please check that:

- [ ] Tests for this change have been added/updated. — N/A, CI configuration only
- [ ] Documentation has been added/updated (where applicable). — N/A

This PR automates releases end to end: it enforces Conventional Commits on PR titles, adopts release-please for the release PR, and publishes to the prod catalog automatically when a release PR merges. Since merges are squash-only with `PR_TITLE` as the commit subject, the title check guarantees conventional commits on `main`, which is what release-please consumes.

`push.yml` and `publish.yml` are thin callers of the shared `datasource-cicd.yml` and `datasource-publish.yml` reusable workflows in [data-sources-ci-workflows](https://github.com/grafana/data-sources-ci-workflows) (#16), so release-please, the prod publish, and the docs publish all live in one shared place. There is no per-repo `release-please.yml`.

**What changes:**

- Every PR title must be Conventional Commits format (`feat`, `fix`, `perf`, `refactor`, `docs`, `test`, `build`, `ci`, `chore`, `revert`), subject starting lowercase.
- After each push to `main`, release-please maintains a release PR (`chore(main): release X.Y.Z`) with the version bump (`package.json` + `package-lock.json`) and a generated `CHANGELOG.md` section. The PR is created with the `grafana-oss-big-tent` app token so CI runs on it.
- Merging the release PR creates the `vX.Y.Z` tag and GitHub release, and the shared pipeline's prod-publish job runs the CD workflow from that tag (`environment: prod`, `scopes: grafana_cloud`). Building from the tag guarantees the published artifact is exactly the released commit, even if `main` has moved on. CD's tag-creation step is `continue-on-error` and its release step updates the release that release-please created, attaching the plugin zips — the interplay the [official plugin-ci-workflows release-please example](https://github.com/grafana/plugin-ci-workflows/blob/main/examples/extra/release-please.yml) documents (it also explicitly warns against `skip-github-release`, which is why this does not use it).
- Renovate now uses `chore(deps)` for routine updates, and `chore` is hidden from the changelog, keeping dependency noise out (matching the current hand-curated style). Security updates are carved out via `vulnerabilityAlerts`: Renovate PRs raised from vulnerability alerts use `fix(deps)`, which is changelog-visible and bumps the patch version, so a CVE fix produces a release PR on its own. The `vulnerabilityAlerts` block merges on top of `packageRules` (and with the shared preset's existing `security` label and branch prefix), so the carve-out wins over the blanket `chore` rule. Routine deps-only churn still never triggers a release — if one is ever needed without a security alert, use a `Release-As: x.y.z` commit footer.

**Docs and manual publishes:** `publish.yml` is now a thin caller of the shared `datasource-publish.yml`. It auto-publishes docs to the website on the GitHub `release` event, and its manual `workflow_dispatch` remains the path for on-prem publishes, docs-only publishes, and emergency re-runs.

**First release PR** after this merges will be 4.21.0 (post-4.20.0 commits include `feat` #2019 plus several fixes). The manifest is bootstrapped at 4.20.0 with `bootstrap-sha` at the v4.20.0 tag commit. The removed `## Unreleased` CHANGELOG entry (#2092) regenerates from its merged commit; richer wording can be restored by editing the release PR branch before merge.

**Merge prerequisite:** the GitHub App token grants for the `push.yml` and `publish.yml` workflow paths must be in place before this merges — the `grafana-oss-big-tent` grant on `push.yml` for the release-please job, and the `website-docs` grant on `publish.yml` for the release-triggered docs publish. The grants are configured internally and are in review. Without the first, the initial run on `main` fails at token minting; without the second, the first docs auto-publish fails at the docs-publish step.

**Reviewer notes:**

- Release-please behaviour is only exercisable once it runs on `main`.
- Merging a release PR is now the "publish to prod (cloud scope)" decision. If the `prod` environment has protection rules, the auto-triggered CD run will wait on them.
- A routine version bump that silently contains a CVE fix without a matching vulnerability alert still lands as `chore(deps)` — the carve-out covers alert-driven PRs, which is the same coverage the `security` labelling has today.
- Config validated with `renovate-config-validator` and `zizmor` (no findings); non-Grafana actions are SHA-pinned per the zizmor policy.


codyconfer (dismissed):

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,65 @@
...
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: read
+
+concurrency:
+  group: release-please
+  cancel-in-progress: false
```
aangelisc: Is it possible for concurrent merges to create conflicting modifications of the changelog? I expect not because they should be appends but better to be sure.

release-please-config.json
```diff
@@ -0,0 +1,24 @@
...
+    { "type": "docs", "section": "📝 Documentation" },
+    { "type": "test", "section": "✅ Tests" },
+    { "type": "build", "section": "🏗️ Builds" },
+    { "type": "ci", "section": "🤖 Continuous Integration" },
+    { "type": "revert", "section": "⏪ Reverts" },
+    { "type": "chore", "section": "🔧 Chores", "hidden": true }
+  ],
+  "packages": {
+    ".": {
+      "package-name": "clickhouse-datasource"
```
aangelisc: Should this match the plugin slug just for consistency?

.github/workflows/release-please.yml
adamyeats: There are two guards here. The workflow-level concurrency group serialises release-please runs (`cancel-in-progress: false` keeps at most one running while the newest queued run waits), and release-please itself is regenerative rather than append-based. Each run recomputes the whole release PR (version and changelog section) from every conventional commit since the last release and force-updates its own branch, so a queued run that gets superseded loses nothing. `CHANGELOG.md` on `main` only changes when the release PR itself merges, so there is no competing writer to conflict with.

release-please-config.json
adamyeats: Changed to `grafana-clickhouse-datasource` in be065d9. With `include-component-in-tag: false` the field is mostly cosmetic, and CD already names GitHub releases with the plugin ID, so matching the slug is the more consistent choice.

aangelisc (dismissed):

.github/workflows/push.yml
```diff
@@ -1,62 +1,51 @@
...
-    # Required for checkout + OIDC in shared workflows
+  cicd:
+    name: CI/CD
+    # Intentionally pinned to @main: data-sources-ci-workflows does not publish
+    # versioned tags for datasource-cicd.yml yet. Allowed by the zizmor policy
+    # (`grafana/*: any`); review periodically.
+    uses: grafana/data-sources-ci-workflows/.github/workflows/datasource-cicd.yml@main # zizmor: ignore[unpinned-uses]
     permissions:
-      contents: read
+      contents: write
```
aangelisc, line 54: Is it possible to remove this permission for PRs?

adamyeats: These are a ceiling for the called workflow rather than a grant to each job, and on a `pull_request` run only `ci` executes, which declares `contents: read` itself. So nothing on the PR path gets a write-capable token, and the write entries are there for the release path on `main`.

release-please-config.json
```diff
@@ -0,0 +1,24 @@
+{
+  "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json",
+  "release-type": "node",
+  "include-v-in-tag": true,
+  "include-component-in-tag": false,
+  "bootstrap-sha": "973c4a508220c378ef8a477f3133bba935ca2345",
+  "changelog-sections": [
+    { "type": "feat", "section": "🎉 Features" },
```
yesoreyeram, line 8: @adamyeats - How the breaking changes will be documented in changelog? Any pointers much appreciated.

adamyeats: Here's an example from another Grafana repo: https://github.com/grafana/otel-profiling-ruby/pull/43

aangelisc (approved):

PR #2018, opened 2026-07-10
ci(otel): add scheduled canary for collector schema drift

### LINKED ISSUE #1901: Add scheduled canary workflow to detect OTel collector schema drift
## Background

`opentelemetry-collector-contrib`'s `clickhouseexporter` ships breaking schema changes from time to time. Most recently, [v0.151.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.151.0) removed the `TimestampTime` column from the default `otel_logs` table — see #1882 for the user-facing impact and #1899 for the fix.

We currently learn about these changes the same way users do: when something breaks in production and someone files an issue. The plugin's e2e suite uses a hand-rolled SQL fixture ([`tests/e2e/fixtures/otel_logs_v0151.sql`](https://github.com/grafana/clickhouse-datasource/blob/main/tests/e2e/fixtures/otel_logs_v0151.sql)) that mirrors what the v0.151.0 exporter produces — which locks in the SQL the plugin emits, but doesn't catch the next breaking change in the exporter.

## Proposal

Add a separate scheduled GitHub Actions workflow (weekly cron, plus manual `workflow_dispatch` and a `schema-canary` PR label trigger) that:

1. Spins up `otel/opentelemetry-collector-contrib:latest` and a fresh ClickHouse cluster via docker compose.
2. Configures the collector with the `clickhouse` exporter and `create_schema: true`.
3. Pushes a small synthetic OTLP payload (logs + traces) to the collector via OTLP/HTTP — enough to trigger schema creation.
4. Waits for the tables to materialize, then runs `DESC TABLE otel_logs` / `otel_traces` and dumps the column names + types.
5. Asserts the columns match the plugin's `latest` OTel schema map ([`src/otel.ts`](https://github.com/grafana/clickhouse-datasource/blob/main/src/otel.ts)) — for both logs and traces.

When the assertion fails, the workflow opens an issue (or comments on a pinned tracking issue) with the schema diff. That's the signal to add a new entry to `src/otel.ts` — exactly the trigger we lacked for #1882.

## Why this and not per-PR e2e

Running the real collector container per PR would gate every change on:

- Pulling the collector image
- An OTLP sender (curl/Go shim) to generate enough data
- Async ingest — table creation isn't instant; tests need polling
- A pinned collector version we'd have to bump manually anyway

A weekly cron gives us the regression coverage without the per-PR cost, and the failure mode is \"a tracking issue gets filed\" — non-blocking, with the diff already laid out.

## Scope

- Logs and traces only. Metrics tables haven't been touched by exporter rewrites yet; add later if they start drifting.
- Compare column names + types only. Don't try to match `ORDER BY` / `PARTITION BY` / index definitions — those are exporter implementation details that don't break our SQL emission.
- Pin the canary to a moving target (`:latest` or `:0`), not a frozen version — the whole point is detecting drift.

## Out of scope

- Auto-applying the schema fix. The workflow flags the diff; a human picks the right migration path.
- Replacing the existing hand-rolled fixture in [`tests/e2e/otelLogsV151.spec.ts`](https://github.com/grafana/clickhouse-datasource/blob/main/tests/e2e/otelLogsV151.spec.ts) — that locks in *our* SQL emission and runs on every PR. This canary is complementary, not a replacement.

## Related

- #1882 — the schema bump that motivated this follow-up
- #1899 — the fix; the new \`tests/e2e/fixtures/otel_logs_v0151.sql\` fixture is what this canary would protect against rotting
- #1900 — runtime auto-detection of the schema (compile-time canary vs. runtime detection are complementary)

adamyeats:
## Type of Change

- [x] 🚀 Feature

---

### What is this feature?

A scheduled canary workflow (`.github/workflows/otel-schema-canary.yml`) that detects schema drift in `opentelemetry-collector-contrib`'s `clickhouseexporter` before users hit it:

1. Boots the real collector (`:latest`, intentionally unpinned) against a fresh ClickHouse via docker compose.
2. Lets the exporter create its own schema (`create_schema: true`), then pushes a small synthetic OTLP logs + traces payload and waits for the rows to land, so insert-breaking changes fail the canary too.
3. Diffs the created `otel_logs`, `otel_traces`, and `otel_traces_trace_id_ts` tables (column names, types, and default kinds) against the checked-in snapshot `tests/canary/expected-columns.json`.
4. On drift, files (or comments on) a tracking issue with the markdown diff and resolution steps. Infrastructure failures (image pull, startup, ingest timeout) exit with a distinct code and file no issue, to avoid noise.

It runs weekly (Monday 07:00 UTC), on `workflow_dispatch`, and on PRs labelled `schema-canary`. Everything also runs locally: `./tests/canary/run.sh` to check, `./tests/canary/run.sh --update` to regenerate the snapshot.

A second guard closes the loop on every PR: new unit tests in `src/otel.test.ts` assert the plugin's latest column maps only reference columns present in the snapshot, so `src/otel.ts` and the snapshot cannot drift apart silently.

The exporter ships breaking schema changes from time to time. v0.151.0 removed `TimestampTime` from `otel_logs` (#1882, fixed by #1899), and we learned about it from user bug reports. This canary turns that into a tracking issue with the diff already laid out, which is exactly the trigger we lacked for #1882.

Closes #1901

### Who is this feature for?

Plugin maintainers. Users benefit indirectly by getting schema-map updates before their upgrades break queries.

### How to test this feature

1. Run `./tests/canary/run.sh` locally (requires Docker). It should report that the schema matches the snapshot and exit 0.
2. Delete or retype any column in `tests/canary/expected-columns.json` and rerun. It should print a markdown diff and exit 1.
3. Trigger the workflow via `workflow_dispatch` once merged, or add the `schema-canary` label to this PR to run it here.

---

## Please check that:

While building this I found the canary already has something to say: the current collector (v0.156.0) has drifted from the v0.151 layout captured in `tests/e2e/fixtures/otel_logs_v0151.sql`. `otel_logs` gained eight `__otel_materialized_k8s.*` MATERIALIZED columns and `EventName` changed from `LowCardinality(String)` to `String`. Neither change affects the SQL the plugin emits (no `src/otel.ts` change needed), and the snapshot in this PR captures v0.156.0 as the baseline going forward.

The comparison deliberately ignores `ORDER BY`, `PARTITION BY`, and index definitions, which are exporter implementation details that don't affect the plugin's SQL. The tracking issue is found by title search rather than by label, so the workflow doesn't depend on a label existing in the repo. Verified end-to-end locally against real containers: the match path, all three drift modes (added, removed, retyped), and both `--update` and compare runs. actionlint is clean apart from intentional SC2016 notices (single-quoted markdown backticks in the issue body heredoc).


SpencerTorres (approved):
Good addition 🚀

tests/canary/compare-schema.mjs
```diff
@@ -0,0 +1,107 @@
...
+  const url = new URL(baseUrl);
+  url.searchParams.set(
+    'query',
+    'SELECT name, type, default_kind FROM system.columns' +
+      ' WHERE database = {db:String} AND table = {table:String} ORDER BY position FORMAT JSON'
+  );
+  url.searchParams.set('param_db', database);
+  url.searchParams.set('param_table', table);
+
+  const res = await fetch(url);
```
ryanmorris708, line 35: This should be wrapped in a try/catch block to handle connection failures

ryanmorris708 (commented):
Nice work! Just one comment from me 😄

ryanmorris708 (approved):

PR #2019, opened 2026-07-12
feat: add time series statement macros ($__columns, $__rateColumns, $__perSecondColumns, $__increaseColumns, $__lttb)

### LINKED ISSUE #2024: Add time series statement macros
Shaping time-series results with column pivots, rates, per-second and increase counters, or LTTB downsampling required verbose hand-written SQL. This adds the `$__columns`, `$__rateColumns`, `$__perSecondColumns`, `$__increaseColumns` and `$__lttb` statement macros to the backend macro engine, with tests and query editor documentation.

Created retroactively to record effort attribution for https://github.com/grafana/clickhouse-datasource/pull/2019.

adamyeats:
## Type of Change

- [x] 🚀 Feature

### What is this feature?

Five new SQL editor macros that expand to a complete statement rather than a single expression. Written in place of the `SELECT` clause, they collapse the canonical "one time series per label value" query into one line:

```sql
$__perSecondColumns(EventTime, ServiceName, RequestsTotal) FROM requests WHERE ServiceName != ''
```

expands to a single flat query that applies the panel time filter, buckets timestamps with `$__timeInterval`, groups per bucket and key, and computes the per-second counter rate with `lagInFrame` over a window partitioned by key:

```sql
SELECT toStartOfInterval(toDateTime(EventTime), INTERVAL 20 second) AS t, toString(ServiceName) AS metric,
       if((t - lagInFrame(t, 1, t) OVER w) = 0 OR (max(RequestsTotal) - lagInFrame(max(RequestsTotal), 1, max(RequestsTotal)) OVER w) < 0, nan,
          (max(RequestsTotal) - lagInFrame(max(RequestsTotal), 1, max(RequestsTotal)) OVER w) / (t - lagInFrame(t, 1, t) OVER w)) AS value
FROM requests
WHERE EventTime >= toDateTime(...) AND EventTime <= toDateTime(...) AND (ServiceName != '')
GROUP BY t, metric
WINDOW w AS (PARTITION BY metric ORDER BY t ASC ROWS BETWEEN 1 PRECEDING AND CURRENT ROW)
ORDER BY t, metric
```

The five macros:

| Macro | Purpose |
| ----- | ------- |
| `$__columns(timeCol, key, value)` | One series per key value |
| `$__rateColumns(timeCol, key, value)` | Per-second smoothing for gauges |
| `$__perSecondColumns(timeCol, key, value)` | Counter rate, like Prometheus `rate()` |
| `$__increaseColumns(timeCol, key, value)` | Counter delta, like Prometheus `increase()` |
| `$__lttb(buckets, x, y)` | Downsampling via ClickHouse's native `lttb()`, with `auto` bucket derivation |

Graphing a counter per host or service currently requires hand-writing bucketing, time filtering, grouping, and window arithmetic in every panel, and the window arithmetic in particular (frame bounds, counter-reset handling, first-point handling) is easy to get subtly wrong. These macros make the common patterns one-liners with correct semantics: windows are partitioned per series rather than relying on physical row order, counter resets and first points are emitted as `nan`, and output is long format which the plugin renders as multi-line series natively.

### Who is this feature for?

Users graphing Prometheus-style counter and gauge patterns from ClickHouse tables in the SQL editor, and users of other ClickHouse dashboarding tools who expect this family of macros to exist.

### How to test this feature

1. Point the data source at any table with a DateTime column, a string label column, and a numeric column.
2. In the SQL editor, run `$__columns(timeCol, labelCol, count() AS c) FROM db.table` with the Time series query type and confirm one line per label value.
3. Run `$__perSecondColumns(timeCol, labelCol, counterCol) FROM db.table` and confirm per-second rates, with gaps (`nan`) at counter resets and series starts.
4. Run `$__lttb(auto, timeCol, valueCol) FROM db.table` and confirm the series is downsampled to roughly one point per interval.
5. Error paths: omit the `FROM` clause, add a `GROUP BY`, or place the macro inside a subquery, and confirm the clear downstream-classified error messages.

## Please check that:

- These are statement macros, so they cannot be macropro `MacroFunc` handlers (those only see their own arguments). They run as a pre-pass in `pkg/macros/statement.go` before `macropro.Interpolate`, and the generated SQL contains `$__timeInterval`/`$__timeFilter` calls so macropro remains the single source of truth for time expansion. Upstreaming a statement-macro concept into macropro was considered and deliberately deferred: this plugin is the only consumer of the pattern in the fleet, so the API would be designed on one data point. The pre-pass reuses `macropro.StripComments` (length-preserving) so comment handling is not duplicated.
- The time column is an explicit first argument because the SQL editor has no configured timestamp column to infer it from.
- Generated SQL was executed and semantically verified against ClickHouse 25.8 (minimum supported) and latest, including counter-reset isolation between series and stable `Float64` output types.
- One statement macro per query, at the top level only, with clear errors otherwise. Text before the macro (such as a `WITH` clause) is preserved.

Closes #2024


docs/sources/query-editor.md
```diff
@@ -300,6 +300,31 @@ WHERE $__timeFilter(date_time)
...
 
+### Statement macros
+
+Statement macros expand to a complete query rather than a single expression. Write the macro in place of the `SELECT` clause, followed by the rest of the query. Everything from the macro to the end of the query is replaced with a generated statement that applies the panel time filter, buckets timestamps with `$__timeInterval()`, and returns one series per key value.
+
+| Macro                                         | Description                                                                                                                                                                                   |
+| --------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `$__columns(timeColumn, key, value)`          | One series per `key` value, with `value` aggregated per time bucket.                                                                                                                          |
```
itsmylife, line 309: When I use this query 
```
$__columns(timestamp, `service,name`, count())
FROM (
  SELECT timestamp, service AS `service,name`
  FROM e2e_test.events
)
```
this gives me 
```
Could not apply macros: unexpected number of arguments: $__columns expected 3 argument(s), received 4
```

This was working though 
```
SELECT timestamp, `service,name`
FROM (
  SELECT timestamp, service AS `service,name`
  FROM e2e_test.events
)
```

docs/sources/query-editor.md
```diff
@@ -300,6 +300,31 @@ WHERE $__timeFilter(date_time)
...
+| `$__rateColumns(timeColumn, key, value)`      | Like `$__columns`, but divides the aggregated value by the number of seconds since the previous bucket of the same series. Use it to smooth gauge-style values into per-second rates.         |
+| `$__perSecondColumns(timeColumn, key, value)` | Per-second rate of a monotonic counter per series, like the Prometheus `rate()` function. `value` is aggregated with `max()`.                                                                 |
+| `$__increaseColumns(timeColumn, key, value)`  | Raw increase of a monotonic counter per bucket and series, like the Prometheus `increase()` function. `value` is aggregated with `max()`.                                                     |
+| `$__lttb(buckets, x, y)`                      | Downsamples dense series with ClickHouse's `lttb()` aggregate function (Largest-Triangle-Three-Buckets). Pass a bucket count, or `auto` to derive one from the panel time range and interval. |
```
itsmylife, line 313: This report is coming from agent review directly.

`$__lttb` allows an aggregated y-expression, but generates nested aggregation such as `lttb(..., avg(value)),` which ClickHouse rejects with `ILLEGAL_AGGREGATION`; please pre-aggregate through a subquery or reject this usage explicitly.

Working:
```
$__lttb(5, ts, value)
FROM (
  SELECT now() - number * 60 AS ts, toFloat64(number) AS value
  FROM numbers(10)
)
```

Broken:
```
$__lttb(5, ts, avg(value))
FROM (
  SELECT now() - number * 60 AS ts, toFloat64(number) AS value
  FROM numbers(10)
)
GROUP BY ts
```

pkg/macros/statement.go
```diff
@@ -0,0 +1,609 @@
...
+		}
+	}
+	return found, nil
+}
+
+// scanQuoted advances past the string literal opening at pos and returns the
+// index after the closing quote. Both the doubled-quote escape and ClickHouse
+// backslash escapes are handled, mirroring the BackslashEscape flag set in
+// clickHouseComments. An unterminated literal consumes the rest of the input.
+func scanQuoted(s string, pos int) int {
```
itsmylife, line 178: I'm not sure about how ClickHouse handles backticks but does this have to handle backticks too?

itsmylife: Actually this is not a question specific to this function but a general question.

itsmylife (commented):
I did some agentic and manual testing and shared but findings. I'm not familiar with ClickHouse too much so if my findings doesn't make sense please ignore. But I'd appreciate some explanation.

docs/sources/query-editor.md
adamyeats: The argument splitter only treated single and double quotes as opaque, so the comma inside `` `service,name` `` split the argument list. Fixed in da5e0f1b7bc976c89e5d612224ca07c70ed5cac0. Backtick-quoted identifiers are now treated as quoted regions everywhere the statement scanner reads the query (argument splitting, parenthesis matching, clause detection and alias parsing), and there is a regression test for this exact shape.

docs/sources/query-editor.md
adamyeats: The expansion nested `avg()` inside the `lttb` aggregate call, which ClickHouse rejects with code 184, exactly as you reported. Fixed in da5e0f1b7bc976c89e5d612224ca07c70ed5cac0 via the pre-aggregation route you suggested. When the tail contains a `GROUP BY`, the aggregation now runs in a subquery and `lttb` consumes its aliased output columns. A generated `ORDER BY` on the x column is included because `GROUP BY` output order is unspecified and `lttb` buckets by input order. The failing example now returns the five downsampled points. One caveat worth noting is that an aggregate y with no `GROUP BY` at all still fails in ClickHouse itself, the same as writing the nesting by hand.

pkg/macros/statement.go
adamyeats: It does, as it turns out. ClickHouse accepts backtick-quoted identifiers with the same escape rules as string literals (a doubled quote or a backslash escape, verified against a live server), and the scanner missing them was the root cause of the `` `service,name` `` report above. Fixed in da5e0f1b7bc976c89e5d612224ca07c70ed5cac0 by treating backticks as a third quote character. The comment-stripping layer already handles them through the `BacktickQuote` flag set in `clickHouseComments`.

itsmylife (approved):

PR #2020, opened 2026-07-13
fix(logs): correct and align attribute filters added from the log view

alex-fedotyev:
## Summary

Filtering a log on a `Map` attribute key from the log view had two problems, both of which this PR fixes so the field "+" behaves like the Add filter flow:

1. It generated invalid SQL and crashed the query.
2. The resulting filter chip was labelled inconsistently with the Add filter chip.

### 1. Invalid SQL (crash)

The field filter-for/out buttons emitted dot access on a `Map` column:

```
... AND ( LogAttributes.`user.tier`::Nullable(String) = 'basic' )
```

ClickHouse rejects this with `code: 47, Unknown expression or function identifier ... Maybe you meant: ['LogAttributes.keys']`. A `Map` key needs bracket access:

```
... AND ( LogAttributes['user.tier'] = 'basic' )
```

`buildFilter` set the filter type from the selected column and defaulted to `JSON` when that column had no type. The default OTel columns are stored without a type, and a query loaded from the Explore URL keeps them that way, so an attribute map-key filter fell into the JSON dot-path branch of the SQL generator instead of the `Map` bracket branch. The Add filter editor was unaffected because it reads the type from the full schema; only the field filter-for/out actions (`modifyQuery` / `toggleQueryFilter`) went through the defaulting path.

### 2. Inconsistent chip label

The field "+" stores a hinted column as `{ key: '', hint }` (by design, so the SQL generator resolves the column through the hint), while the Add filter flow stores the raw column name in `key`. `FilterTagBar` rendered the hint verbatim, so the same field showed as `log attributes.event.name` from the "+" and `LogAttributes.event.name` from Add filter. The chip now resolves a hinted filter back to its column name, so both read `LogAttributes.event.name`, matching the log detail panel.

## Changes

- Resolve the real column type from the fetched schema cache in `modifyQuery` and `toggleQueryFilter`, so `Map` vs `JSON` is decided authoritatively even when the selected column has no type. This also lets a non-selected `Map` column (for example a trace `SpanAttributes` key) split and type correctly.
- Default an unknown-type attribute map-key filter to `Map` bracket access instead of `JSON` dot access. Columns that report a `JSON` type still use dot access.
- In `FilterTagBar`, resolve a hinted filter to its column name for display. This is presentation only; the filter key/hint shape, generated SQL, and dedup are unchanged.

## Local verification

Ran a local Grafana with the official plugin and this build side by side against the public demo (`sql-clickhouse.clickhouse.com`, `otel_v2`), filtering on attribute keys from the compact query builder.

**Logs, official plugin (before):** each attribute chip shows the hint label (`log attributes.*`) and the query fails with `code: 47`.

logs before

**Logs, this build (after):** `LogAttributes`, `ResourceAttributes`, and `ScopeAttributes` chips all show the raw column name and the query runs (bracket access).

logs after

**Traces, this build:** `ResourceAttributes` and `SpanAttributes` attribute filters render the raw column name and run with bracket access, consistent with logs.

traces after

## Test plan

- [x] `npx jest src/data/sqlGenerator.test.ts src/data/CHDatasource.test.ts src/components/queryBuilder`
- [x] `npx tsc --noEmit`
- [x] `npx eslint` on the changed files
- [x] `npx prettier --check` on the changed files
- [x] `npm run spellcheck`
- [x] Manual check on the local Grafana build against the demo, logs and traces

### DIFF
--- a/src/components/queryBuilder/QueryBuilder.test.tsx
+++ b/src/components/queryBuilder/QueryBuilder.test.tsx
@@ -104,13 +104,22 @@ describe('QueryBuilder', () => {
     expect(result.container.firstChild).not.toBeNull();
   });
 
-  it('shows the raw attribute column name in the compact filter chip for a hinted "+" filter', async () => {
-    const originalGetSignalType = mockDs.getSignalType;
-    const originalIsSingleTableMode = mockDs.isSingleTableMode;
-    // Compact editor renders only when the datasource is single-table with a signal type.
-    mockDs.getSignalType = jest.fn(() => 'logs' as any);
-    mockDs.isSingleTableMode = jest.fn(() => true);
-    try {
+  describe('compact filter chip label', () => {
+    let getSignalTypeSpy: jest.SpyInstance;
+    let isSingleTableModeSpy: jest.SpyInstance;
+
+    beforeEach(() => {
+      // Compact editor renders only when the datasource is single-table with a signal type.
+      getSignalTypeSpy = jest.spyOn(mockDs, 'getSignalType').mockReturnValue('logs' as any);
+      isSingleTableModeSpy = jest.spyOn(mockDs, 'isSingleTableMode').mockReturnValue(true);
+    });
+
+    afterEach(() => {
+      getSignalTypeSpy.mockRestore();
+      isSingleTableModeSpy.mockRestore();
+    });
+
+    it('shows the raw attribute column name in the compact filter chip for a hinted "+" filter', async () => {
       render(
         <QueryBuilder
           app={CoreApp.PanelEditor}
@@ -143,15 +152,12 @@ describe('QueryBuilder', () => {
         />
       );
 
-      // The log-view "+" stores { key: '', hint }, so the chip must resolve the hint to
-      // the column name and read "LogAttributes.user.tier" (matching the Add filter path),
-      // not the friendly hint label "log attributes.user.tier".
+      // The log-view "+" stores { key: '', hint }, so the chip must resolve the hint to the
+      // column name and read "LogAttributes.user.tier" (matching the Add filter path), not the
+      // friendly hint label "log attributes.user.tier".
       expect(await screen.findByText('LogAttributes.user.tier')).toBeInTheDocument();
       expect(screen.queryByText('log attributes.user.tier')).not.toBeInTheDocument();
-    } finally {
-      mockDs.getSignalType = originalGetSignalType;
-      mockDs.isSingleTableMode = originalIsSingleTableMode;
-    }
+    });
   });
 
   it('renders TableQueryBuilder when queryType is Table', () => {
--- a/src/components/queryBuilder/QueryBuilder.test.tsx
+++ b/src/components/queryBuilder/QueryBuilder.test.tsx
@@ -3,7 +3,7 @@ import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { getCompactFilterColumns, QueryBuilder } from './QueryBuilder';
 import { getDefaultCompactMode } from './CompactModeBar';
 import { Datasource } from 'data/CHDatasource';
-import { BuilderMode, ColumnHint, QueryType, TimeUnit } from 'types/queryBuilder';
+import { BuilderMode, ColumnHint, FilterOperator, QueryType, TimeUnit } from 'types/queryBuilder';
 import { CoreApp } from '@grafana/data';
 
 jest.mock('./views/TableQueryBuilder', () => ({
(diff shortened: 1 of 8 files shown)


src/data/CHDatasource.ts
```diff
@@ -134,11 +151,16 @@ function resolveFilterColumn(builderOptions: QueryBuilderOptions, key: string):
     : undefined;
   const column = lookupByAlias || lookupByName || lookupByLogsAlias;
 
+  // Prefer the selected column's type, but fall back to the live schema type so
+  // map-key filters resolve Map vs JSON authoritatively even when the selected
+  // column is stored without a type (the default OTel columns are).
+  const columnType = column?.type || knownColumns?.find((c) => c.name === columnName)?.type || '';
```
adamyeats: When the attribute column resolves via its alias, `columnName` holds the alias rather than the real column name, so this schema fallback misses and an aliased untyped JSON column would fall through to the Map default. Consider looking up `column?.name ?? columnName` here instead.

src/data/CHDatasource.ts
```diff
@@ -123,6 +132,14 @@ function resolveFilterColumn(builderOptions: QueryBuilderOptions, key: string):
     } else if (legacyAttributeColumns.has(columnPrefix)) {
       mapKey = columnName.substring(prefixIndex + 1);
       columnName = columnPrefix;
+    } else if (
+      knownColumns?.some((c) => c.name === columnPrefix && (c.type?.startsWith('Map') || c.type?.startsWith('JSON')))
```
adamyeats: nit: this Map/JSON prefix test also appears in `getAttributeColumnByDisplayPrefix`. It might be worth a small `isAttributeType(type)` helper so the two stay in sync.

src/data/CHDatasource.ts
```diff
@@ -150,7 +172,10 @@ function buildFilter(resolved: ResolvedColumn, operator: StringFilter['operator'
     key: resolved.column?.hint ? '' : resolved.columnName,
     hint: resolved.column?.hint || undefined,
     mapKey: resolved.hasMapKey ? resolved.mapKey : undefined,
-    type: resolved.hasMapKey ? (resolved.columnType.startsWith('Map') ? 'Map(String, String)' : 'JSON') : 'string',
+    // Map columns use bracket access (`col['key']`); only genuine JSON columns use
+    // the dot-path branch. Default an unknown-type attribute filter to Map, which
+    // matches the standard OTel schema and avoids emitting invalid `col.key` SQL.
+    type: resolved.hasMapKey ? (resolved.columnType.startsWith('JSON') ? 'JSON' : 'Map(String, String)') : 'string',
```
adamyeats, line 182: Worth noting the flipped default means an untyped filter on a genuine JSON column now emits bracket access when the schema cache is cold. The schema fallback covers the warm case and Map matches the standard OTel schema, so this seems the right trade, just flagging the behaviour change.

src/data/CHDatasource.test.ts
```diff
@@ -1609,6 +1609,93 @@ describe('ClickHouseDatasource', () => {
...
+        const traceQuery: CHBuilderQuery = {
+          ...untypedQuery,
+          builderOptions: {
+            ...untypedQuery.builderOptions,
+            table: 'otel_traces',
+            columns: [{ name: 'Timestamp', hint: ColumnHint.Time }],
+          },
+        };
+        // Prime the schema cache the way getColumnsCached would (key is db\0table).
+        (datasource as any)._columnCache.set((datasource as any).columnCacheKey('default', 'otel_traces'), [
```
adamyeats: nit: priming the private `_columnCache` couples the test to internals it has to name via `as any`. Consider mocking `fetchColumns` and awaiting `getColumnsCached` so the cache is primed through the public path.

src/components/queryBuilder/QueryBuilder.test.tsx
```diff
@@ -104,6 +104,56 @@ describe('QueryBuilder', () => {
     expect(result.container.firstChild).not.toBeNull();
   });
 
+  it('shows the raw attribute column name in the compact filter chip for a hinted "+" filter', async () => {
+    const originalGetSignalType = mockDs.getSignalType;
```
adamyeats: nit: `jest.spyOn(mockDs, 'getSignalType').mockReturnValue('logs')` with restoration in `afterEach` would avoid the manual save and restore and the try/finally.

src/data/CHDatasource.ts
alex-fedotyev: Good catch, that is a real bug for aliased columns. Fixed in 7ac49d0: the schema fallback now looks up by the real column name (`column?.name ?? columnName`), so an aliased untyped column resolves its type correctly. Added a regression test for the aliased-JSON case.

src/data/CHDatasource.ts
alex-fedotyev: Good call. Extracted an `isAttributeColumnType` helper in 7ac49d0, now shared by both the split branch and `getAttributeColumnByDisplayPrefix` so the Map/JSON check stays in sync.

src/data/CHDatasource.ts
alex-fedotyev: Agreed, that is the intended trade. Bracket access matches the standard OTel Map schema and the schema fallback covers typed JSON columns; the only regression is a cold-cache untyped JSON column, which is rare. Thanks for flagging it.

src/data/CHDatasource.test.ts
alex-fedotyev: Good call. In 7ac49d0 the test primes the cache through the public `getColumnsCached` path (mocking `fetchColumns`) instead of reaching into `_columnCache`, and I added an aliased-column case alongside it.

src/components/queryBuilder/QueryBuilder.test.tsx
alex-fedotyev: Done in 7ac49d0. Switched to `jest.spyOn` with `mockRestore`. I kept explicit spy handles in a small nested describe rather than a top-level `restoreAllMocks`, because restoring at the shared describe scope let the mocked signal type leak into the sibling render tests.

adamyeats (approved):
@alex-fedotyev Great work as ever, thanks!

PR #2022, opened 2026-07-14
feat(driver): support comma-separated hosts for load-balancing and failover

### LINKED ISSUE #279: Support load-balancing and failover hosts
**Is your feature request related to a problem? Please describe.**

I would like to connect to multiple ClickHouse servers to support our load-balancing and failover requirements.

**Describe the solution you'd like**

Provide the values of host(s) as a comma-separated list per  the [clickhouse-go](https://github.com/ClickHouse/clickhouse-go) driver's DSN `hosts` field. See https://github.com/ClickHouse/clickhouse-go#dsn

**Describe alternatives you've considered**

Deploying a separate  TCP proxy that would support native ClickHouse connections. The downside is that the client would not be directly aware of host connection failures (e.g., rolling update) and simply retry.

**Additional context**

N/A

Tarasusrus:
## Summary

Adds support for configuring multiple ClickHouse hosts as a comma-separated list in the datasource `host` setting, enabling clickhouse-go's native load-balancing and failover across nodes.

Closes #279

## Details

`clickhouse-go` load-balances and fails over across every entry in `Options.Addr`, and its own DSN parser already treats the host as a comma-separated list (`clickhouse_options.go`: `o.Addr = append(o.Addr, strings.Split(dsn.Host, ",")...)`). This plugin builds `Options` directly rather than via a DSN, so it previously collapsed the host into a single `host:port` address. This change mirrors the driver's convention:

- `host` is split on commas into multiple `host:port` addresses.
- Each entry may carry its own port (`ch1:9001,ch2`); entries without one use the shared port field.
- Empty entries and surrounding whitespace are ignored.
- TLS `ServerName` now uses the first host so SNI/verification keeps working when a list is configured.

The default `ConnOpenStrategy` (round-robin + failover) requires no extra configuration.

## Testing

- New table-driven unit test `TestBuildConnectionAddresses` covers single host, comma lists, whitespace trimming, per-host port override, empty-entry skipping, and bracketed IPv6.
- Full backend suite green (`go test ./pkg/...`), `go vet` and `go build` clean.

### DIFF
--- a/pkg/plugin/driver.go
+++ b/pkg/plugin/driver.go
@@ -45,12 +45,52 @@ type Clickhouse struct {
 	SchemaDatasource *schemas.SchemaDatasource
 }
 
+// buildConnectionAddresses turns the configured host setting into the list of
+// "host:port" addresses passed to clickhouse-go, which natively load-balances
+// and fails over across multiple entries. Host may be a single value or a
+// comma-separated list (see #279); each entry may carry its own "host:port",
+// otherwise the shared port is appended. Empty entries and surrounding
+// whitespace are ignored.
+func buildConnectionAddresses(host string, port int64) []string {
+	parts := strings.Split(host, ",")
+	addrs := make([]string, 0, len(parts))
+	for _, part := range parts {
+		p := strings.TrimSpace(part)
+		if p == "" {
+			continue
+		}
+		// If the entry already includes a port (host:port or [ipv6]:port),
+		// keep it verbatim; otherwise append the shared port.
+		if _, _, err := net.SplitHostPort(p); err == nil {
+			addrs = append(addrs, p)
+		} else {
+			addrs = append(addrs, fmt.Sprintf("%s:%d", p, port))
+		}
+	}
+	return addrs
+}
+
+// firstHost returns the first host in a possibly comma-separated host setting,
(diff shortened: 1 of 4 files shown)


adamyeats:
@Tarasusrus Thank you for this submission! However, for this one I think we want to discuss internally how we'd want to show this in the UI. A comma-seperated list of URLs in a single form field seems a little unintuitive to me as it stands. I'll talk to some people internally about this and get back to you!

Tarasusrus:
Thanks, that's a fair point. Looking back at the issue, I think I focused a bit too much on the proposed comma-separated input instead of the actual requirement, which is just supporting multiple ClickHouse nodes.

I've reworked it so the backend isn't tied to any particular UI. It now accepts a `hosts` array in `jsonData`, while keeping the existing `host`/`port` fields for backwards compatibility. That means the frontend can use whatever UI makes the most sense without any backend changes.

One thing worth calling out: `ServerName` is no longer set in `getTLSConfig`. `clickhouse-go` reuses a single `tls.Config` for every connection, so pinning `ServerName` would fail cert verification for all but one host. Leaving it unset lets Go infer it from the address being dialed for each connection. Covered by a test.

Happy to tweak the schema if you'd prefer a different shape, or just wait until you've had a chance to discuss the UI internally.

PR #2023, opened 2026-07-14
fix(variables): don't break variable loading when version() returns no rows

### LINKED ISSUE #1061: Grafana variable loading issue: "Templating [VARIABLE_NAME] Error updating options: Cannot read properties of undefined (reading 'split')"
**What happened**:
After setting up Grafana dashboard, having issue during dashboard loading where it is unable to load variables into filters.
The error thrown is - "Templating [VARIABLE_NAME] Error updating options: Cannot read properties of undefined (reading 'split').
It does work after reloading Grafana multiple times. This is frustrating to do all the times.

Expected dashboard to successfully load the variables in filters as the query is correct and tested before saving the dashboard.

Take any Clickhouse SQL query where you can fetch distinct values of column from any table. This is the same as we are facing.

1. Go to Dashboard -> Variables
2. Click on Any variable that you have created using Clickhouse query
3. Scroll down to bottom and check if query output is rendered and save the variable as well as dashboard
4. Load the Grafana in browser and see the error

**Screenshots**
1).

Image
2).

Image

**Anything else we need to know? **:
As we checked the console errors in browser, we found following info:

styles.ts:5 Unable to parse ClickHouse version: TypeError: Cannot read properties of undefined (reading 'split')

eval @ styles.ts:5Understand this errorAI

actions.ts:1028 TypeError: Cannot read properties of undefined (reading 'split')
    at eval (styles.ts:5:15)
    at Generator.next (<anonymous>)
    at Re (styles.ts:5:15)
    at o (styles.ts:5:15)
    at s.invoke (module.ts:31:16)
    at c.run (module.ts:31:16)
    at eval (module.ts:31:16)
    at s.invokeTask (module.ts:31:16)
    at c.runTask (module.ts:31:16)
    at m (module.ts:31:16)
(anonymous) @ actions.ts:1028Understand this errorAI

module.ts:31 Uncaught TypeError: Cannot read properties of undefined (reading 'split')
    at eval (styles.ts:5:15)
    at Generator.next (<anonymous>)
    at Re (styles.ts:5:15)
    at o (styles.ts:5:15)
    at s.invoke (module.ts:31:16)
    at c.run (module.ts:31:16)
    at eval (module.ts:31:16)
    at s.invokeTask (module.ts:31:16)
    at c.runTask (module.ts:31:16)
    at m (module.ts:31:16)Understand this errorAI

deprecationWarning.ts:14 [Deprecation warning] DataSource grafana-clickhouse-datasource: templateSrv.getAdhocFilters is deprecated. Use filters property on the request (DataQueryRequest). Or if this is called from interpolateVariablesInQueries or applyTemplateVariables it is passed as a new argument instead

This last one is important as it is throwing deprecationWarning.

**Environment**:

- Grafana version: Grafana v10.4.8 
- Plugin version: Latest
- OS Grafana is installed on: Linux
- User OS & Browser: MacOS
- Others: Google Chrome browser

Tarasusrus:
## Summary

Fixes a crash where dashboard template variables fail to load with:

> Templating [VARIABLE_NAME] Error updating options: Cannot read properties of undefined (reading 'split')

Closes #1061

## Root cause

`metricFindQuery` calls `canUseAdhocFilters()` to detect whether ClickHouse is >= 22.7. That method runs `SELECT version()` and did:

```ts
const verString = (data[0] as unknown as string).split('.');
```

When `version()` returns an empty response (reported as transient — "works after reloading multiple times"), `data[0]` is `undefined`, `.split` throws, and the `catch` block **rethrew**. The rejection propagated out of `metricFindQuery`, breaking variable loading for the whole dashboard.

## Fix

Guard the empty/unparseable version case: log a warning and return `AdHocFilterStatus.none` instead of throwing. Leaving the status unresolved means it is re-checked on the next query (self-healing for the transient case) rather than permanently disabling ad-hoc filters or crashing variable loading.

## Testing

- New test `does not throw when SELECT version() returns no rows (#1061)` reproduces the crash (red) and passes with the fix (green).
- Full `CHDatasource` suite green (165 passed), `tsc --noEmit` and `eslint` clean.

### DIFF
--- a/src/data/CHDatasource.test.ts
+++ b/src/data/CHDatasource.test.ts
@@ -127,6 +127,19 @@ describe('ClickHouseDatasource', () => {
 
       expect(querySpy).toHaveBeenCalledWith(expect.objectContaining({ scopedVars }));
     });
+
+    it('does not throw when SELECT version() returns no rows (#1061)', async () => {
+      const instance = cloneDeep(mockDatasource);
+      instance.adHocFiltersStatus = 0; // AdHocFilterStatus.none -> forces the CH version check
+      const metricFrame = toDataFrame({ fields: [{ name: 'field', type: 'string', values: ['a'] }] });
+      jest.spyOn(instance, 'query').mockImplementation((request) =>
+        request.targets[0].rawSql === 'SELECT version()'
+          ? of({ data: [toDataFrame([])] }) // empty version() response (transient / race)
+          : of({ data: [metricFrame] })
+      );
+
+      await expect(instance.metricFindQuery('SELECT 1', {})).resolves.toEqual([{ text: 'a', value: 'a' }]);
+    });
   });
 
   describe('applyTemplateVariables', () => {
--- a/src/data/CHDatasource.ts
+++ b/src/data/CHDatasource.ts
@@ -1774,8 +1774,17 @@ export class Datasource
   private async canUseAdhocFilters(): Promise<AdHocFilterStatus> {
     this.skipAdHocFilter = true;
     const data = await this.fetchData(`SELECT version()`);
+    const versionValue = data[0];
+    if (typeof versionValue !== 'string') {
+      // `SELECT version()` occasionally returns no rows (e.g. a transient/racey
+      // empty response right after Grafana loads). Leaving the status
+      // unresolved lets it be re-checked on the next query instead of throwing
+      // and breaking variable/template loading. See #1061.
+      console.warn('Unable to determine ClickHouse version: empty version() response');
+      return AdHocFilterStatus.none;
+    }
     try {
-      const verString = (data[0] as unknown as string).split('.');
+      const verString = versionValue.split('.');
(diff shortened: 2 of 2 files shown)


adamyeats (approved):
Thanks @Tarasusrus ! 🥳

adamyeats (changes requested):
Ah, `spellcheck` is failing... @Tarasusrus can you address that please? 🙏

Tarasusrus:
@adamyeats Fixed — the only complaint was `Unknown word (racey)` in a comment in `CHDatasource.ts`. Since "transient" already carries that meaning, I dropped the redundant word rather than extend the cspell dictionary for a single use. `cspell@6.13.3` now reports `Issues found: 0` locally with the same glob CI uses.

Also rebased onto `main` since the branch had fallen behind. No conflicts — the upstream commits did not touch `CHDatasource.ts`, and `CHDatasource.test.ts` still passes (165 tests).

Ready for another look whenever you have a moment.

adamyeats (approved):
Great, thanks @Tarasusrus ! 🙌

PR #1896, opened 2026-06-16
Fix datasource variable refs in Data Analysis dashboard

hkrutzer:
## Type of Change

_Please check the relevant option._

- [ ] 🚀 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧹 Refactor / Chore

---

### What is the bug?

The bundled Data Analysis dashboard defines its ClickHouse datasource template variable as datasource, but some panel/query datasource UID references still used the legacy import placeholder
${DS_GRAFANA-CLICKHOUSE-DATASOURCE}.

When the dashboard is provisioned, Grafana cannot resolve that placeholder and shows:

```txt
Datasource ${DS_GRAFANA-CLICKHOUSE-DATASOURCE} was not found
```

### How to reproduce

1. Provision or import the bundled Data Analysis dashboard from src/dashboards/data-analysis.json.
2. Select/configure a ClickHouse datasource using the dashboard’s datasource variable.
3. Open affected panels and observe that Grafana reports Datasource ${DS_GRAFANA-CLICKHOUSE-DATASOURCE} was not found because those panels reference the wrong datasource UID variable.

PR #2044, opened 2026-07-16
test(e2e): expand E2E coverage, extract shared helpers, and fix nightly Cloud flakiness

adamyeats:
## Type of Change

- [x] 🧹 Refactor / Chore

A coverage and tidy-up pass over the Playwright E2E suite, plus a fix for the recent [Scheduled Cloud E2E](https://github.com/grafana/clickhouse-datasource/actions/workflows/cron.yml) flakiness. Test-only, no user-facing changes. The suite goes from 83 tests in 17 flat files to 115 tests in 26 files grouped by theme, while removing over 1,000 lines of copy-pasted test code.

### Tidy-up (commit 1)

- New `tests/e2e/helpers/` modules replace per-spec copies that had drifted apart: `exploreUrl()` existed in 15 slightly different versions, `enterSql()` in 12, the query-response body wrapper in 10, and the env/datasource preamble in 15 (one spec had lost the cloud fallback entirely). The response accessors are now typed, removing ~40 `eslint-disable no-explicit-any` casts.
- Specs are grouped into thematic directories (`config/`, `editor/`, `builder/`, `filters/`, `traces/`, `logs/`, `types/`, `annotations/`). Pure `git mv` plus import-depth fixes, no Playwright config change needed (`testDir` discovery is recursive). No open PR touches `tests/e2e/`, so this creates no rebase conflicts.
- Naming normalised (describe casing, an internal ticket reference replaced with a behavioural title), compact query mode split out of `queryEditor.spec.ts`, cloud-skip messages now name the correct fixture file via a shared helper, a missing `serial` guard added to `columnAutodetect`, three generic fixture smokes removed as subsumed by richer suites, and the orphaned, unreferenced `tests/fixtures/property-prices.sql` deleted.

### New coverage (commit 2), prioritised by regression history

| Suite | Covers |
| ----- | ------ |
| `variables/variableEditor` | Guided query types and custom SQL previews, including the numeric-values regression (#2021) |
| `variables/variableInterpolation` | `$__conditionalAll`, multi-value and `sqlstring` interpolation through a real imported dashboard (#262) |
| `macros/macros` | First E2E execution of the backend macros: `$__timeFilter[_ms]`, `$__dateFilter`, `$__fromTime`/`$__toTime`, `$__timeInterval` (#534), `$__interval_s`, and a backslash-escape guard (#1991) |
| `filters/adhocDashboard` | The real dashboard ad hoc picker flow: `getTagKeys`, `getTagValues`, applied filters re-querying the panel (#1443) |
| `logs/logsVolume` | Supplementary logs-volume request, per-level grouping (#363), and the SQL-mode gate |
| `panels/panelRendering` | Rendered visualisations from fixture data: time series canvas, one series per group (#735), table view, Explore format mapping |
| `builder/builderQueries` | Builder-UI-to-executed-SQL wiring: aggregate mode, GROUP BY, ORDER BY, LIMIT, filter operators, editor round-trip |
| `dashboards/prebuiltDashboards` | Bundled Data/Query/Cluster Analysis dashboards import and render with zero panel errors (#535, #1896) |
| `config/configEditor` (extended) | HTTP-protocol health check and protocol port auto-defaults |

### Cloud cron flakiness fix (commit 3)

The nightly cron failed 6 times in the last 11 days, always on the same Explore rendering tests. Each test gets a fresh browser context, so on the shared Cloud instance every test re-downloads the plugin bundle and Monaco assets, and when that exceeds Playwright's 5s expect default the whole cluster fails through both retries. Local runs never see this because assets are served by the local container. The fix raises the test/expect ceilings to 90s/30s only when `GRAFANA_URL` is set (polling ceilings, so fast runs are unaffected), bumps the explicit Monaco suggest-widget wait to 15s, and records video only on failure outside CI (full-parallel local runs were CPU-bound enough to cause spurious timeouts).

## Please check that:

- Verified locally with three consecutive fully green runs against docker Grafana 12.2.0 (113 passed, 3 skipped, the skips being the CI-gated save-and-test credential checks). Opened as draft until the CI Grafana version matrix has run, since a handful of selectors in the new suites are version-sensitive by nature (noted inline where relevant).
- While writing the panel-rendering suite we found what looks like a product bug, deliberately not addressed here: a v4-shaped query with an empty or missing `pluginVersion` and no `format` field is misclassified as v3 by `isV3CHQuery` (`src/data/migration.ts`), so its explicit `queryType` is silently reset to Table. Reproducible on Grafana 12.2.0 by loading Explore with a pane state carrying `editorType: 'sql'`, `pluginVersion: ''` and `queryType: 'timeseries'`. Happy to open a follow-up issue.
- `.gitignore` gains a scoped negation because the pre-existing bare `logs` pattern (meant for log output) also matched the new `tests/e2e/logs/` directory.


ryanmorris708:
Hey @adamyeats! Dropping a comment to let you know I started reviewing this one and will be done on Monday 😄

tests/e2e/builder/columnAutodetect.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+      // User explicitly picks `service` (a valid String column) for Log Level.
+      await pickBuilderSelect(page, 'Log Level', 'service');
+      await expect(levelRow).toContainText('service');
+
+      // Toggling back to a different table should re-run auto-fill for *empty*
+      // slots but must not clobber the explicit `service` pick if the user
+      // returns to the original table. Simulate by flipping database and back.
+      // (The seed provides only one database, so instead we verify the pick
+      // survives a page re-render without the heuristic firing again.)
+      await expect(levelRow).toContainText('service');
```
ryanmorris708, line 58: This test is missing a step to trigger a re-render

tests/e2e/builder/builderQueries.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+
+/**
+ * Pick a value in a classic Grafana Select given its combobox locator
+ * directly, for Selects that are not label-anchored (the aggregate column
+ * and the order-by direction): open, type to filter, commit with Enter.
+ */
+async function pickSelectValue(page: Page, combobox: Locator, value: string) {
+  await combobox.click();
+  await page.keyboard.type(value);
+  await page.keyboard.press('Enter');
```
ryanmorris708: There may be a race condition where this can hit enter before the value is selectable

tests/e2e/builder/builderQueries.spec.ts
```diff
@@ -0,0 +1,283 @@
...
+
+      await expect(sqlPreview(page)).toContainText('ORDER BY timestamp DESC');
+
+      const body = await runQueryAndGetBody(page, explorePage);
+      // seed.sql inserts 10 rows, so all of them arrive under the default limit.
+      expect(rowCount(body)).toBe(10);
+      // The latest seeded row is 2024-03-15 10:09:00 UTC. Time fields are
+      // encoded as epoch millisecond numbers in the data frame JSON.
+      const firstTimestamp = frameValues(body)[0][0];
+      expect(Number(firstTimestamp)).toBe(Date.UTC(2024, 2, 15, 10, 9, 0));
```
ryanmorris708: The expected `Date` has a typo for the month

tests/e2e/variables/variableEditor.spec.ts
```diff
@@ -0,0 +1,204 @@
...
+ * Pick a value in one of the plugin editor's selects, located by its
+ * aria-label: open the combobox, type the value to filter the option list,
+ * commit with Enter and close any lingering menu with Escape (the same
+ * pattern as pickBuilderSelect in helpers/builder.ts).
+ */
+async function pickEditorSelect(page: Page, label: string, value: string): Promise<void> {
+  const combobox = pluginEditorSelect(page, label);
+  await combobox.click();
+  await page.keyboard.type(value);
+  await page.keyboard.press('Enter');
```
ryanmorris708: This is the same race condition I mentioned in builderQueries.spec.ts

tests/fixtures/seed.sql
ryanmorris708, line 1: There are a few fixtures that aren't reseed-safe and cause tests to fail if the docker compose stack is restarted

ryanmorris708 (dismissed):
Great work! I found just a few nits

tests/e2e/builder/columnAutodetect.spec.ts
adamyeats: Good catch, the closing assertion re-checked state the test had already asserted, so it could never observe a late clobber.

A table or database flip can't exercise this either: `SetTable`/`SetDatabase` rebuild the whole builder state by design and wipe every pick, so the heuristic re-arms against an empty slot rather than an explicit one (those reset and re-arm paths are covered by the unit tests in `logsQueryBuilderHooks.test.ts`). What does re-fire the effect is changing the pick itself, since the selected columns are in its dependency list. The test now gives that re-fired effect a window to mis-fire and asserts the explicit `service` pick isn't clobbered back to `level`.

tests/e2e/builder/builderQueries.spec.ts
adamyeats: You're right, this helper still had the blind type-and-Enter commit that `pickBuilderSelect` was hardened against earlier in the branch. It now waits for the exact option to appear and clicks it instead.

tests/e2e/builder/builderQueries.spec.ts
adamyeats: `Date.UTC` months are zero-indexed, so `2` is March here and the assertion does pass against the seeded 2024-03-15 row. That said, it reads like an off-by-one to anyone who doesn't have the zero-indexing loaded, so I've swapped it for `Date.parse('2024-03-15T10:09:00.000Z')`, which says what it means.

tests/e2e/variables/variableEditor.spec.ts
adamyeats: Fixed the same way as in `builderQueries.spec.ts`: it now waits for the option to appear and clicks it, rather than pressing Enter and hoping the list has caught up.

Worth flagging that this one was the more likely of the two to bite. `pickEditorSelect` drives both the `Variable type` select and the schema pickers, and the schema pickers populate their options from `/api/ds/query` (databases, then tables, then columns). The `Column values` test chains four picks in a row, each of the last three waiting on a network round trip, so a blind Enter had a real window in which to commit against a stale or empty list.

One wrinkle surfaced while verifying the fix, which is why the final version matches on a prefix rather than an exact name. My first attempt used `{ name: value, exact: true }` and timed out on the very first pick:

```
Error: locator.click: Test timeout of 30000ms exceeded.
Call log:
  - waiting for getByRole('option', { name: 'List databases', exact: true }).first()
```

The query-type options are defined with descriptions (`CHVariableSupport.tsx`):

```ts
{ label: 'List databases', value: 'databases', description: 'All databases on the server' },
```

and Grafana's Select folds the description into the option's accessible name, so the real name is `List databases All databases on the server`. The schema-picker options have no descriptions, so exact matching worked there but not here. The helper now anchors on the value as a word-boundary prefix (`^List databases\b`), which matches both shapes, with the interpolated value regex-escaped so a future label containing regex metacharacters cannot silently change the pattern.

tests/fixtures/seed.sql
adamyeats: Agreed, and thanks for spotting it. All five remaining fixtures (`seed.sql`, `map_events.sql`, `otel_logs_v0151.sql`, `simple_aggregate_functions.sql`, `trace_id_ts.sql`) now drop and recreate their tables on load, matching the pattern `trace_spans.sql` already used.

Verified by running the e2e-data-loader a second time against a live stack: row counts stayed at their single-seed values (events 10, map_events 6, otel_logs_v151 5, simple_aggregate_events 5, trace_ts_spans 6) and the suite stayed green afterwards.

adamyeats:
Superseded by #2122, #2123 and #2124 — split for reviewability. Verified the split carries every change from this branch (same content, US-spelling convention adjustment only). Closing in favour of the split.

PR #2075, opened 2026-07-18
fix(traces): trace-ID lookup, companion-table validation, and span-link fixes

adamyeats:
## Type of Change

- [x] 🐛 Bug Fix

---

### What is the bug?

Three trace defects from the v4.18.0/v4.19.0 cycles, all in the trace-ID lookup optimisation and span-link flow (#1786, #1853, #1890, #1919, #1920, #1937, #1992, #1994 all reworked this area), one commit each:

1. **Trace-ID lookup aliases shadow table columns.** `generateTraceIdQuery` emitted `WITH '<id>' as trace_id, ... trace_start, ... trace_end`. On tables with physical columns of those names (common on non-OTel tables, and #1791's heuristics auto-map `trace_id`), the ClickHouse alias shadows the column: `WHERE traceID = trace_id` becomes a tautology and the trace view renders every span in the time window mislabelled with the searched trace id (reproduced against live ClickHouse 24.8). The aliases are now `__gf_`-prefixed.
2. **Companion-table check validates a different suffix than the query uses.** `hasTraceTimestampTable` probed the table named by the datasource config suffix, while the generated SQL and the deep-link transform use the per-query `meta.traceTimestampTableSuffix`. When a saved query's baked suffix differs from the current config, the check passed against one table and the query hit another (UNKNOWN_TABLE). The check now validates the same suffix the query will use.
3. **A followed span link permanently overrides the trace ID.** Grafana injects a one-shot top-level `query` field when a span link is followed, but nothing ever cleared it, so the pane stayed pinned to the linked trace: typing a different trace ID showed the new ID in the UI while the data stayed on the old trace, and saving persisted the stale field. The field is now stripped on the first user edit, restoring one-shot semantics.

### How to reproduce

Each fix's regression tests document the exact scenario (`sqlGenerator.test.ts`, `CHDatasource.test.ts`, `utils.test.ts`, `CHQueryEditor.test.tsx`). Representative example for 1: non-OTel traces table `t(trace_id, span_id, start_time, ...)` with companion `t_trace_id_ts`; open a trace via "View trace" and observe spans from unrelated traces rendered under the opened trace id before this PR.

- **ClickHouse version**: any supported
- **Grafana version**: any supported

---

## Please check that:

Supersedes #2046, #2047 and #2048 (same fixes, previously one PR each, closed in favour of this bundle). Commits are kept separate, one per fix. The renamed WITH aliases in fix 1 are generated per execution and never persisted, so no saved dashboards are affected.

Found by an automated audit of regressions introduced while integrating PRs across the v4.18.0 and v4.19.0 release cycles (range v4.17.0..main). Related PRs: #1786, #1853, #1994, #1890, #1791.

### DIFF
--- a/src/data/CHDatasource.test.ts
+++ b/src/data/CHDatasource.test.ts
@@ -2630,6 +2630,30 @@ describe('ClickHouseDatasource', () => {
 
       await expect(ds.hasTraceTimestampTable('default', 'traces')).resolves.toBe(true);
     });
+
+    it('prefers an explicit suffix argument over the configured one', async () => {
+      // A saved query can bake a meta.traceTimestampTableSuffix that differs
+      // from the current datasource config. The check must probe the companion
+      // the generated SQL will reference (the query's suffix), and the two
+      // suffixes must not share a cache entry.
+      const ds = cloneDeep(mockDatasource);
+      ds.settings = {
+        ...ds.settings,
+        jsonData: {
+          ...ds.settings.jsonData,
+          traces: { ...(ds.settings.jsonData.traces || {}), traceTimestampTableSuffix: '_idx_ts' },
+        },
+      };
+      jest.spyOn(ds, 'fetchTables').mockResolvedValue(['traces', 'traces_saved_ts']);
+      const columnsSpy = jest.spyOn(ds, 'fetchColumns').mockResolvedValue(timestampColumns());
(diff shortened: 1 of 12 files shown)


itsgareth (approved):
Nice work, all looks good to me!

PR #2076, opened 2026-07-18
fix(config): v1/v2 editor validation, Map keys switch, and feature toggle fixes

adamyeats:
## Type of Change

- [x] 🐛 Bug Fix

---

### What is the bug?

Four config editor defects from the v4.18.0/v4.19.0 cycles, spanning the parallel v1 and v2 editors and the flag plumbing they share, one commit each:

1. **The v1 classic-mode Map keys discovery switch ignores its stored value.** #1885 wired `enableMapKeysDiscovery` into three of the four `QuerySettingsConfig` call sites, missing the v1 classic-mode block (the default editor), so the switch always rendered checked and the setting could never be re-enabled from that UI. The missing `value` prop is now passed.
2. **The v2 Username "required" error never clears.** `onBlur` set the error unconditionally, but the clearing logic lived behind the validation API, which is absent by default, so the error persisted for the whole session after one blur on the empty field. Errors now clear eagerly based on the field value, mirroring the `ServerAndEncryptionSection` pattern from #1831.
3. **The v2 config editor became unreachable on Grafana 11.6-12.2.** #1921 replaced the legacy `config.featureToggles` read with OpenFeature, but Grafana only registers that provider from 12.3, so the no-op provider always returned false and the ini toggle stopped working on older Grafana (worked in v4.18.0, `plugin.json` still supports >=11.6.0). The legacy toggle value is now passed as the OpenFeature default, keeping 12.3+ behaviour identical.
4. **Enabling `clickHouseConfigValidation` disabled v1 inline validation entirely.** #1995's structural fallback for the host/port required errors was keyed on `validation` truthiness, but the toggle-on path creates a local no-op ValidationAPI that nothing ever invokes on Grafana < 13.1, so no inline errors appeared at all. The structural check is now unconditional.

### How to reproduce

Each fix's regression tests document the exact scenario (`CHConfigEditor.test.tsx`, `DatabaseCredentialsSection.test.tsx`, `ConfigEditor.test.tsx`). Representative example for 3: Grafana 12.2 with `feature_toggles.newClickhouseConfigPageDesign = true`; before this PR the v1 editor renders silently, after it the v2 editor renders as it did in v4.18.0.

- **ClickHouse version**: any supported
- **Grafana version**: any supported (fix 3 specific to 11.6-12.2)

---

## Please check that:

Supersedes #2059, #2060, #2061 and #2070 (same fixes, previously one PR each, closed in favour of this bundle). Commits are kept separate, one per fix.

Found by an automated audit of regressions introduced while integrating PRs across the v4.18.0 and v4.19.0 release cycles (range v4.17.0..main). Related PRs: #1885, #2014, #1769, #1831, #1995, #1921.

### DIFF
--- a/src/views/CHConfigEditor.test.tsx
+++ b/src/views/CHConfigEditor.test.tsx
@@ -190,6 +190,44 @@ describe('ConfigEditor', () => {
     );
   });
 
+  describe('classic mode map keys discovery switch', () => {
+    // dialTimeout keeps the collapsible "Additional settings" section
+    // initially open so the query settings render.
+    const overrides: Partial<CHConfig> = { dialTimeout: '10', enableMapKeysDiscovery: false };
+
+    const getMapKeysSwitch = (): HTMLInputElement => {
+      const labelText = screen.getByText(allLabels.components.Config.QuerySettingsConfig.enableMapKeysDiscovery.label);
+      // The switch is inside a grafana-ui Field: the input lives in a sibling
+      // of the label element, under the shared field container.
+      const input = labelText.closest('label')?.parentElement?.parentElement?.querySelector('input[type="checkbox"]');
+      if (!(input instanceof HTMLInputElement)) {
+        throw new Error('map keys discovery switch not found');
+      }
+      return input;
+    };
+
+    it('reflects a stored enableMapKeysDiscovery=false', () => {
+      render(<ConfigEditor {...mockConfigEditorProps(overrides)} />);
+      expect(getMapKeysSwitch()).not.toBeChecked();
+    });
+
+    it('writes true when toggled back on', () => {
+      const props = mockConfigEditorProps(overrides);
+      render(<ConfigEditor {...props} />);
+
+      (props.onOptionsChange as jest.Mock).mockClear();
+      fireEvent.click(getMapKeysSwitch());
+
+      expect(props.onOptionsChange).toHaveBeenCalledWith(
+        expect.objectContaining({
+          jsonData: expect.objectContaining({
+            enableMapKeysDiscovery: true,
+          }),
+        })
+      );
+    });
+  });
+
   it('persists the single-table logs trace correlation setting', () => {
     const props = mockConfigEditorProps({
       configMode: 'single-table',
--- a/src/views/CHConfigEditor.test.tsx
+++ b/src/views/CHConfigEditor.test.tsx
@@ -235,16 +235,8 @@ describe('ConfigEditor', () => {
     // initially open so the query settings render.
     const overrides: Partial<CHConfig> = { dialTimeout: '10', enableMapKeysDiscovery: false };
 
-    const getMapKeysSwitch = (): HTMLInputElement => {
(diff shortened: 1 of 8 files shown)


src/views/CHConfigEditor.tsx
```diff
@@ -96,15 +96,16 @@ export const ConfigEditor: React.FC<ConfigEditorProps> = (props) => {
...
+  // The structural empty check must run regardless of the validation plumbing:
+  // with the clickHouseConfigValidation toggle off `validation` is undefined,
+  // and with it on the local ValidationAPI only gains errors once Grafana
+  // calls validate() (13.1+). Either way an empty required host/port field
+  // must surface an inline error immediately, as it did in v4.17.0, with any
+  // API-driven error message taking precedence when present.
+  const hostInvalid = Boolean(fieldErrors.host) || !jsonData.host;
+  const hostError = fieldErrors.host || (jsonData.host ? undefined : labels.serverAddress.error);
+  const portInvalid = Boolean(fieldErrors.port) || !jsonData.port;
+  const portError = fieldErrors.port || (jsonData.port ? undefined : labels.serverPort.error);
```
akkikumar72, line 108: I couldn't find any path where fieldErrors.host differs from the empty-check (no validate() calls pre-13.1; on 13.1+ only this file's validator writes these errors, same strings). Keep the precedence as future-proofing, or simplify? If simplifying, the "prefers the validation API error message" test goes with it.

src/views/CHConfigEditor.test.tsx
```diff
@@ -190,6 +230,44 @@ describe('ConfigEditor', () => {
...
+    const getMapKeysSwitch = (): HTMLInputElement => {
+      const labelText = screen.getByText(allLabels.components.Config.QuerySettingsConfig.enableMapKeysDiscovery.label);
+      // The switch is inside a grafana-ui Field: the input lives in a sibling
+      // of the label element, under the shared field container.
+      const input = labelText.closest('label')?.parentElement?.parentElement?.querySelector('input[type="checkbox"]');
+      if (!(input instanceof HTMLInputElement)) {
+        throw new Error('map keys discovery switch not found');
+      }
+      return input;
+    };
```
akkikumar72: This DOM traversal (closest('label')?.parentElement?.parentElement?.querySelector('input[type="checkbox"]')) couples the test to grafana-ui's internal Field markup a grafana-ui upgrade would fail it with the opaque "switch not found" throw. The repo already has a convention for exactly this: enableRowLimit puts a data-testid on the switch and the test uses getByTestId (the labels object already carries testid fields). Suggest adding a testid to the QuerySettingsConfig switch and using that here instead.

akkikumar72 (commented):
Rest looks good

src/views/CHConfigEditor.tsx
adamyeats: Kept as is for now. You're right that today the two never diverge, but on 13.1+ core calls validate() on submit, and the precedence is what lets any future API-driven message (for example a malformed-host error) win over the generic required text. Happy to simplify if you feel the extra branch isn't earning its keep.

akkikumar72 (approved):
Nice job 🎖️

PR #1832, opened 2026-05-11
feat(config): introduce single-table mode for logs and traces configuration

fleon:
## Type of Change

- [x] 🚀 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧹 Refactor / Chore

---

### What is this feature?

Adds a single-source configuration mode for ClickHouse data sources. Users can choose whether the data source should target all databases or focus on one logs or traces table, then configure the signal type and table schema for that focused setup.

### Why is this feature needed?

Some ClickHouse data sources are created for a specific logs or traces table. A focused mode reduces setup friction for those users and lets the query experience be tailored to the selected signal source.

### Who is this feature for?

Users configuring ClickHouse data sources for a known logs or traces table, especially OpenTelemetry-style datasets where the database, table, and schema mappings are known up front.

### How to test this feature

1. Open a ClickHouse data source configuration page.
2. Select `Single source` in the configuration mode section.
3. Choose `Logs` or `Traces`, configure the database/table and schema fields, then save and test the data source.
4. Reopen the configuration page and verify the selected mode, signal type, and schema settings are preserved.
5. Switch back to `All databases` and verify the full logs and traces configuration remains available.

---

## Screenshots / Videos

| | Before | After |
|-|------ | ----- |
|Config v1|    CleanShot 2026-05-12 at 16 22 34@2x   |   CleanShot 2026-05-12 at 16 20 42@2x    |
|Config v2| CleanShot 2026-05-15 at 15 13 35@2x | CleanShot 2026-05-15 at 15 11 29@2x |

---

## Please check that:

- [x] Tests for this change have been added/updated.
- [x] Documentation has been added/updated (where applicable).

## Special notes for your reviewer

No special notes.

PR #2079, opened 2026-07-18
fix(adhoc): single-table sources, stateless Map keys, and pre-24.3 discovery

adamyeats:
## Type of Change

- [x] 🐛 Bug Fix

---

### What is the bug?

Three ad-hoc filter defects from the v4.18.0/v4.19.0 cycles (the #1793/#1885/#1993 Map-key work plus single-table mode #1832), one commit each:

1. **Key discovery ignores single-table configuration.** Discovery resolved its source solely from `jsonData.defaultDatabase` (coerced to the literal `default`), but single-table mode stores the database and table under `jsonData.logs`/`jsonData.traces` and hides the classic field, so the key dropdown listed the wrong database and picked keys never matched the compact query's FROM table. Discovery now falls back to the single-table logs (else traces) source when the classic default is unset, with `$clickhouse_adhoc_query` and explicit defaults keeping precedence.
2. **Saved Map-key filters generate wrong SQL on dashboard load.** `escapeKey` only rendered dotted Map keys as bracket access after `setMapColumns` ran, which happens exclusively inside `getTagKeys` (lazy, on dropdown open). On a fresh dashboard load every panel errored with UNKNOWN_IDENTIFIER until the user opened the filter dropdown once. Newly minted keys now use a self-describing bracket form (`table.mapcol['key']`) that applies statelessly, and legacy dotted keys keep working via the existing discovery path.
3. **Map-key discovery breaks on ClickHouse before 24.3.** The #1993 bounded probe read the `.keys` sub-column from a subquery, which the old analyzer (default before 24.3, or `allow_experimental_analyzer=0`) rejects with UNKNOWN_IDENTIFIER, and every call site swallows the error, so Map-key filters and dropdowns silently vanished on 22.7-24.2 (reproduced on 23.8 vs 24.8). The probe now uses the `mapKeys()` function, which both analyzers resolve.

### How to reproduce

Each fix's regression tests document the exact scenario (`CHDatasource.test.ts`, `adHocFilter.test.ts`). Representative example for 2: add an ad-hoc filter on a custom Map column key, save the dashboard, reload it; before this PR every panel errors until the filter key dropdown is opened once.

- **ClickHouse version**: any supported (fix 3 specific to pre-24.3 or old-analyzer configs)
- **Grafana version**: any supported

---

## Please check that:

Supersedes #2052, #2055 and #2064 (same fixes, previously one PR each, closed in favour of this bundle). Commits are kept separate, one per fix. Draft because fix 2 introduces a new user-visible key format in the ad-hoc dropdown (bracket form instead of dotted), which is a design decision worth explicit sign-off, fixes 1 and 3 are straightforward. One cross-fix interaction was resolved while bundling: fix 2's key-escaping test stubbed the probe with the pre-fix-3 SQL shape and was updated to the `mapKeys()` form.

Found by an automated audit of regressions introduced while integrating PRs across the v4.18.0 and v4.19.0 release cycles (range v4.17.0..main). Related PRs: #1793, #1885, #1993, #1832, #1841.

### DIFF
--- a/src/data/CHDatasource.ts
+++ b/src/data/CHDatasource.ts
@@ -1579,7 +1579,10 @@ export class Datasource
 
     // Replace each top-level Map-column entry with one entry per discovered
     // map key. If probing returned nothing (empty set, stripped by the
-    // filter), keep the original entry as a no-op fallback.
+    // filter), keep the original entry as a no-op fallback. Keys are minted
+    // in explicit bracket form (`col['key']`) so that filter application is
+    // stateless: a saved filter must render correct SQL on fresh dashboard
+    // load, before this method has run and populated the Map-column caches.
     const expandedKeyByCol = new Map<string, string[]>();
     for (const p of probed) {
       if (p.keys.length > 0) {
@@ -1601,7 +1604,7 @@ export class Datasource
         return;
       }
       for (const k of mapKeys) {
-        expanded.push({ text: `${baseText}.${k}` });
+        expanded.push({ text: `${baseText}['${escapeCHStringLiteral(k)}']` });
       }
     });
     return expanded;
@@ -1700,7 +1703,9 @@ export class Datasource
     // distinct Map values rather than stringified Map objects (which the
     // Grafana frame layer renders as `[object Object]`). The map key is
     // escaped for CH string-literal embedding so that keys containing `'`
-    // or `\` produce valid SQL.
+    // or `\` produce valid SQL. Keys minted by getTagKeys arrive already in
+    // bracket form (`mapCol['key']`, literal body pre-escaped) and pass
+    // through unchanged as valid bracket access.
     const mapAccess = this.asMapAccess(col, source);
     const selectExpr = mapAccess ? `${mapAccess.column}['${escapeCHStringLiteral(mapAccess.key)}']` : col;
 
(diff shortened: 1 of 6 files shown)


src/data/CHDatasource.ts
```diff
@@ -1793,12 +1823,14 @@ export class Datasource
   private getTagSource() {
     // @todo https://github.com/grafana/grafana/issues/13109
     const ADHOC_VAR = '$clickhouse_adhoc_query';
-    const defaultDatabase = this.getDefaultDatabase();
     let source = getTemplateSrv().replace(ADHOC_VAR);
-    if (source === ADHOC_VAR && isEmpty(defaultDatabase)) {
-      return { type: TagType.schema, source: undefined };
+    if (source === ADHOC_VAR) {
+      const defaultSource = this.getDefaultAdhocSource();
+      if (!defaultSource) {
```
akkikumar72, line 1838: `if (!defaultSource)` can't actually be true: `getDefaultAdhocSource()` always
falls through to `getDefaultDatabase()`, which defaults to the literal
`'default'` rather than returning falsy. This mirrors the dead
`isEmpty(defaultDatabase)` check it replaced, so it's not new, just worth a
follow-up cleanup if you touch this again.

akkikumar72 (approved):
Reviewed and tested this live against  Grafana + ClickHouse 26.3.9 stack and it looks good. Minor nit added but its not blocking this PR

src/data/CHDatasource.ts
adamyeats: Resolving in another PR.

adamyeats: Follow-up cleanup opened as #2104. It is based on main, so it removes the check in its current `isEmpty(defaultDatabase)` form, along with the downstream `tagSource.source === undefined` branch in `fetchTags()` that only that check could reach. Whichever of the two PRs lands second will need a small conflict resolution here.

PR #2080, opened 2026-07-18
feat(dashboards): add database selector variable to OTel dashboards

AndreZiviani:
## What
Adds a `database` template variable (query type) to the OpenTelemetry **Logs**, **Traces** and **Service** dashboards, and qualifies every `otel_logs` / `otel_traces` table reference with `${database}`.

Each dashboard's variable lists only the databases that actually contain the tables it queries, via `system.tables`:
- **Logs:** `SELECT DISTINCT database FROM system.tables WHERE name = 'otel_logs' ORDER BY database`
- **Traces:** same, `WHERE name = 'otel_traces'`
- **Service:** `WHERE name IN ('otel_logs', 'otel_traces') ... HAVING count(DISTINCT name) = 2` (needs both)

## Why
Today these dashboards query the datasource's default database only. This lets a single datasource serve otel tables that live in different databases (e.g. one per environment) without hand-editing every panel.

The **Service Map** keeps its upstream query logic — only the table name is qualified.

---
- #2083 — multi-value quoting

### DIFF
--- a/docs/sources/dashboards.md
+++ b/docs/sources/dashboards.md
@@ -32,11 +32,13 @@ The dashboards expect the standard table layout produced by the [ClickHouse expo
 - A logs table named `otel_logs`.
 - A traces table named `otel_traces`.
 
-The three Map-schema dashboards (Logs Explorer, Traces Explorer, and the single-service deep dive) reference both tables by their bare names in raw SQL, so the data source's **Default database** setting needs to point at the database that holds them.
+All four dashboards expose a **Database** template variable and qualify every table reference as `${database}.otel_logs` / `${database}.otel_traces`, so you pick the database that holds the OTel tables from the dashboard itself instead of relying on the data source's **Default database** setting.
 
-The JSON-schema Logs Explorer instead exposes a **Database** selector variable and qualifies its queries as `${database}.otel_logs`, so it can run against any database without changing the data source default. Its panels only need `otel_logs`; the `otel_traces` table is required solely by the optional deployment annotation, which is **disabled by default**, so a logs-only database works out of the box.
+The three Map-schema dashboards (Logs Explorer, Traces Explorer, and the single-service deep dive) link to each other and forward the selected database between them, so their **Database** variable lists only databases that hold **both** `otel_logs` and `otel_traces`. Splitting logs and traces across separate databases is not supported by these three: the selector would list nothing rather than offer a database whose panels cannot all resolve.
 
-If you renamed the OTel exporter tables, the dashboards do not pick the new names up automatically; either restore the standard names or duplicate the dashboards and edit the SQL.
+The JSON-schema Logs Explorer lists databases whose `otel_logs.ResourceAttributes` column uses the native `JSON` type. Its panels only need `otel_logs`; the `otel_traces` table is required solely by the optional deployment annotation, which is **disabled by default**, so a logs-only database works there.
+
+Both selectors read `system.tables` / `system.columns`, so the data source user needs read access to them. If you renamed the OTel exporter tables, the dashboards do not pick the new names up automatically; either restore the standard names or duplicate the dashboards and edit the SQL.
 
 The dashboards rely on the columns the exporter ships with: `Timestamp`, `Body`, `SeverityText`, `ServiceName`, `TraceId`, `SpanId`, `ResourceAttributes` for logs; `TraceId`, `ServiceName`, `SpanName`, `Timestamp`, `Duration`, `StatusCode`, `SpanKind`, `ParentSpanId`, `SpanAttributes`, `ResourceAttributes` for traces.
 
@@ -52,7 +54,7 @@ Below the overview, a per-service row repeats once per service in the **Service*
 
 Both per-service panels have an **Open in Explore** link in the panel header that pre-fills a matching query for that row's service in Explore. The Log Volume link drops you into a time-series query; the Log Samples link drops you into the logs visualization with the same column projection as the dashboard panel. Because the per-service row repeats once per selected service, each panel-header link is scoped to its row's `$service` value.
 
-Filter variables: **Service** (multi, defaults to top 10 by volume), **Level** (multi), **Search** (textbox; passes through to `positionCaseInsensitive(Body, ...)`, a case-insensitive substring match).
+Filter variables: **Database** (selects the database that holds the OTel tables; needs read access to `system.tables`), **Service** (multi, defaults to top 10 by volume), **Level** (multi), **Search** (textbox; passes through to `positionCaseInsensitive(Body, ...)`, a case-insensitive substring match).
 
 Annotations: deployment markers derived from `service.version` changes in `otel_traces` over 30-second buckets.
 
@@ -72,7 +74,7 @@ Below trace search, a per-service row repeats once per service: a 1-in-500 sampl
 
 Each per-service panel has an **Open in Explore** link in its panel header that pre-fills a matching query in Explore.
 
-Filter variables: **Service** (multi), **Operation** (multi), **Status** (multi), **Min Duration (ms)** (textbox; leave blank for no minimum, passes through `$__conditionalAll`), **Search** (textbox; free-text match on SpanName applied to every trace panel).
+Filter variables: **Database** (selects the database that holds the OTel tables), **Service** (multi), **Operation** (multi), **Status** (multi), **Min Duration (ms)** (textbox; leave blank for no minimum, passes through `$__conditionalAll`), **Search** (textbox; free-text match on SpanName applied to every trace panel).
 
 ## OpenTelemetry Service Dashboard
 
@@ -88,7 +90,7 @@ Single-service deep dive, broken out by SpanKind. Top to bottom:
 
 Click any operation in any Slowest Operations or Top Errors table to open the matching trace search in OpenTelemetry Traces Explorer.
 
-Filter variables: **Service** (single-select), **Operation** (multi), **Search** (textbox; free-text match on SpanName for trace panels and on Body for the Logs panel).
+Filter variables: **Database** (selects the database that holds the OTel tables), **Service** (single-select), **Operation** (multi), **Search** (textbox; free-text match on SpanName for trace panels and on Body for the Logs panel).
 
 Each non-row panel has an **Open in Explore** link in its header that pre-fills the panel's underlying query in Explore. Useful when you want to refine the SQL, change visualization, or read log details with more horizontal space.
 
--- a/src/dashboards/__tests__/dashboards.test.ts
+++ b/src/dashboards/__tests__/dashboards.test.ts
@@ -147,4 +147,21 @@ describe('OTel dashboards', () => {
       expect(content).not.toMatch(/= '\$\{?\w+\}?'/);
     });
   });
+
+  describe('database variable', () => {
+    it.each(otelDashboards)('%s exposes a "database" query variable', (filename) => {
+      const dashboard = JSON.parse(fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8')) as {
+        templating?: { list?: Array<{ name?: string; type?: string }> };
+      };
+      const variable = dashboard.templating?.list?.find((v) => v.name === 'database');
+      expect(variable).toBeDefined();
+      expect(variable?.type).toBe('query');
+    });
+
+    it.each(otelDashboards)('%s qualifies every otel table with ${database}', (filename) => {
+      const content = fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8');
+      // No bare (unqualified) references to the otel tables should remain.
+      expect(content).not.toMatch(/(?:FROM|JOIN)\s+otel_(?:logs|traces)\b/);
+    });
+  });
 });
--- a/src/dashboards/__tests__/dashboards.test.ts
+++ b/src/dashboards/__tests__/dashboards.test.ts
@@ -149,6 +149,46 @@ describe('OTel dashboards', () => {
   });
 
   describe('database variable', () => {
+    // The three Map-schema dashboards link to each other with includeVars, so a database
+    // selected on one is forwarded to the others. Their selectors must therefore offer the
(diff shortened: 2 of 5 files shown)


src/dashboards/otel-logs-explorer.json
```diff
@@ -291,6 +291,25 @@
...
+        "definition": "",
+        "hide": 0,
+        "includeAll": false,
+        "label": "Database",
+        "multi": false,
+        "name": "database",
+        "options": [],
+        "query": {
+          "editorType": "sql",
+          "rawSql": "SELECT DISTINCT database FROM system.tables WHERE name = 'otel_logs' ORDER BY database"
```
adamyeats: This lists databases that only contain `otel_logs`, but the deployment annotation on this dashboard queries `${database}.otel_traces` and is enabled by default, so selecting a logs-only database fails the annotation query on every refresh (reproduced live, ClickHouse error 60 and a 400 on each 30s cycle). Consider pairing the check with `otel_traces` as the service dashboard does, or leaving the annotation off by default as main now does.

src/dashboards/otel-service-dashboard.json
```diff
@@ -420,7 +439,7 @@
             "uid": "${datasource}"
           },
           "editorType": "sql",
-          "rawSql": "SELECT SpanName AS Operation, count() AS Count, quantile(0.5)(Duration) AS P50, quantile(0.99)(Duration) AS P99 FROM otel_traces WHERE $__timeFilter(Timestamp) AND ServiceName = '${service}' AND SpanKind = 'Server' AND $__conditionalAll(SpanName IN ($operation), $operation) AND $__conditionalAll(hasToken(SpanName, '$search'), $search) GROUP BY SpanName ORDER BY P99 DESC LIMIT 20",
+          "rawSql": "SELECT SpanName AS Operation, count() AS Count, quantile(0.5)(Duration) AS P50, quantile(0.99)(Duration) AS P99 FROM ${database}.otel_traces WHERE $__timeFilter(Timestamp) AND ServiceName = '${service}' AND SpanKind = 'Server' AND $__conditionalAll(SpanName IN ($operation), $operation) AND $__conditionalAll(hasToken(SpanName, '$search'), $search) GROUP BY SpanName ORDER BY P99 DESC LIMIT 20",
```
adamyeats: The drilldown links on this panel and its siblings (lines 414, 688, 773, 1088 and 1375) carry the service, operation and datasource but not the selected database, so clicking through from a non-default database lands on the traces explorer with its own default and every panel silently shows no data (reproduced live). Consider appending `&var-database=${database}` to each link.

src/dashboards/__tests__/dashboards.test.ts
```diff
@@ -48,4 +48,21 @@ describe('OTel dashboards', () => {
...
+      expect(variable?.type).toBe('query');
+    });
+
+    it.each(otelDashboards)('%s qualifies every otel table with ${database}', (filename) => {
+      const content = fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8');
+      // No bare (unqualified) references to the otel tables should remain.
+      expect(content).not.toMatch(/(?:FROM|JOIN)\s+otel_(?:logs|traces)\b/);
+    });
+  });
 });
```
adamyeats, line 210: This only asserts the absence of bare references, so qualifying with anything other than `${database}.` would still pass. It might be worth also asserting that `${database}.otel_` appears at least once per dashboard.

src/dashboards/otel-logs-explorer.json
```diff
@@ -291,6 +291,25 @@
         "regex": "",
         "type": "datasource"
       },
+      {
+        "current": {},
+        "datasource": { "type": "grafana-clickhouse-datasource", "uid": "${datasource}" },
+        "definition": "",
```
adamyeats: nit: the sibling query variables set `definition` to the same SQL as `query.rawSql`, which is what the variable list in the dashboard settings displays. Consider mirroring that here for consistency (also applies to the copies in otel-traces-explorer.json and otel-service-dashboard.json).

adamyeats (changes requested):
This is now conflicting with main: #2077 reshaped the annotations (added a `customSql` field and disabled the logs-explorer one by default) and #2092 replaced `hasToken` with `positionCaseInsensitive` in the logs queries. On rebase the new `customSql` fields will also need `${database}.` qualification, and helpfully the new test regex will catch them if missed.

One small thing on the description: it says the variable queries `SELECT name FROM system.databases`, but the implementation queries `system.tables` filtered to the OTel tables, which is better. It might be worth updating the description to match, since the docs already describe the `system.tables` behaviour.

AndreZiviani:
Thanks @adamyeats — rebased over #2077 + #2092 and addressed everything:

- **customSql qualification:** #2077's new deployment-detection `customSql` annotation queries are now `${database}.otel_traces`-qualified in all three dashboards; the strengthened test catches any miss.
- **Logs-explorer database variable:** kept it logs-only (`WHERE name = 'otel_logs'`). After the #2077 rebase the deployment annotation is disabled by default and no enabled-by-default query in the logs dashboard touches `otel_traces`, so selecting a logs-only database no longer fails the 30s refresh. Requiring `otel_traces` here would wrongly hide logs-only databases from a logs dashboard — the service dashboard requires both because its panels query both.
- **Drill-through links:** appended `&var-database=${database}` to all 5 hand-built links (all in `otel-service-dashboard.json`).
- **Test:** now also asserts `${database}.otel_…` appears at least once per dashboard, proving qualification actually happened (not just that bare names are gone).
- **definition:** the `database` variable's `definition` now mirrors its `rawSql` in all three.
- Also corrected the PR description (it said `system.databases`; the implementation uses `system.tables`).

`tsc --noEmit` clean; dashboards suite green.

AndreZiviani:
Thanks @adamyeats — all addressed and rebased onto latest `main`: the new `customSql` annotation fields are `${database}`-qualified, the logs-explorer deployment annotation is disabled by default (so logs-only databases don't fail), the drill-through links now carry `&var-database=${database}`, the test asserts `${database}.otel_` appears per dashboard, and `definition` mirrors `rawSql`. Also corrected the description (`system.tables`, not `system.databases`). Re-requesting review.

src/dashboards/otel-logs-explorer.json
adamyeats: The annotation now stays off by default, and the new test pins that, so the refresh loop is covered.

The same database selection still reaches the traces explorer by another route. All three dashboards share the `clickhouse` and `otel` tags and carry a dashboard-level link with `includeVars: true`. This PR adds the `database` variable to all three, so the dropdown now forwards `var-database` to the traces explorer. Every table reference there is `${database}.otel_traces`, so a logs-only database leaves every panel empty. This path needs no opt-in from the user.

The service dashboard pairs the check with `otel_traces`. The same approach works here. If you remove `includeVars` from that link, the path also closes, but then no variable forwards at all.

adamyeats (changes requested):

adamyeats:
@AndreZiviani this branch is behind main and now conflicts in two files.

`src/dashboards/__tests__/dashboards.test.ts` conflicts because #2084 on main appended a `describe` block at the same point as this PR. The resolution is to keep both blocks.

`docs/sources/dashboards.md` needs more care. Main rewrote the paragraph to say only the JSON-schema dashboard has a Database selector, which this PR makes untrue. Main also renamed the search function from `hasToken` to `positionCaseInsensitive`, so neither side is correct on its own.

On ordering: this PR overlaps both #2083 and #2082. Against #2083 it conflicts in all three OTel dashboard JSON files, and against #2082 in two of them. It is worth landing this one after #2083 and before #2082.

One collision to watch: #2082 appends a different parameter to the same five data links in `otel-service-dashboard.json`. The correct resolution keeps both, as `&var-database=${database}&var-interval=${interval}`. No test covers link forwarding, so CI stays green if one is dropped.

AndreZiviani:
Rebased and force-pushed, and the drill-through leak is fixed.

**On the leak.** You were right, and it was worse than the drill-through: the logs explorer already broke on its own under a logs-only database, because its deployment annotation reads `${database}.otel_traces`.

The root cause was that the three selectors disagreed about what to offer — logs listed databases holding `otel_logs`, traces `otel_traces`, and only the service dashboard required both. So I adopted the service dashboard's query in all three:

```sql
SELECT database FROM system.tables WHERE name IN ('otel_logs', 'otel_traces')
GROUP BY database HAVING count(DISTINCT name) = 2 ORDER BY database
```

Every selector now offers the same set, so forwarding is valid by construction and `includeVars` stays — dropping it would have closed the path at the cost of forwarding nothing, as you noted.

The tradeoff, stated plainly in the docs: splitting logs and traces across separate databases is now unsupported for these three. The selector lists nothing rather than offering a database whose panels cannot all resolve. That seemed better than silently-empty panels, but say the word if you would rather keep it permissive.

I left the JSON-schema dashboard alone — its selector detects the JSON column *type*, which is a different question, and its traces annotation is off by default, so logs-only genuinely works there.

**On the docs conflict.** Rewrote the paragraph rather than picking a side: it now covers your `hasToken` → `positionCaseInsensitive` rename and the fact that all four dashboards have a Database selector.

**On the missing test.** Added two guards: the three variable queries must stay identical, and the drill-through links must keep `var-database`. Both fail if reverted — I checked, rather than just checking they pass.

One correction for the merge with #2082: the resolution is now

```
&var-database=${database}&var-interval=${interval:text}
```

not `${interval}` — see that PR for why. With the link guards on both sides, dropping either parameter now fails CI instead of going green.

src/dashboards/__tests__/dashboards.test.ts
```diff
@@ -147,4 +147,64 @@ describe('OTel dashboards', () => {
...
+      expect(new Set(queries).size).toBe(1);
+    });
+
+    it.each(linkedOtelDashboards)('%s only offers databases holding both otel tables', (filename) => {
+      const sql = databaseVariableSql(filename) ?? '';
+      expect(sql).toContain("name IN ('otel_logs', 'otel_traces')");
+      expect(sql).toContain('count(DISTINCT name) = 2');
+    });
+
+    it.each(linkedOtelDashboards)('%s carries the database through drill-through links', (filename) => {
```
adamyeats, line 182: nit: the drill-through guard is vacuous for `otel-logs-explorer.json` and `otel-traces-explorer.json`. Neither contains a `/d/otel-*?` URL, so the loop body never runs for them. Consider asserting a non-zero match count for the service dashboard, or scoping the test to it.

docs/sources/dashboards.md
```diff
@@ -32,11 +32,13 @@ The dashboards expect the standard table layout produced by the [ClickHouse expo
...
+All four dashboards expose a **Database** template variable and qualify every table reference as `${database}.otel_logs` / `${database}.otel_traces`, so you pick the database that holds the OTel tables from the dashboard itself instead of relying on the data source's **Default database** setting.
 
-The JSON-schema Logs Explorer instead exposes a **Database** selector variable and qualifies its queries as `${database}.otel_logs`, so it can run against any database without changing the data source default. Its panels only need `otel_logs`; the `otel_traces` table is required solely by the optional deployment annotation, which is **disabled by default**, so a logs-only database works out of the box.
+The three Map-schema dashboards (Logs Explorer, Traces Explorer, and the single-service deep dive) link to each other and forward the selected database between them, so their **Database** variable lists only databases that hold **both** `otel_logs` and `otel_traces`. Splitting logs and traces across separate databases is not supported by these three: the selector would list nothing rather than offer a database whose panels cannot all resolve.
```
adamyeats, line 37: nit: the intro sentence at line 26 still says the cross-links preserve the time range and datasource. After this PR they also carry the selected database, which is the point of the unified selector.

adamyeats (dismissed):

AndreZiviani:
Thanks @adamyeats — both addressed, and rebased onto latest main.

- The drill-through guard no longer passes vacuously: it still checks every cross-link it finds, and a separate assertion pins that the service dashboard has links to check, so deleting them fails rather than emptying the loop. Verified by removing `&var-database=${database}` from the five links and watching it fail.
- `docs/sources/dashboards.md` now says the cross-links carry the selected database alongside the time range and datasource.

The rebase also merged the required-schema paragraph with main's rewrite, so it covers all four dashboards: the three Map-schema ones list databases holding both tables (they forward the selection between them), while the JSON-schema one lists by attribute column type.

Dashboards suite green (57/57).

adamyeats:
@AndreZiviani #2095 and #2083 are on main now, so this branch conflicts in `src/dashboards/__tests__/dashboards.test.ts` and the three OTel dashboard JSONs. Could you rebase onto main? Both of your fixes check out, so I will re-approve as soon as CI is green.

AndreZiviani:
Rebased onto main.

The conflicts were all in the same shape: #2083 rewrote each query's variable interpolation to `:singlequote`, and this PR qualifies the same queries with `${database}.`. I checked that every one of the 37 JSON hunks was exactly that — this branch's side differed from the merge base only by the qualification — and then resolved each by taking main's rewritten query and re-applying the qualification to it, so neither change is lost. The test file keeps both `describe` blocks.

**One thing worth flagging.** While rebasing I found that two of this branch's commits had accidentally included my local scratch directories (a `dev-test/` harness and a jest cache, ~1000 files), added in one commit and deleted in a later one. The PR's net diff was clean, which is why nothing caught it, but the commits themselves carried the files. I rebuilt the four commits without them, so the SHAs changed by more than the rebase alone. Messages, authorship and content are otherwise unchanged, and the branch now touches only `src/dashboards/*` and `docs/sources/dashboards.md`.

Dashboards suite green (64/64), `tsc` clean, signatures verified.

src/dashboards/__tests__/dashboards.test.ts
adamyeats: Addressed in the 21 September push (d00b2dd). Resolving.

docs/sources/dashboards.md
adamyeats: Addressed in the 21 September push (d00b2dd). Resolving.

adamyeats (approved):
Checked the rebase against both threads and against the quoting guard from #2083. Both changes are present on every query and the diff touches only the dashboards and their docs. Approving.

PR #2081, opened 2026-07-18
feat(dashboards): make OTel dashboards importable via datasource input

AndreZiviani:
## What
Adds an `__inputs` datasource entry (`DS_GRAFANA_CLICKHOUSE_DATASOURCE`) plus `__requires` to the OpenTelemetry **Logs**, **Traces** and **Service** dashboards, and wires the existing `datasource` template variable's default to that input.

## Why
With a non-empty `__inputs` block Grafana prompts for a ClickHouse datasource when the dashboard JSON is imported, instead of silently binding to whichever datasource happens to resolve (this matches the pattern already used by `system-dashboards.json`). The `${datasource}` template variable is preserved, so the in-dashboard datasource switcher still works after import.

---
- #2083 — multi-value quoting

BEGIN_COMMIT_OVERRIDE
feat(dashboards): make OTel dashboards importable via datasource input (#2081)
END_COMMIT_OVERRIDE

### DIFF
--- a/src/dashboards/__tests__/dashboards.test.ts
+++ b/src/dashboards/__tests__/dashboards.test.ts
@@ -147,4 +147,27 @@ describe('OTel dashboards', () => {
       expect(content).not.toMatch(/= '\$\{?\w+\}?'/);
     });
   });
+
+  describe('importable', () => {
+    it.each(otelDashboards)('%s declares a datasource __input so import prompts for one', (filename) => {
+      const dashboard = JSON.parse(fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8')) as {
+        __inputs?: Array<{ name?: string; type?: string; pluginId?: string }>;
+      };
+      const dsInput = dashboard.__inputs?.find((i) => i.type === 'datasource');
+      expect(dsInput).toBeDefined();
+      expect(dsInput?.pluginId).toBe('grafana-clickhouse-datasource');
+    });
+
+    it.each(otelDashboards)('%s wires the datasource variable to the __input', (filename) => {
+      // On UI import Grafana substitutes the input token into current.value, so the datasource
+      // the user picks in the import prompt becomes the variable's selection.
+      const dashboard = JSON.parse(fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8')) as {
+        __inputs?: Array<{ name?: string }>;
+        templating?: { list?: Array<{ type?: string; current?: { value?: string } }> };
+      };
+      const inputName = dashboard.__inputs?.[0]?.name;
+      const dsVar = dashboard.templating?.list?.find((v) => v.type === 'datasource');
+      expect(dsVar?.current?.value).toBe(`\${${inputName}}`);
+    });
+  });
 });
--- a/src/dashboards/__tests__/dashboards.test.ts
+++ b/src/dashboards/__tests__/dashboards.test.ts
@@ -162,10 +162,10 @@ describe('OTel dashboards', () => {
       // On UI import Grafana substitutes the input token into current.value, so the datasource
       // the user picks in the import prompt becomes the variable's selection.
       const dashboard = JSON.parse(fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8')) as {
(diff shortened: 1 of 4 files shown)


src/dashboards/otel-logs-explorer.json
```diff
@@ -1,4 +1,58 @@
...
+      "pluginId": "grafana-clickhouse-datasource",
+      "pluginName": "ClickHouse"
+    }
+  ],
+  "__requires": [
+    {
+      "type": "grafana",
+      "id": "grafana",
+      "name": "Grafana",
+      "version": "9.0.1"
```
adamyeats: Consider matching the plugin's `grafanaDependency` (`>=11.6.0-0`) here rather than 9.0.1, which understates the real floor. The sibling `system-dashboards.json` declares `11.2.0-pre`, and the same applies to the other two dashboards.

src/dashboards/otel-logs-explorer.json
```diff
@@ -1,4 +1,58 @@
...
+    },
+    {
+      "type": "datasource",
+      "id": "grafana-clickhouse-datasource",
+      "name": "ClickHouse",
+      "version": "2.0.0"
```
adamyeats: It might be worth declaring a current plugin version here, since `system-dashboards.json` says `4.3.2` and these dashboards only shipped in 4.20, so 2.0.0 reads as misleading even though `__requires` is informational. Same in the other two dashboards.

src/dashboards/otel-logs-explorer.json
```diff
@@ -279,7 +333,11 @@
   "templating": {
     "list": [
       {
-        "current": {},
+        "current": {
+          "selected": false,
+          "text": "ClickHouse",
```
adamyeats: nit: Grafana's exporter writes `text: ''` with `selected: true` for templated datasource variables, which avoids briefly showing "ClickHouse" when the imported datasource has a different name. Same in the other two dashboards.

src/dashboards/__tests__/dashboards.test.ts
```diff
@@ -48,4 +48,27 @@ describe('OTel dashboards', () => {
...
+    });
+
+    it.each(otelDashboards)('%s wires the datasource variable to the __input', (filename) => {
+      // On UI import Grafana substitutes the input token into current.value, so the datasource
+      // the user picks in the import prompt becomes the variable's selection.
+      const dashboard = JSON.parse(fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8')) as {
+        __inputs?: Array<{ name?: string }>;
+        templating?: { list?: Array<{ type?: string; current?: { value?: string } }> };
+      };
+      const inputName = dashboard.__inputs?.[0]?.name;
```
adamyeats: Consider finding the input by `type === 'datasource'` as the previous test does, so this assertion survives another input ever being added ahead of it.

AndreZiviani:
Thanks @adamyeats — addressed all four points and rebased onto `main`:

- `grafanaDependency` `9.0.1` → `>=11.6.0-0` (matches `src/plugin.json`) in all three dashboards.
- `__requires` plugin version `2.0.0` → `4.20.0` (from `package.json`) in all three.
- Datasource template variable `current` → `text: ""`, `selected: true` in all three.
- The test now finds the input by `type === 'datasource'` instead of positionally, so it survives another input being added ahead of it.

`tsc --noEmit` clean; dashboards suite green (29/29).

AndreZiviani:
Thanks @adamyeats — addressed: `grafanaDependency` `>=11.6.0-0`, `__requires` plugin version `4.20.0`, the datasource template variable `current` now uses `text: ""` / `selected: true`, and the test finds the input by `type === 'datasource'`. Rebased onto latest `main`. Re-requesting review.

adamyeats:

This is the only one of the four dashboard PRs that does not overlap the others, so it is worth landing first. I checked the pairwise merges and it touches no JSON region that #2080, #2082 or #2083 also change.

AndreZiviani:
Rebased onto latest main and force-pushed. The only conflict was the `describe` block in `dashboards.test.ts`; kept both, as you said.

No content change — this one needed nothing else. Happy for it to go first.

adamyeats (approved):

PR #2082, opened 2026-07-18
feat(dashboards): add interval variable to control panel min interval

AndreZiviani:
## What
Adds an `interval` template variable (interval type, default `auto`) to the OpenTelemetry **Logs**, **Traces** and **Service** dashboards, and sets it as the **Min interval** query option (`"interval": "${interval}"`) on every panel that buckets by `$__interval_s`.

## Why
Lets users control the time-bucket granularity of the timeseries / heatmap / stat panels from the dashboard, instead of being locked to Grafana's auto interval derived from panel width.

## Drill-through behaviour

The five hand-built drill-through links on the service dashboard forward `&var-interval=${interval}`, which resolves to the bucket in effect when you click. So a drill-through from the default `auto` state pins that concrete bucket (for example `30s`) on the target dashboard rather than leaving it on `auto`.

`${interval:text}` does not avoid this. On scenes, `IntervalVariable` defines no `getValueText`, so the `text` formatter falls through to `getValue()` and forwards the resolved bucket anyway; and the target's `updateFromUrl` takes any non-sentinel string verbatim, so `var-interval=auto` would set a literal `auto` that is not one of its options. Forwarding the plain value at least gives the target something it can honour. The dashboard-level menu link still preserves `auto`, because `includeVars` emits the `$__auto` sentinel.

---
- #2083 — multi-value quoting

### DIFF
--- a/src/dashboards/__tests__/dashboards.test.ts
+++ b/src/dashboards/__tests__/dashboards.test.ts
@@ -147,4 +147,33 @@ describe('OTel dashboards', () => {
       expect(content).not.toMatch(/= '\$\{?\w+\}?'/);
     });
   });
+
+  describe('interval variable', () => {
+    type Panel = { gridPos?: unknown; interval?: string; targets?: Array<{ rawSql?: string }> };
+    type Dashboard = {
+      templating?: { list?: Array<{ name?: string; type?: string }> };
+      panels?: Array<Panel & { panels?: Panel[] }>;
+    };
+
+    const flatten = (d: Dashboard): Panel[] =>
+      (d.panels ?? []).flatMap((p) => [p, ...(p.panels ?? [])]);
+
+    it.each(otelDashboards)('%s exposes an "interval" interval variable', (filename) => {
+      const d = JSON.parse(fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8')) as Dashboard;
+      const variable = d.templating?.list?.find((v) => v.name === 'interval');
+      expect(variable).toBeDefined();
+      expect(variable?.type).toBe('interval');
+    });
+
+    it.each(otelDashboards)('%s sets min interval on every $__interval_s panel', (filename) => {
+      const d = JSON.parse(fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8')) as Dashboard;
+      const bucketed = flatten(d).filter((p) =>
+        (p.targets ?? []).some((t) => (t.rawSql ?? '').includes('$__interval_s'))
+      );
+      expect(bucketed.length).toBeGreaterThan(0);
+      for (const panel of bucketed) {
+        expect(panel.interval).toBe('${interval}');
+      }
+    });
+  });
 });
--- a/src/dashboards/__tests__/dashboards.test.ts
+++ b/src/dashboards/__tests__/dashboards.test.ts
@@ -175,5 +175,17 @@ describe('OTel dashboards', () => {
         expect(panel.interval).toBe('${interval}');
       }
     });
+
+    it.each(otelDashboards)('%s forwards the interval selection, not the resolved bucket', (filename) => {
+      // A data link interpolates ${interval} through to the concrete bucket size, so drilling
+      // through while "auto" is selected would pin that bucket on the target dashboard. The
+      // :text formatter forwards the selection as displayed ("auto", "30s"), which the target
+      // matches back to the same option — auto stays auto, an explicit choice stays explicit.
+      const content = fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8');
+      const forwarding = content.match(/var-interval=\$\{interval[^}]*\}/g) ?? [];
+      for (const param of forwarding) {
+        expect(param).toBe('var-interval=${interval:text}');
+      }
+    });
   });
 });
--- a/src/dashboards/otel-service-dashboard.json
+++ b/src/dashboards/otel-service-dashboard.json
@@ -85,6 +85,33 @@
         "type": "datasource"
       },
       {
+        "auto": true,
+        "auto_count": 30,
+        "auto_min": "10s",
+        "current": {
+          "selected": false,
+          "text": "auto",
+          "value": "$__auto_interval_interval"
+        },
+        "hide": 0,
+        "label": "Min interval",
+        "name": "interval",
+        "options": [
+          { "selected": true, "text": "auto", "value": "$__auto_interval_interval" },
+          { "selected": false, "text": "1m", "value": "1m" },
+          { "selected": false, "text": "5m", "value": "5m" },
+          { "selected": false, "text": "10m", "value": "10m" },
+          { "selected": false, "text": "30m", "value": "30m" },
+          { "selected": false, "text": "1h", "value": "1h" },
+          { "selected": false, "text": "6h", "value": "6h" },
+          { "selected": false, "text": "12h", "value": "12h" },
+          { "selected": false, "text": "1d", "value": "1d" }
+        ],
+        "query": "1m,5m,10m,30m,1h,6h,12h,1d",
+        "refresh": 2,
+        "type": "interval"
+      },
+      {
         "current": {},
         "datasource": {
           "type": "grafana-clickhouse-datasource",
@@ -195,6 +222,7 @@
         "y": 1
       },
       "id": 1,
+      "interval": "${interval}",
       "description": "Server spans per second for ${service}: incoming requests this service handles. The R in RED.",
       "title": "Request Rate (Server)",
       "type": "timeseries",
@@ -257,6 +285,7 @@
         "y": 1
       },
       "id": 2,
+      "interval": "${interval}",
       "description": "Fraction of Server spans for ${service} with StatusCode = 'Error'. The E in RED.",
       "title": "Error Rate (Server)",
       "type": "timeseries",
@@ -327,6 +356,7 @@
         "y": 1
       },
       "id": 3,
+      "interval": "${interval}",
       "description": "P50, P90 (solid) and P99 (dashed orange) Server span duration for ${service}. The D in RED.",
       "title": "Duration Percentiles (Server)",
       "type": "timeseries",
@@ -469,6 +499,7 @@
         "y": 16
       },
       "id": 12,
+      "interval": "${interval}",
       "description": "Client spans per second for ${service}: outgoing requests this service makes to its dependencies. The R in RED.",
       "title": "Request Rate (Client)",
       "type": "timeseries",
@@ -531,6 +562,7 @@
         "y": 16
       },
       "id": 13,
+      "interval": "${interval}",
       "description": "Fraction of Client spans for ${service} with StatusCode = 'Error'. The E in RED.",
       "title": "Error Rate (Client)",
       "type": "timeseries",
@@ -601,6 +633,7 @@
         "y": 16
       },
       "id": 14,
+      "interval": "${interval}",
       "description": "P50, P90 (solid) and P99 (dashed orange) Client span duration for ${service}. The D in RED.",
       "title": "Duration Percentiles (Client)",
       "type": "timeseries",
@@ -869,6 +902,7 @@
             "y": 42
           },
           "id": 16,
+          "interval": "${interval}",
           "description": "Internal spans per second for ${service}: in-process work (function-level instrumentation, no network).",
           "title": "Request Rate (Internal)",
           "type": "timeseries",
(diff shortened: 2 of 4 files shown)


src/dashboards/otel-traces-explorer.json
```diff
@@ -414,6 +442,7 @@
       "fieldConfig": { "defaults": { "color": { "mode": "fixed", "fixedColor": "text" } } },
       "gridPos": { "h": 2, "w": 8, "x": 16, "y": 25 },
       "id": 302,
+      "interval": "${interval}",
```
adamyeats, line 448: Now that bucket size is user-controlled, this stat's `lastNotNull` calc becomes visible behaviour: it reports the last bucket's count rather than a total, so the "Total span count" reading changes with the interval choice (33 at auto vs 249 at 10m when I tried it locally). It might be worth switching the calc to `sum`, as the Errors stat below already does.

src/dashboards/otel-service-dashboard.json
```diff
@@ -83,6 +83,33 @@
...
+        "auto_count": 30,
+        "auto_min": "10s",
+        "current": {
+          "selected": false,
+          "text": "auto",
+          "value": "$__auto_interval_interval"
+        },
+        "hide": 0,
+        "label": "Min interval",
+        "name": "interval",
```
adamyeats, line 98: The hand-built data links in these dashboards (e.g. the Slowest Server Operations drill-through to the traces explorer) forward service, operation and datasource but not this new variable, so the interval selection resets on drill-through, while the top-level menu links carry it via `includeVars`. Consider appending `&var-interval=${interval}` to those links, either here or as a follow-up.

src/dashboards/otel-traces-explorer.json
```diff
@@ -76,6 +76,33 @@
...
+          { "selected": false, "text": "1m", "value": "1m" },
+          { "selected": false, "text": "5m", "value": "5m" },
+          { "selected": false, "text": "10m", "value": "10m" },
+          { "selected": false, "text": "30m", "value": "30m" },
+          { "selected": false, "text": "1h", "value": "1h" },
+          { "selected": false, "text": "6h", "value": "6h" },
+          { "selected": false, "text": "12h", "value": "12h" },
+          { "selected": false, "text": "1d", "value": "1d" }
+        ],
+        "query": "1m,5m,10m,30m,1h,6h,12h,1d",
```
adamyeats: nit: with `auto_min` at 10s, auto can bucket finer than any explicit option, so picking anything from this list on a short time range coarsens to at least 1m. Consider adding 10s and 30s options.

AndreZiviani:
Thanks @adamyeats — addressed and rebased onto `main`:

- "Total span count" stat: calc `lastNotNull` → `sum`, matching the Errors stat. Left the P99/Mean duration stats on `lastNotNull` (summing quantiles/averages would be wrong).
- Appended `&var-interval=${interval}` to all 5 hand-built drill-through links (all in `otel-service-dashboard.json`).
- Added `10s` and `30s` options to the interval variable in all three dashboards.

`tsc --noEmit` clean; dashboards suite green (29/29).

AndreZiviani:
Thanks @adamyeats — addressed: the "Total span count" stat uses `sum`, the drill-through links carry `&var-interval=${interval}`, and I added `10s`/`30s` options. Rebased onto latest `main`. Re-requesting review.

src/dashboards/otel-service-dashboard.json
adamyeats: The five links now carry `&var-interval=${interval}`, which covers the explicit selections.

The default `auto` case is still open. The `interval` variable is `type: "interval"` with `auto: true`, and `auto` is selected by default. A data link resolves `${interval}` to the concrete bucket size, so a drill-through from the default state pins an interval on the traces explorer. The dashboard-level link forwards the `$__auto_interval_interval` sentinel through `includeVars`, so the menu route keeps `auto`.

Consider a check against a live dashboard first. If a pinned interval reads acceptably on the target, this is fine to leave.

adamyeats (changes requested):

adamyeats:

On ordering: this PR overlaps #2080 in `otel-logs-explorer.json` and `otel-service-dashboard.json`. It is worth landing this one last of the four, so it rebases onto the settled SQL rather than the other way round.

Both PRs append a different parameter to the same five data links in `otel-service-dashboard.json`. The correct resolution keeps both, as `&var-database=${database}&var-interval=${interval}`. No test covers link forwarding, so CI stays green if one is dropped.

AndreZiviani:
Rebased and force-pushed. Same `dashboards.test.ts` conflict, both blocks kept.

**On the `auto` case.** I fixed it rather than leaving it — I had no live dashboard to check the pinned interval against, and a fix that costs one formatter seemed better than a judgement call I could not actually make.

The five links now forward `${interval:text}` instead of `${interval}`. That is the selection as displayed rather than the resolved bucket, so `auto` stays `auto` and an explicit `30s` stays `30s` — which lines up with what the dashboard-level link already does via the `$__auto_interval_interval` sentinel. If the target ever fails to match the text it falls back to its own default, which is also `auto`, so the failure mode is safe either way.

Happy to drop this back to the pinned-interval behaviour if you check a live dashboard and it reads fine — it is a one-line revert.

**On the missing test.** Added a guard over the link parameters, since nothing covered forwarding. Verified it fails when reverted.

Note this changes the resolution you flagged for the #2080 collision — it is now

```
&var-database=${database}&var-interval=${interval:text}
```

Both PRs now carry a link guard, so dropping either parameter fails CI rather than staying green. Still happy for this to land last of the four.

src/dashboards/otel-service-dashboard.json
adamyeats: Checked `${interval:text}` against the scenes source, since scenes is the dashboard runtime from Grafana 11.3 and the only one in 12 and 13. The `text` formatter calls `variable.getValueText` only when the variable defines it. `IntervalVariable` does not, so the formatter falls through to `getValue()`, which is the resolved bucket. With `auto` selected the link still forwards `30s` or similar.

The target has the same problem. `IntervalVariable.updateFromUrl` maps `$__auto_interval_*` to auto and takes any other string verbatim, so `var-interval=auto` sets a literal `auto` that is not an option. What does preserve the selection is `getUrlState`, which emits `$__auto`, and that is only reachable through `${__all_variables}` or a dashboard link with `includeVars`.

Given that, your one-line revert to `${interval}` is the option I would take, with a sentence in the body that drill-throughs pin the current bucket. If you would rather try `${__all_variables}`, it needs a check of how a duplicate `var-operation` resolves, since the link also sets `var-operation=${__value.raw}`.

src/dashboards/__tests__/dashboards.test.ts
```diff
@@ -147,4 +147,45 @@ describe('OTel dashboards', () => {
...
+        (p.targets ?? []).some((t) => (t.rawSql ?? '').includes('$__interval_s'))
+      );
+      expect(bucketed.length).toBeGreaterThan(0);
+      for (const panel of bucketed) {
+        expect(panel.interval).toBe('${interval}');
+      }
+    });
+
+    it.each(otelDashboards)('%s forwards the interval selection, not the resolved bucket', (filename) => {
+      // A data link interpolates ${interval} through to the concrete bucket size, so drilling
```
adamyeats, line 180: The comment states that `:text` forwards `auto` and that the target matches it back. Neither holds on scenes, see the thread on the service dashboard. Worth updating alongside whichever fix you pick.

src/dashboards/__tests__/dashboards.test.ts
```diff
@@ -147,4 +147,45 @@ describe('OTel dashboards', () => {
...
+      // through while "auto" is selected would pin that bucket on the target dashboard. The
+      // :text formatter forwards the selection as displayed ("auto", "30s"), which the target
+      // matches back to the same option — auto stays auto, an explicit choice stays explicit.
+      const content = fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8');
+      const forwarding = content.match(/var-interval=\$\{interval[^}]*\}/g) ?? [];
```
adamyeats, line 185: The guard only iterates matches that exist, so a link that drops `var-interval` entirely still passes. Consider asserting that every `/d/otel-*` URL contains `var-interval=`, or that the match count equals the link count.

adamyeats (changes requested):
nit: the JSON-schema logs dashboard from #2084 has its own interval variable without the 10s and 30s options this PR adds to the other three. Fine as a follow-up, noting the four are now inconsistent.

AndreZiviani:
Thanks @adamyeats — that scenes dig was the answer, and I took the one-line revert.

- The five links forward plain `${interval}` again. Your reading matches the source: `IntervalVariable` defines no `getValueText`, so `:text` fell through to `getValue()` and forwarded the resolved bucket regardless, and `updateFromUrl` would have taken `var-interval=auto` verbatim as a non-option. I did not try `${__all_variables}` — the duplicate `var-operation` question you raised is real, and pinning the current bucket is at least a value the target can honour.
- The test comment claiming `:text` forwards `auto` is gone, replaced by what actually happens and why.
- The link guard no longer passes when a link drops the parameter: it checks each link it finds, plus a separate assertion that the service dashboard has links at all. Checked by deleting `&var-interval=${interval}` and watching it fail.
- Folded in the JSON-schema dashboard's missing `10s`/`30s` options, since its `auto_min` is `10s` too, so all four option lists now match.

The PR description records the drill-through behaviour. Still happy for this to land last of the four; the #2080 collision resolution is back to `&var-database=${database}&var-interval=${interval}`.

Dashboards suite green (54/54), rebased onto latest main.

adamyeats:
@AndreZiviani #2083 landed, so this branch conflicts in `src/dashboards/__tests__/dashboards.test.ts`. Could you rebase onto main? Your three fixes check out, so I will re-review as soon as CI is green. Landing this last of the four still makes sense.

AndreZiviani:
Rebased onto main — the `dashboards.test.ts` conflict was #2083's quoting guard landing next to this PR's interval block; both kept.

Nothing else changed: the five drill-through links still forward plain `${interval}`, the four option lists still match, and the link guard still fails when a link drops the parameter.

Dashboards suite green (61/61), `tsc` clean. Still happy for this to land last of the three remaining.

src/dashboards/otel-service-dashboard.json
adamyeats: Addressed in the 21 September push (cd2b3df). Resolving.

src/dashboards/__tests__/dashboards.test.ts
adamyeats: Addressed in the 21 September push (cd2b3df). Resolving.

src/dashboards/__tests__/dashboards.test.ts
adamyeats: Addressed in the 21 September push (cd2b3df). Resolving.

adamyeats (dismissed):
Checked the rebase against all three threads. The five links forward plain interval again, all four option lists match, the guard fails when a link drops the parameter, and the quoting from #2083 is intact. Approving. This will need one more rebase once #2080 lands, since both touch the service dashboard links.

adamyeats:
@AndreZiviani #2080 is on main now, so this branch conflicts again in `src/dashboards/otel-service-dashboard.json` and `src/dashboards/__tests__/dashboards.test.ts`. Could you rebase onto main once more? The approval stands on content, so once CI is green on the rebase I will re-approve and merge. The resolution should be `&var-database=${database}&var-interval=${interval}` on the five links, as you had before.

AndreZiviani:
Rebased onto main with #2080 landed. The five links are `&var-database=${database}&var-interval=${interval}`, as you called it.

Three conflicts, all expected:

- **`otel-logs-explorer.json`** — the `database` variable from #2080 and this PR's `interval` variable claim the same slot in `templating.list`. Both kept; the order is now datasource, database, interval, service, level_filter, search.
- **`otel-service-dashboard.json` links** — resolved to main's URL plus this PR's parameter, so neither is dropped.
- **`dashboards.test.ts`** — #2080's guards landed next to this PR's block; both kept.

The two intermediate commits that introduced and then reverted `${interval:text}` resolve to main's plain `${interval}` on those lines, since that is where the branch ends up anyway. The `:text` reasoning stays in the second commit's message for the record.

Checked after the rebase: dashboards suite green (80/80), and both link guards still fail under mutation — dropping `var-interval` fails this PR's guard, dropping `var-database` fails #2080's, so neither can regress silently. `tsc` clean, signatures verified.

adamyeats (approved):
Checked the rebase: the branch sits on current main, the five links carry both the database and the plain interval, all four option lists match, the #2083 quoting and #2080 qualification are intact, and all four guards are present. Approving.

adamyeats:
And thank you for your patience getting everything over the line here @AndreZiviani !

AndreZiviani:
thank you @adamyeats !

PR #2083, opened 2026-07-18
fix(dashboards): quote template variables in bundled dashboards

AndreZiviani:
## What
Interpolate template variables with the `:singlequote` format option instead of the bare form, across all bundled dashboards that filter on variables in raw SQL:

- **OpenTelemetry Logs / Traces / Service:** `service`, `level_filter`, `operation`, `status` in `IN (...)`, and `= '${service}'`.
- **Data Analysis:** `database`, `table` in `IN (...)`.
- **Query Analysis:** `query_kind`, `type`, `user` in `IN (...)`.

## Why
The datasource's default variable formatter (`format` in `src/data/CHDatasource.ts`) quotes a multi-value selection as `'a','b'`, but it does **not** escape a single quote inside a value, and it leaves a single-value selection **unquoted**. So:

- A value containing a single quote (e.g. a service named `o'brien-service`) interpolates to `IN ('o'brien-service')` — invalid SQL.
- A single-value variable used in `IN (...)` (e.g. `IN (${database})` in Data Analysis) interpolates to `IN (mydb)` — unquoted, also invalid.

`${var:singlequote}` uses Grafana's built-in formatter, which quotes **and** escapes embedded quotes, fixing both cases. Several panels already used `:singlequote`, so this also makes the quoting consistent.

A regression test in `dashboards.test.ts` asserts no bare `IN ($var)` / `IN (${var})` or `= '${var}'` remains in these dashboards (excluding `$__` macros).

---
_Part of a small series of independent improvements to the pre-built dashboards:_
- #2083 — template variable quoting (this PR)

All commits are signed.

### DIFF
--- a/src/dashboards/__tests__/dashboards.test.ts
+++ b/src/dashboards/__tests__/dashboards.test.ts
@@ -159,11 +159,13 @@ describe('OTel dashboards', () => {
     it.each(allDashboards)('%s quotes template variables in SQL', (filename) => {
       const content = fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8');
       // Template variables must be interpolated with :singlequote, not bare. The datasource's
-      // default formatter quotes multi-value selections ('a','b') but does not escape a single
-      // quote inside a value, and leaves single-value selections unquoted; :singlequote does
-      // both. ($__... macros are excluded.)
-      expect(content).not.toMatch(/IN \(\$\{?(?!__)\w+\}?\)/);
-      expect(content).not.toMatch(/= '\$\{?\w+\}?'/);
+      // default formatter (CHDatasource.format) already renders a multi-value selection as
+      // 'a','b', but it does not escape a single quote inside a value, so a value such as
+      // bob's-service produces invalid SQL. :singlequote escapes the quote; the plain formatter
+      // does not. Dashboards mix uppercase IN and lowercase in, so match case-insensitively.
+      // ($__... macros are excluded.)
+      expect(content).not.toMatch(/IN \(\$\{?(?!__)\w+\}?\)/i);
+      expect(content).not.toMatch(/= '\$\{?\w+\}?'/i);
     });
   });
 });
--- a/src/dashboards/__tests__/dashboards.test.ts
+++ b/src/dashboards/__tests__/dashboards.test.ts
@@ -147,4 +147,23 @@ describe('OTel dashboards', () => {
       expect(content).not.toMatch(/= '\$\{?\w+\}?'/);
     });
   });
+
+  describe('template variable quoting', () => {
+    // Guard every bundled dashboard, so a bare interpolation added to any of them is caught.
+    const allDashboards = (
+      JSON.parse(fs.readFileSync(PLUGIN_JSON, 'utf8')) as { includes: Array<{ type: string; path: string }> }
+    ).includes
+      .filter((i) => i.type === 'dashboard')
+      .map((i) => path.basename(i.path));
+
+    it.each(allDashboards)('%s quotes template variables in SQL', (filename) => {
+      const content = fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8');
+      // Template variables must be interpolated with :singlequote, not bare. The datasource's
+      // default formatter quotes multi-value selections ('a','b') but does not escape a single
+      // quote inside a value, and leaves single-value selections unquoted; :singlequote does
+      // both. ($__... macros are excluded.)
+      expect(content).not.toMatch(/IN \(\$\{?(?!__)\w+\}?\)/);
+      expect(content).not.toMatch(/= '\$\{?\w+\}?'/);
+    });
+  });
 });
--- a/src/dashboards/query-analysis.json
+++ b/src/dashboards/query-analysis.json
@@ -138,7 +138,7 @@
             }
           },
           "queryType": "sql",
-          "rawSql": "SELECT count() as \"Total queries\" FROM system.query_log WHERE type in ($type) AND initial_user IN (${user:singlequote}) AND query_kind IN (${query_kind:singlequote}) AND $__timeFilter(event_time) ",
+          "rawSql": "SELECT count() as \"Total queries\" FROM system.query_log WHERE type in (${type:singlequote}) AND initial_user IN (${user:singlequote}) AND query_kind IN (${query_kind:singlequote}) AND $__timeFilter(event_time) ",
           "refId": "A"
         }
       ],
@@ -220,7 +220,7 @@
             }
           },
           "queryType": "sql",
-          "rawSql": "SELECT avg(memory_usage) as \"Avg query memory\", $__timeInterval(query_start_time) as time FROM system.query_log WHERE type in ($type) AND initial_user IN (${user:singlequote}) AND query_kind IN (${query_kind:singlequote}) AND $__timeFilter(event_time) GROUP BY time ORDER BY time",
+          "rawSql": "SELECT avg(memory_usage) as \"Avg query memory\", $__timeInterval(query_start_time) as time FROM system.query_log WHERE type in (${type:singlequote}) AND initial_user IN (${user:singlequote}) AND query_kind IN (${query_kind:singlequote}) AND $__timeFilter(event_time) GROUP BY time ORDER BY time",
           "refId": "A"
         }
       ],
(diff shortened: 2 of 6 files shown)


src/dashboards/__tests__/dashboards.test.ts
```diff
@@ -48,4 +48,14 @@ describe('OTel dashboards', () => {
       expect(fs.existsSync(filepath)).toBe(true);
     });
   });
+
+  describe('multi-value variable quoting', () => {
+    it.each(otelDashboards)('%s quotes multi-value variables in IN clauses', (filename) => {
+      const content = fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8');
+      // Multi-value variables must be interpolated with :singlequote, never bare,
+      // otherwise a multi-selection expands to unquoted / comma-joined SQL.
```
adamyeats: This overstates the failure mode: the datasource's custom formatter (`format` in `src/data/CHDatasource.ts`) already renders multi-value selections as `'a','b'`, so the bare form only breaks on values containing a single quote, which `:singlequote` escapes and the formatter does not. Consider rewording the comment (and the PR description) so the guard documents the real reason.

src/dashboards/__tests__/dashboards.test.ts
```diff
@@ -48,4 +48,14 @@ describe('OTel dashboards', () => {
...
+      expect(content).not.toMatch(/IN \(\$(?:service|level_filter|operation|status)\)/);
```
adamyeats: Consider generalising this to something like `/IN \(\$\{?(?!__)\w+\}?\)/` so it also catches the braced form `IN (${service})` and variables added later, for example the `database` variable from #2080.

src/dashboards/__tests__/dashboards.test.ts
```diff
@@ -48,4 +48,14 @@ describe('OTel dashboards', () => {
...
+      expect(content).not.toMatch(/= '\$\{service\}'/);
```
adamyeats: nit: this only guards the braced `service` form. It might be worth widening to `/= '\$\{?\w+\}?'/` to cover other variables and the unbraced spelling.

src/dashboards/__tests__/dashboards.test.ts
```diff
@@ -48,4 +48,14 @@ describe('OTel dashboards', () => {
       expect(fs.existsSync(filepath)).toBe(true);
     });
   });
+
+  describe('multi-value variable quoting', () => {
+    it.each(otelDashboards)('%s quotes multi-value variables in IN clauses', (filename) => {
```
adamyeats: nit: the name mentions only IN clauses but the second assertion guards the equality form too. Consider something like "quotes template variables in SQL".

adamyeats (commented):
Verified locally with seeded OTel data: multi-select interpolates to correctly quoted `IN ('a','b')` everywhere, and the new test fails on main as intended. One correction: the plugin's custom formatter already quotes multi-value selections, so what this actually fixes is values containing a single quote (`bob's-service` errors on main and works with this change). Worth updating the description and test comment to match, suggestions inline. LGTM.

src/dashboards/__tests__/dashboards.test.ts
```diff
@@ -48,4 +48,14 @@ describe('OTel dashboards', () => {
       expect(fs.existsSync(filepath)).toBe(true);
     });
   });
+
+  describe('multi-value variable quoting', () => {
```
adamyeats: It might be worth a follow-up applying the same treatment to `data-analysis.json` and `query-analysis.json`, which still use bare `IN (${database})` and `IN ($type)` and carry the same quote-escaping gap.

AndreZiviani:
thanks @adamyeats , I've addressed your reviews. I've implemented the fix on `data-analysis.json` and `query-analysis.json` on this same PR, let me know if you want me to move it to separate one

src/dashboards/query-analysis.json
```diff
@@ -138,7 +138,7 @@
             }
           },
           "queryType": "sql",
-          "rawSql": "SELECT count() as \"Total queries\" FROM system.query_log WHERE type in ($type) AND initial_user IN ($user) AND query_kind IN ($query_kind) AND $__timeFilter(event_time) ",
+          "rawSql": "SELECT count() as \"Total queries\" FROM system.query_log WHERE type in ($type) AND initial_user IN (${user:singlequote}) AND query_kind IN (${query_kind:singlequote}) AND $__timeFilter(event_time) ",
```
adamyeats: It looks like the lowercase `in ($type)` kept the bare form here and in seven other queries (lines 223, 311, 383, 466, 565, 673 and 1385), while the uppercase one on line 777 was converted, so `$type` is now quoted in one panel and bare in the rest. Consider converting these too.

src/dashboards/__tests__/dashboards.test.ts
```diff
@@ -48,4 +48,23 @@ describe('OTel dashboards', () => {
...
+      .filter((i) => i.type === 'dashboard')
+      .map((i) => path.basename(i.path));
+
+    it.each(allDashboards)('%s quotes template variables in SQL', (filename) => {
+      const content = fs.readFileSync(path.join(DASHBOARDS_DIR, filename), 'utf8');
+      // Template variables must be interpolated with :singlequote, not bare. The datasource's
+      // default formatter quotes multi-value selections ('a','b') but does not escape a single
+      // quote inside a value, and leaves single-value selections unquoted; :singlequote does
+      // both. ($__... macros are excluded.)
+      expect(content).not.toMatch(/IN \(\$\{?(?!__)\w+\}?\)/);
```
adamyeats: Consider adding the `i` flag to both patterns: ClickHouse SQL in these dashboards mixes `IN` and `in`, and the lowercase `in ($type)` occurrences in query-analysis.json currently slip past this guard.

AndreZiviani:
Thanks @adamyeats — addressed and rebased over #2092 (`positionCaseInsensitive`) + #2077:

- Reworded the test comment to document the real reason: the formatter (`CHDatasource.format`) already renders a multi-value selection as `'a','b'`; the bare form breaks on a value containing a single quote (e.g. `bob's-service`), which `:singlequote` escapes and the plain formatter does not.
- Generalized the guards to `/IN \(\$\{?(?!__)\w+\}?\)/i` and `/= '\$\{?\w+\}?'/i` — braced form + case-insensitive.
- Converted the 9 remaining lowercase `type in ($type)` occurrences in `query-analysis.json` to `in (${type:singlequote})` (the uppercase one on line 777 was already done). `data-analysis.json` had no bare forms left.
- The new #2077 `customSql` annotation variables are quoted too.
- Renamed the test to "quotes template variables in SQL".

`tsc --noEmit` clean; dashboards suite green (30/30).

SpencerTorres (approved):
Thank you for updating this 👌

adamyeats:

On ordering: this PR overlaps #2080 heavily. A three-way merge of the two heads conflicts in all three OTel dashboard JSON files, because both rewrite the same SQL strings. It is worth landing this one before #2080, so the quoting test here is in place when #2080 adds `${database}` to those queries.

AndreZiviani:
Rebased onto latest main and force-pushed. Same `dashboards.test.ts` conflict, both blocks kept.

No content change. The generalized quoting guard now also covers `otel-logs-explorer-json.json` from #2084, which passes as-is.

Agreed on landing this before #2080.

src/dashboards/__tests__/dashboards.test.ts
```diff
@@ -147,4 +147,25 @@ describe('OTel dashboards', () => {
       expect(content).not.toMatch(/= '\$\{?\w+\}?'/);
     });
   });
+
+  describe('template variable quoting', () => {
+    // Guard every bundled dashboard, so a bare interpolation added to any of them is caught.
+    const allDashboards = (
```
adamyeats, line 153: nit: the `quotes template variables in SQL` case in the JSON-schema block above (lines 142 to 148) is now fully covered by this generalised, case-insensitive guard. It can go.

adamyeats, line 153: nit: this inner `allDashboards` shadows the module-level one at line 14. The two lists are built differently, plugin.json includes against `readdirSync`, so a dashboard present in only one is checked by only one guard. Consider reusing the module-level list, or renaming this one.

adamyeats (dismissed):

AndreZiviani:
Thanks @adamyeats — both addressed, and rebased onto latest main (the same `dashboards.test.ts` conflict with #2084, both blocks kept).

- Dropped the JSON-schema block's own `quotes template variables in SQL` case; the generalised case-insensitive guard covers that file.
- The inner `allDashboards` is gone — the guard now uses the module-level list. That one reads the directory rather than `plugin.json`, so a dashboard is guarded even before it is registered.

Dashboards suite green (48/48), and the guard still runs over all eight bundled dashboards.

adamyeats (approved):
Checked the 21 September push against both threads. All addressed. Approving.

src/dashboards/__tests__/dashboards.test.ts
adamyeats: Addressed in the 21 September push (619a741). Resolving.

adamyeats: Addressed in the 21 September push (619a741). Resolving.

PR #2084, opened 2026-07-18
feat(dashboards): add JSON-schema OTel logs dashboard

AndreZiviani:
## What
Adds `otel-logs-explorer-json.json`, a variant of the OpenTelemetry Logs Explorer for ClickHouse tables that store OTel attributes with the native **JSON** type (the ClickStack schema) instead of `Map(String, String)`. Registered in `plugin.json` alongside the Map-schema dashboard.

## Why a separate dashboard
The two schemas need different attribute-access syntax and **cannot be served by one static query**: on a JSON column `ResourceAttributes.service.namespace` reads the nested path, whereas the Map form `ResourceAttributes['service.namespace']` errors on a JSON column (`ILLEGAL_TYPE_OF_ARGUMENT`). ClickHouse can't branch on column type inside a static query, so a duplicate is the clean option.

- Only the **Log Samples** panel reads `otel_logs` attributes; it uses dot access with a `::String` cast (validated against a live ClickHouse 26.2 JSON table).
- The deployment annotation reads `otel_traces`, which remains `Map`, so its access is left unchanged.
- The dashboard mirrors the improved Map edition (database selector, importable inputs, interval variable, consistent quoting).

---
- #2083 — multi-value quoting

### DIFF
--- a/docs/sources/dashboards.md
+++ b/docs/sources/dashboards.md
@@ -21,9 +21,9 @@ review_date: 2026-08-12
 
 # ClickHouse OpenTelemetry dashboards
 
-The plugin ships three pre-built dashboards for OpenTelemetry data stored in ClickHouse. Together they cover top-down log exploration, service topology and trace search, and a single-service deep dive that ties RED metrics, errors, logs, and trace detail into one view.
+The plugin ships four pre-built dashboards for OpenTelemetry data stored in ClickHouse. Three cover top-down log exploration, service topology and trace search, and a single-service deep dive that ties RED metrics, errors, logs, and trace detail into one view; a fourth is a variant of the logs explorer for tables that store OTel attributes as the native JSON type.
 
-The three dashboards link to each other via dashboard data links: clicking a service or operation in one dashboard preserves the time range and datasource and lands you on the matching view in another.
+The core three dashboards link to each other via dashboard data links: clicking a service or operation in one dashboard preserves the time range and datasource and lands you on the matching view in another.
 
 ## Required schema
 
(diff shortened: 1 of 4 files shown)


akkikumar72:
Hi @AndreZiviani  thank you for this contribution. we really appreciate the work you’ve put into it.

Apologies for the delay in starting the review. Before I continue, could you please review and address the Copilot comments above, and update the branch with the latest changes from main?

AndreZiviani:
Thanks for the review — addressed all four findings:

1. Replaced `hasToken` with `positionCaseInsensitive(Body, '$search') > 0` in all six panel queries (matching the other OTel dashboards), so free-text searches containing spaces/separators work.
2. Disabled the deployment annotation by default and renamed it "Deployments (service.version, requires traces)", so a logs-only database doesn't fail on load.
3. Added the dashboard to the shared `otelDashboards` test suite, so it now runs the common checks (including the `hasToken`/substring guard).
4. Updated `docs/sources/dashboards.md` to distinguish the `${database}`-selector JSON dashboard from the three Map dashboards.

akkikumar72 (commented):
Thanks @AndreZiviani, the Copilot follow-up looks good. 

Two things before I approve:
1. Please rebase onto current `main`. This branch is still behind.

2. The Database picker currently lists every database that has an `otel_logs` table, including Map-schema ones:

`SELECT DISTINCT database FROM system.tables WHERE name = 'otel_logs'`

If both schemas exist, Grafana can auto-select a Map database. Log Samples then fails with `ILLEGAL_TYPE_OF_ARGUMENT` because it queries `ResourceAttributes.service.namespace::String`.

Please limit the picker to databases where `ResourceAttributes` is a JSON column, for example with `system.columns` and `startsWith(type, 'JSON')`.

AndreZiviani:
Thanks @akkikumar72 — both addressed:

1. **Rebased** onto current `main`.
2. **Database picker** now lists only databases where `otel_logs.ResourceAttributes` is a JSON column, so a Map-schema database can't be auto-selected (which was causing the `ILLEGAL_TYPE_OF_ARGUMENT` on Log Samples):
   ```sql
   SELECT DISTINCT database FROM system.columns
   WHERE table = 'otel_logs' AND name = 'ResourceAttributes' AND startsWith(type, 'JSON')
   ORDER BY database
   ```
   (`definition` mirrors the query, matching the other OTel dashboards.) Verified against a live instance — it correctly excludes Map-schema databases.

akkikumar72 (approved):
Thanks @AndreZiviani for your contribution 🎖️

PR #2086, opened 2026-07-21
Surface log table columns as browsable, filterable fields

alex-fedotyev:
## Type of Change

- [x] 🚀 Feature

---

### What is this feature?

Two new options in the logs **Default columns** section of the data source config, both off by default:

- **Include all columns**: a toggle that shows every column in the table as a log field.
- **Additional columns**: an explicit list of column names to show, used when the toggle is off.

The selected columns show up in the Fields list, in the log row details, and in the field filters, and can be added as displayed columns. Under the hood the top-level scalar columns are merged into the log labels under their plain names, so they render in the log details under a **Fields** group next to the grouped **Resource attributes** / **Log attributes**, and filter as real columns. Whether those attribute groups appear as separate named sections or flat under a single **Fields** list depends on the Grafana version's log-details rendering; filtering and the folded columns behave the same either way. The OTel attribute maps keep their existing grouped rendering, unchanged.

Today the logs view only surfaces the role columns (time, level, message) and the OTel attribute maps as browsable, filterable fields. Any other column, for example `ServiceName` on an OTel table, or `status` and `method` on a flat access-log table, is only visible if you hand-edit the query to select it, and even then it does not appear in the log details with filter buttons. People who explore logs by pointing and clicking rather than writing SQL cannot see or filter by those columns.

This lets an operator expose the columns they care about once, in the data source config, and then browse and filter by them across every logs query, the same way attributes already work.

### Who is this feature for?

Anyone using the ClickHouse data source for logs, especially point-and-click users in Explore who filter by clicking values rather than editing SQL, and operators who want a consistent set of fields available across dashboards without editing each panel.

### How to test this feature

1. Open a ClickHouse data source set up for logs (`Connections` -> `Data sources`), open the **Logs** section, and under **Default columns** turn on **Include all columns** (or add a few column names under **Additional columns**). Save and test.
2. In Explore, run a logs query against that data source.
3. Confirm the extra columns appear in the Fields list on the left and can be checked to show as columns.
4. Open a log row's details and confirm the extra columns appear under a **Fields** group, next to **Log attributes** / **Resource attributes**, each with filter-for and filter-out buttons.
5. Click **Filter for value** on one of them and confirm the query gains a real column filter (for example `ServiceName = '...'`) and the logs volume histogram still loads.

---

## Screenshots / Videos

Verified locally with `test-pr.sh` against the official plugin build, using the public OTel demo dataset (`otel_v2.otel_logs`). "Before" is the feature off; "After" is **Include all columns** on. Same query and time range in both.

**Log row details.** The top-level columns render under a new **Fields** group, next to **Log attributes** and **Resource attributes**, each with filter-for and filter-out buttons.

| Before (feature off) | After (this PR) |
| --- | --- |
| Log details, feature off | Log details with Fields group |

**Fields list.** The same columns become browsable, filterable entries in the Fields list. Both shots are scrolled to the same alphabetical region, below the attribute-map keys, so the added columns (`ServiceName`, `SeverityNumber`, `SpanId`, `TraceFlags`, `ScopeName`, `ScopeVersion`, `ResourceSchemaUrl`) stand out from the dimmer `ResourceAttributes.*` keys.

| Before (feature off) | After (this PR) |
| --- | --- |
| Fields list, feature off | Fields list with columns surfaced |

---

## Please check that:

- The columns are merged into the frame's `labels` field on the frontend (`foldDiscoveredLogFieldsIntoLabels` in `src/data/utils.ts`, called from `CHDatasource.query()`), reusing the same `labels` field the backend already builds for the OTel attribute maps. `getLabelDisplayTypeFromFrame` returns `Fields` for the unprefixed keys so they render under their own group. No backend change.
- Filtering uses the plain column name (no alias) on purpose, so a "Filter for value" click resolves against the real column in both the main query and the separate logs-volume query. An alias would only exist in the main query's projection and would break the volume query.
- The options render in both config editors: the classic one and the new config page design behind the `newClickhouseConfigPageDesign` feature toggle.
- One known rough edge: with **Include all columns** on, a cold Explore load may need one **Run** before the columns appear, because the column list comes from the table schema, which is fetched asynchronously. Dashboards and warm loads are unaffected, and **Additional columns** does not have this since the names come from config. Happy to change how this is gated if you would prefer.


SpencerTorres (approved):
Looks good, only thing standing out from a code perspective is the comments (I worry comments of that length may become outdated as other changes accumulate)

alex-fedotyev:
Good call on the comments. I tightened the two longest doc comments (`appendAdditionalLogColumns` and the field-fold helper) down to the essential rationale so they are less likely to drift as the surrounding code changes.

I also merged main to resolve the conflict with #2073, which refactored the same compact-editor init. I unified the two column-threading paths onto the richer `TableColumn` type, kept the OTel latest-schema detection and the message/level auto-detection from #2073, and updated the compact-defaults tests to match. Full suite passes locally.

SpencerTorres (approved):

adamyeats (commented):
@alex-fedotyev Thanks! Added some comments below.

In answer to your questions on Slack: the need is real and the approach fits it. I verified the loop locally and filtering on plain column names rather than aliases seems like the right call to me.

The structural concern (detail in the inline comments): for a non-OTel table on an all-databases datasource, **Include all columns** is applied by neither hook, so it silently does nothing in the classic builder. The customers you describe are served today only by **Additional columns**, which may be the better answer anyway. It works in every mode, needs no schema fetch, and does not widen every query on large tables. Consider either wiring the toggle through the non-OTel builder path or scoping its tooltip to the modes where it works.

src/components/queryBuilder/views/logsQueryBuilderHooks.ts
```diff
@@ -47,6 +48,11 @@ export const useLogDefaultsOnMount = (
       }
     }
 
+    // Extra fields configured in Default columns. No schema here (new-query defaults),
+    // so only the explicit list applies; the "all" mode is handled by useOtelColumns
+    // and the compact editor, which have the column list.
+    appendAdditionalLogColumns(datasource, [], nextColumns, includedColumns);
```
adamyeats, line 54: With OTel disabled, `useOtelColumns` returns early, so for a non-OTel table in the classic builder the include-all mode is applied by neither hook and the toggle silently does nothing (only the explicit list works there). It might be worth wiring the fetched schema through this path too, or noting the limitation in the tooltip.

src/components/queryBuilder/QueryBuilder.tsx
```diff
@@ -209,21 +209,35 @@ const CompactQueryEditor = (props: CompactQueryEditorProps) => {
       return;
     }
 
+    // With "Include all columns" on, wait for the schema so the first logs query already
+    // carries the extra columns (a single query instead of a base query then a re-run).
+    if (signalType === 'logs' && datasource.shouldIncludeAllLogColumns() && allColumns.length === 0) {
```
adamyeats: If the schema fetch fails, `useColumns` swallows the error and keeps returning an empty array, so this gate never lifts and the compact editor never initialises a query at all (previously it fell back to the base defaults). Consider only deferring while the fetch is in flight, or falling back to the base defaults on error.

src/data/utils.ts
```diff
@@ -677,3 +678,99 @@ export const dataFrameHasLogLabelWithName = (frame: DataFrame | undefined, name:
...
+    const builderOptions = (target as CHBuilderQuery).builderOptions;
+    if (!builderOptions || builderOptions.queryType !== QueryType.Logs) {
+      continue;
+    }
+
+    // The columns the field options selected are the hint-less, non-collection ones. Role columns
+    // (time / body / level / trace id) carry a hint; the attribute maps are collection-typed.
+    const discovered = (builderOptions.columns || [])
+      .filter((c) => c.hint === undefined && !isCollectionColumnType(c.type))
+      .map((c) => c.alias || c.name);
```
adamyeats: An aliased hint-less column gets folded under its alias, so a filter-for click builds a filter keyed on the alias, which the logs-volume query cannot resolve (the exact failure the PR description says plain names were chosen to avoid). Consider skipping columns that carry an alias so they stay as ordinary frame fields.

src/components/queryBuilder/views/columnNameHeuristics.ts
```diff
@@ -65,3 +65,16 @@ export const isNumericColumn = (column: TableColumn): boolean => {
...
+/** Type predicate: ClickHouse collection / compound column types (Map, Array, Tuple, Nested, Object, JSON). */
+export const isCollectionColumnType = (type?: string): boolean => {
+  const s = (type || '').trim();
+  return (
+    s.startsWith('Map(') ||
+    s.startsWith('Array(') ||
+    s.startsWith('Tuple(') ||
+    s.startsWith('Nested') ||
+    s.startsWith('Object(') ||
+    s === 'JSON'
```
adamyeats: Parameterised JSON types such as `JSON(max_dynamic_paths=100)` won't match this equality check, so include-all would select and fold such a column as if it were a scalar. `s === 'JSON' || s.startsWith('JSON(')` would cover both forms.

src/data/CHDatasource.ts
```diff
@@ -2125,7 +2145,7 @@ export class Datasource
     if (labelKey.startsWith('LogAttributes.')) {
       return 'Log attributes';
     }
-    return null;
+    return 'Fields';
```
adamyeats: This changes the grouping for every unprefixed label key, including datasources with both new options off, rather than only for folded fields. It looks harmless since the default section renders much the same, but it might be worth confirming that is intended rather than gating it on the feature.

src/data/utils.ts
```diff
@@ -677,3 +678,99 @@ export const dataFrameHasLogLabelWithName = (frame: DataFrame | undefined, name:
...
+            obj = {};
+          }
+        } else {
+          obj = (raw as Record<string, unknown>) || {};
+        }
+        if (!obj || typeof obj !== 'object') {
+          obj = {};
+        }
+        obj[name] = typeof v === 'string' ? v : String(v);
+        labelsField.values[i] = wasString ? JSON.stringify(obj) : obj;
```
adamyeats, line 770: For the backend-serialised shape this parses and re-serialises the whole labels object once per folded column per row, so include-all on a wide table does that round trip many times per row. Parsing each row once, folding every column, then serialising once would keep it linear.

src/components/queryBuilder/compactQueryDefaults.ts
```diff
@@ -9,6 +16,44 @@ import {
...
 } from './defaultQueryOptions';
+import { isCollectionColumnType, isDateTimeColumn } from './views/columnNameHeuristics';
+
+/**
+ * Appends the datasource's configured extra log columns so they surface as first-class fields
+ * (both modes default off): `includeAllColumns` adds every detected top-level scalar column and
+ * needs the table schema; `additionalColumns` adds an explicit list and needs none. Mutates
+ * `columns` / `includedColumns` in place.
+ */
+export const appendAdditionalLogColumns = (
```
adamyeats, line 27: The skip rules here (collection types, the `__` prefix, DateTime columns, and the `split('[')` dedupe) have no direct tests, and the compactQueryDefaults suite only adapts existing cases to the new signature. Since the function is exported, a small table of cases would pin the behaviour down.

src/components/configEditor/LogsConfig.tsx
```diff
@@ -24,6 +24,8 @@ interface LogsConfigProps {
   onSelectContextColumnsChange: (v: boolean) => void;
   onContextColumnsChange: (v: string[]) => void;
   onShowLogLinksChange: (v: boolean) => void;
+  onIncludeAllColumnsChange?: (v: boolean) => void;
```
adamyeats: nit: every sibling handler prop is required and all three call sites pass these two, so consider making them required and dropping the `?.` calls.

src/components/queryBuilder/ColumnsEditor.tsx
```diff
@@ -1,11 +1,26 @@
...
 import { selectors } from 'selectors';
 import { styles } from 'styles';
 
+// Let the selected-column chips wrap onto multiple lines within a bounded width, instead of forming
+// one long row that fills the builder edge to edge (and clips when many columns are selected, for
+// example logs with "Include all columns" on). The dropdown menu still lists every column. The
+// builder's column width is driven by its widest row, so a bounded value list also keeps the whole
+// builder from stretching to fit a long column list.
+const wrappedColumnValues = css`
+  & [class*='grafana-select-value-container'] {
```
adamyeats, line 16: nit: this selector reaches into react-select internals that a Grafana UI upgrade could rename, and the hard-coded 760px wrap applies to every query type's columns editor, not just logs. It might be worth confirming the visual change is wanted builder-wide.

src/components/queryBuilder/views/logsQueryBuilderHooks.ts
alex-fedotyev: Good catch. Wired the schema through the non-OTel path in 356df0d. Added `useIncludeAllLogColumns`, the non-OTel counterpart to `useOtelColumns`, which applies the detected columns once the schema loads, so include-all now works on a non-OTel table in the classic builder. It dispatches a new `mergeColumns` reducer action that appends columns by name against live state (mirroring `setColumnByHint`), so it does not clobber the role columns the sibling hooks set on the same render.

src/components/queryBuilder/QueryBuilder.tsx
alex-fedotyev: Good catch. Dropped the wait-guard in 7e9b1e4. It keyed on `allColumns` becoming non-empty, but `useColumns` swallows fetch errors and keeps returning `[]`, so a failed fetch or an empty table blocked initialization entirely. It now initializes with base defaults and re-runs via the existing key (which flips once column names arrive), the same base-then-rerun the OTel detection already relies on.

src/data/utils.ts
alex-fedotyev: Good catch. Fixed in f92af6f. The fold now skips columns that carry an alias and folds under the real column name, so a filter-for click keys on the real column and resolves in the logs-volume query. Aliased columns stay as ordinary frame fields.

src/components/queryBuilder/views/columnNameHeuristics.ts
alex-fedotyev: Good call. Fixed in f92af6f: `isCollectionColumnType` now matches `JSON(...)` too (`s === 'JSON' || s.startsWith('JSON(')`), so a parameterized JSON column is not treated as a scalar.

src/data/CHDatasource.ts
alex-fedotyev: Good call, gated it in f92af6f. `getLabelDisplayTypeFromFrame` returns `'Fields'` for unprefixed keys only when `includeAllColumns` or `additionalColumns` is set, and `null` (the prior behavior) otherwise. Nothing is folded when both are off, so there are no unprefixed keys to group in that case.

src/data/utils.ts
alex-fedotyev: Good call. Reworked in f92af6f: the fold parses and serializes each row's `labels` once and folds every column in that single pass, instead of a round trip per column per row.

src/components/queryBuilder/compactQueryDefaults.ts
alex-fedotyev: Added in f92af6f: a case table for `appendAdditionalLogColumns` covering the collection/JSON/`__`-prefix/DateTime skips, the already-included skip, the explicit-list ordering, and the `split('[')` map-key dedupe.

src/components/configEditor/LogsConfig.tsx
alex-fedotyev: Done in f92af6f. Made both handlers required and dropped the optional calls; all three call sites already pass them.

src/components/queryBuilder/ColumnsEditor.tsx
alex-fedotyev: The builder-wide wrap is intentional, it is a general no-overflow improvement for every query type's column editor, so I have left it applied across builders. You are right that the `[class*='grafana-select-value-container']` selector leans on a Grafana internal; happy to scope it to the logs builder, or swap the fixed 760px for a token-derived width, if you would prefer that instead.

alex-fedotyev:
Thanks for the review. I pushed fixes for all nine threads in f92af6f (rebased on the latest main). The two structural ones: include-all now applies on a non-OTel table in the classic builder via a dedicated schema-arrival hook, and the compact editor no longer blocks initialization when the schema fetch fails or the table is empty. The rest are the alias-fold, the JSON type predicate, feature-gated grouping, a linear fold, the config handler props, and added tests. Replied inline on each thread. Ready for another look.

alex-fedotyev:
One more self-review pass before your re-review turned up two things worth folding in (aa04fdb, 4cac7fe, bd53d1b):

- `setColumnByHint` filtered only by hint, so a plain column added by "Include all columns" that then matched a Message/Level name heuristic could be projected twice (for example `SeverityText as "level", SeverityText`). It now also drops a same-named hint-less column when promoting one to a role.
- The single-table (compact) editor never applied include-all for a non-OTel table, and because it assigned no Time column the logs query had no time field and did not render. It now runs the same schema-arrival hooks the classic builder uses (a detected Time column plus the include-all columns), both no-ops when OTel is on.

Added tests for the OTel and compact include-all wiring, the duplicate-projection fix, and the compact non-OTel path. Verified locally against a non-OTel host-logs table in both the classic and single-table builders and against the OTel demo. Ready for another look.

alex-fedotyev:
Thanks for the review. I've pushed fixes for all nine comments, with replies on each thread. The notable changes since your pass:

- Include all columns now works in the classic non-OTel builder (it was a silent no-op before); the compact editor also assigns a time column and the detected columns for non-OTel tables.
- Fixed a double projection where a column promoted to a role (for example SeverityText to level) was selected twice.
- The logs builder no longer emits an invalid empty query during the brief window before a non-OTel table's schema loads; filterQuery now skips a query with no SQL.
- The collection-type guard is deduped into a shared helper, and the label fold skips aliased columns.

Tests added or updated: reducer dedup, the include-all wiring for OTel and compact, the empty-query and filterQuery guards, the cold-load compact case, and by-name message/level detection. Ready for another look when you have time.

alex-fedotyev:
Closing in favor of #2108, which replaces this with a cleaner history and the design we settled on.

The main change from this PR: I dropped the **Include all columns** toggle. Per the discussion here, an explicit column list is the better primitive (it works in every mode, needs no query-time schema, and does not widen every query on a large table), so the config now has a single **Columns** field, and **Add all columns** covers the "I want them all" case as a one-click action. In single source mode the Columns field and the time, level, and message roles are pickers of the table's real columns; other modes take a typed list.

All the review comments here are carried forward in #2108. One note on the label-grouping comment: I kept `getLabelDisplayTypeFromFrame` returning `Fields` for unprefixed keys rather than gating it, because the fold can now come from an ad-hoc column selection in the query editor, not only the config. The backend only writes prefixed attribute keys, so an unprefixed key is a folded column (or a table's own `labels` map, which is harmless to group under the same section). I corrected the code comment to say exactly that. New tests cover the Add-all skip rules and the schema-backed config field.

Thanks for the review here, it shaped the final approach.

PR #2092, opened 2026-07-22
fix(logs): apply log message search to logs volume and sample queries

alex-fedotyev:
## Summary

The log message search box in the Explore logs view is applied to the main logs query but not to the supplementary **logs volume** histogram or **logs sample** queries. `getSupplementaryLogsVolumeQuery` and `getSupplementaryLogsSampleQuery` both rebuild `QueryBuilderOptions` from `filters`/`columns`/`aggregates` without carrying `meta`, and `meta.logMessageLike` is where the message search lives. The result is a histogram that shows bars while the filtered sample list is empty.

This applies the same message search to both supplementary queries as an explicit `LIKE` filter on the LogMessage column.

## Changes

- Add a private `getLogMessageSearchFilter` helper that builds a `FilterOperator.Like` filter from `meta.logMessageLike` on the LogMessage column, or returns nothing when the search is empty or no LogMessage column is configured.
- Append that filter in `getSupplementaryLogsVolumeQuery` and `getSupplementaryLogsSampleQuery`.
- Unit tests for both builders (message applied, message unset, no LogMessage column).

## Why

The filter targets the LogMessage column's real name (for example `Body`), not the `body` alias the main query uses in its SELECT. The volume query does not project that alias, so an alias-based predicate would not resolve there. The real column name resolves in both the main and the supplementary queries.

## Local verification

Reproduced against the public ClickHouse demo (`sql-clickhouse.clickhouse.com`, `otel_v2.otel_logs`) in Explore: query type Logs, filter `SeverityText LIKE error`, message search `found`. In this dataset "found" only appears in info-level bodies, so the error-level sample is empty. Confirmed the generated SQL before and after against the demo (last 1 hour):

| Query | Predicate | Rows |
| ----- | --------- | ---- |
| Logs sample (what the user sees) | `error AND Body LIKE '%found%'` | 0 |
| Logs volume, before | `error` only | 562 |
| Logs volume, after (this PR) | `error AND Body LIKE '%found%'` | 0 |

After the change the histogram matches the sample list. With a term that does appear in error bodies (`failed`) the sample and the volume query return the same non-zero count (207), so the fix narrows the histogram rather than blanking it.

## Test plan

- [x] npx jest (full suite, 998 passed)
- [x] npx tsc --noEmit
- [x] npx eslint
- [x] npx prettier --check
- [x] npm run spellcheck

### DIFF
--- a/src/data/CHDatasource.test.ts
+++ b/src/data/CHDatasource.test.ts
@@ -1477,22 +1477,6 @@ describe('ClickHouseDatasource', () => {
         expect(result).toBeDefined();
         expect(result?.rawSql).not.toContain('LIKE');
       });
-
-      it('should use the configured LogMessage column name for non-OTel tables', () => {
-        jest.spyOn(logs, 'getTimeFieldRoundingClause').mockReturnValue('toStartOfInterval("ts", INTERVAL 1 DAY)');
-        const result = datasource.getSupplementaryLogsVolumeQuery(request, {
-          ...query,
-          builderOptions: {
-            ...query.builderOptions,
-            columns: [
-              { name: 'ts', hint: ColumnHint.Time },
-              { name: 'message', hint: ColumnHint.LogMessage },
-            ],
-            meta: { logMessageLike: 'oops' },
-          },
-        });
-        expect(result?.rawSql).toContain("( message LIKE '%oops%' )");
-      });
     });
 
     describe('getSupplementaryLogsSampleQuery', () => {
@@ -1653,21 +1637,6 @@ describe('ClickHouseDatasource', () => {
         }) as CHBuilderQuery;
         expect(result).toBeDefined();
         expect(result.rawSql).not.toContain('LIKE');
-      });
-
-      it('should use the configured LogMessage column name for non-OTel tables', () => {
-        const result = datasource.getSupplementaryLogsSampleQuery({
-          ...query,
-          builderOptions: {
-            ...query.builderOptions,
-            columns: [
-              { name: 'ts', hint: ColumnHint.Time },
-              { name: 'message', hint: ColumnHint.LogMessage },
-            ],
-            meta: { logMessageLike: 'oops' },
-          },
-        }) as CHBuilderQuery;
-        expect(result.rawSql).toContain("( message LIKE '%oops%' )");
       });
     });
 
--- a/src/data/CHDatasource.test.ts
+++ b/src/data/CHDatasource.test.ts
@@ -1418,6 +1418,81 @@ describe('ClickHouseDatasource', () => {
             `ORDER BY time ASC`
         );
       });
+
+      it('should apply the log message search using the real column name', () => {
+        jest
+          .spyOn(logs, 'getTimeFieldRoundingClause')
+          .mockReturnValue('toStartOfInterval("created_at", INTERVAL 1 DAY)');
+        const result = datasource.getSupplementaryLogsVolumeQuery(request, {
+          ...query,
+          builderOptions: {
+            ...query.builderOptions,
+            columns: [
+              { name: 'created_at', hint: ColumnHint.Time },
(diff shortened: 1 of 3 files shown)


src/data/CHDatasource.ts
```diff
@@ -312,6 +312,36 @@ export class Datasource
...
+      return undefined;
+    }
+
+    return {
+      condition: 'AND',
+      key: logMessageColumn.name,
+      type: 'string',
+      filterType: 'custom',
+      operator: FilterOperator.Like,
+      value: logMessageLike,
```
adamyeats, line 341: The search term is interpolated into the SQL unescaped, since the LIKE branch at sqlGenerator.ts:957 wraps the value in quotes without escaping, so a term containing a single quote breaks the generated query. The main logs query does the same at sqlGenerator.ts:422, so this is consistent rather than a regression, but it might be worth a follow-up to escape the value in both paths.

src/data/CHDatasource.ts
```diff
@@ -371,6 +401,11 @@ export class Datasource
       return f;
```
adamyeats: nit: not introduced here, but the map above mutates the original query's filter objects when it assigns `f.key`, whereas the sample query's version clones each filter with `{ ...f }`. While you are in the area, consider returning a clone here too so the supplementary query stops writing keys back into the user's query.

src/data/CHDatasource.test.ts
```diff
@@ -1418,6 +1418,81 @@ describe('ClickHouseDatasource', () => {
...
+            ...query.builderOptions,
+            columns: [{ name: 'created_at', hint: ColumnHint.Time }],
+            meta: { logMessageLike: 'found' },
+          },
+        });
+        expect(result).toBeDefined();
+        expect(result?.rawSql).not.toContain('LIKE');
+      });
+
+      it('should use the configured LogMessage column name for non-OTel tables', () => {
```
adamyeats: nit: this largely repeats the first new test in the block, which already proves the filter targets the real column name rather than the alias. Consider folding the two together, and the same applies to the sample query counterpart at line 1658.

adamyeats (dismissed):

src/data/CHDatasource.ts
alex-fedotyev: Good catch. Agreed it is pre-existing and consistent across both paths, so I kept it out of this PR to stay scoped and opened #2097 to escape the value in the LIKE branch and the main query's message clause together.

src/data/CHDatasource.ts
alex-fedotyev: Good call. Cloned the filter in the volume builder so resolving the key no longer writes back into the user's query, matching the sample builder's `{ ...f }`. Fixed in d66779a.

src/data/CHDatasource.test.ts
alex-fedotyev: Done. Dropped the redundant non-OTel column-name test in both blocks; the first test in each already locks in the real-column-name behavior. d66779a.

alex-fedotyev:
Thanks for the review. Addressed both nits in d66779a: the logs volume builder now clones each filter before resolving the key (matching the sample builder), so it no longer writes keys back into the user's query, and I dropped the duplicated non-OTel column-name tests since the first test in each block already covers the real-column-name behavior. Opened #2097 for the LIKE-escaping follow-up so the main query and the supplementary queries get it together.

The push reset your approval, so this needs another look when you have a moment.

adamyeats:
@alex-fedotyev Can you rebase this branch first? Once that's done, everything LGTM, so I'll approve and merge. 🚢

alex-fedotyev:
Rebased onto the latest main and both commits are signed. No conflicts (the intervening changes were backend and workflow only), and the suite is green locally. Ready for another look when you have a moment.

adamyeats (approved):
Thanks @alex-fedotyev !

PR #2094, opened 2026-07-22
Support JSON-typed columns in ad-hoc filters

AndreZiviani:
## Type of Change

- [x] 🚀 Feature

---

### What is this feature?

Ad-hoc filter (and filter-variable) support for ClickHouse **`JSON`-typed columns**, paralleling the existing `Map(...)` support.

- **Key discovery**: `getTagKeys` expands a `JSON` column into one selectable key per discovered path (via `JSONAllPaths`, or a companion `<col>Keys` array column when present, bounded by the existing `enableMapKeysDiscovery` toggle).
- **Value listing**: distinct values for a JSON path are listed via cast dot-access (`col.`a`.`b`::Nullable(String)`).
- **Filter emission**: ad-hoc filters emit JSON dot-access per column type, coexisting with `Map` columns on the same table, in both the automatic `additional_table_filters` path and the `$__adHocFilters` macro.
- Works in both **schema mode** (`$clickhouse_adhoc_query` = `db.table`) and **query mode** (`$clickhouse_adhoc_query` = a custom `SELECT ... $__adhoc_column ... FROM db.table WHERE ...`), so scoped value dropdowns and JSON path discovery work together.

JSON path access is built by a single shared helper (`src/data/jsonPath.ts`) now used by the query builder, the ad-hoc filter builder, and tag-value fetching, so all three stay consistent.

The newer OpenTelemetry ClickHouse schema stores `LogAttributes`/`ResourceAttributes`/`ScopeAttributes` as `JSON` (not `Map`). On that schema, ad-hoc filters previously showed only the bare column, couldn't list per-path values, and the filter SQL was invalid: JSON sub-paths are `Dynamic`, which `IN`/`NOT IN` reject and which read back all-null over the native protocol without a `::Nullable(String)` cast.

### Who is this feature for?

Users running the OTel JSON schema (or any table with `JSON` columns) who want to filter dashboards on JSON attributes via ad-hoc filters.

### How to test this feature

1. Point the datasource at a ClickHouse table with a `JSON` column (e.g. an OTel `otel_logs` with `LogAttributes JSON`).
2. Add an **Ad hoc filters** dashboard variable for the datasource, plus a `clickhouse_adhoc_query` variable set to `db.otel_logs`.
3. Open the ad-hoc filter key dropdown → JSON paths appear (e.g. `LogAttributes.k8s.container.name`); pick a key → the value dropdown lists distinct values; apply → the panel filters.
4. (Query mode) set `clickhouse_adhoc_query` to `SELECT DISTINCT $__adhoc_column FROM db.otel_logs WHERE ServiceName = '$service'` with a `$service` variable → the same works with values scoped to the selected service.

---

## Please check that:

- The change is TypeScript-only (frontend datasource); the Go backend already maps `JSON`/`Nullable(JSON)` to `FieldTypeJSON`.
- Cross-table safety: JSON/Map classification is resolved **per table**; a column name typed differently across tables (JSON in one, Map in another) is kept out of the global `escapeKey` JSON set so it retains prior Map behavior.
- Verified end-to-end against a live OTel JSON dataset, including SQL-injection probing of JSON path segments and Map keys (escaping is layered correctly for both the plain value SELECT and the nested `additional_table_filters` string).

### Known pre-existing follow-ups (not addressed here, planned separately)

- On initial dashboard load, a saved filter on a JSON column can hit the panel-apply path before column kinds are resolved and emit Map-bracket access → `ILLEGAL_TYPE_OF_ARGUMENT` (pre-existing on `main`; mitigated today by the `clickhouse_adhoc_use_json` variable). Needs synchronous/eager schema resolution.
- The ad-hoc value SELECT is unbounded (no time window), unlike key discovery.
- Non-`IN` filter values aren't escaped for interior quotes (pre-existing).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 - Add Forward OAuth Identity authentication: forward the signed-in Grafana user's OAuth token to ClickHouse Cloud as a JWT so queries are attributed to the real user instead of a shared service account. Requires a verified TLS connection. Alert and other backend queries have no signed-in user and are blocked by default; enable **Allow service account fallback** to let them run with the configured username/password (#1987)
+- Support JSON-typed columns (e.g. the OTel JSON schema's `LogAttributes`/`ResourceAttributes`) in ad-hoc filters: JSON sub-paths are discovered for the key dropdown and rendered as backtick-quoted dot-access cast to `Nullable(String)`, so every filter operator works and values read back over the native protocol. Keys are minted in a stateless self-describing form (building on #2079), so a saved JSON filter applies correctly on a fresh dashboard load (#2094)
 
 ### Fixes
 
--- a/src/data/CHDatasource.test.ts
+++ b/src/data/CHDatasource.test.ts
@@ -1024,6 +1024,95 @@ describe('ClickHouseDatasource', () => {
     });
   });
 
+  describe('JSON type ad-hoc filters (#2094)', () => {
+    it('expands JSON-typed columns into self-describing backtick path keys', async () => {
+      jest.spyOn(templateSrvMock, 'replace').mockImplementation(() => 'db.events');
+      const ds = cloneDeep(mockDatasource);
+      ds.settings.jsonData.defaultDatabase = 'db';
+      ds.settings.jsonData.defaultTable = 'events';
+      const columnsFrame = arrayToDataFrame([
+        { name: 'level', type: 'String', table: 'events' },
+        { name: 'ResourceAttributes', type: 'JSON', table: 'events' },
+      ]);
+      // fetchUniqueJSONPathsForAdhoc → distinct JSON paths for ResourceAttributes.
+      const pathsFrame = arrayToDataFrame([{ path: 'k8s.pod' }, { path: 'level' }]);
+      jest.spyOn(ds, 'query').mockImplementation((request) => {
+        const sql = request.targets[0].rawSql ?? '';
+        if (sql.includes('JSONAllPaths("ResourceAttributes")')) {
+          return of({ data: [pathsFrame] });
+        }
+        return of({ data: [columnsFrame] });
+      });
+
+      const keys = await ds.getTagKeys();
+      // Each level backtick-quoted: the key is self-describing, so escapeKey
+      // renders cast JSON access with no JSON-column cache.
+      expect(keys).toEqual([
+        { text: 'events.level' },
+        { text: 'events.ResourceAttributes.`k8s`.`pod`' },
+        { text: 'events.ResourceAttributes.`level`' },
+      ]);
+    });
+
+    it('reads a companion <col>Keys array column instead of JSONAllPaths when present', async () => {
+      jest.spyOn(templateSrvMock, 'replace').mockImplementation(() => 'db.events');
+      const ds = cloneDeep(mockDatasource);
+      ds.settings.jsonData.defaultDatabase = 'db';
+      ds.settings.jsonData.defaultTable = 'events';
+      const columnsFrame = arrayToDataFrame([
+        { name: 'ResourceAttributes', type: 'JSON', table: 'events' },
+        { name: 'ResourceAttributesKeys', type: 'Array(String)', table: 'events' },
+      ]);
+      const pathsFrame = arrayToDataFrame([{ path: 'service.name' }]);
+      const spyOnQuery = jest.spyOn(ds, 'query').mockImplementation((request) => {
+        const sql = request.targets[0].rawSql ?? '';
+        if (sql.includes('arrayJoin("ResourceAttributesKeys")')) {
+          return of({ data: [pathsFrame] });
+        }
+        return of({ data: [columnsFrame] });
+      });
+
+      const keys = await ds.getTagKeys();
+      expect(keys).toContainEqual({ text: 'events.ResourceAttributes.`service`.`name`' });
+      // Discovery read the companion array column, not JSONAllPaths.
+      const probedSql = spyOnQuery.mock.calls
+        .map((c) => (c[0] as any).targets[0].rawSql as string)
+        .find((s) => s.includes('arrayJoin('));
+      expect(probedSql).toContain('arrayJoin("ResourceAttributesKeys")');
+      expect(probedSql).not.toContain('JSONAllPaths');
+    });
+
+    it('fetches values for a minted JSON key with the cast, no prior getTagKeys call', async () => {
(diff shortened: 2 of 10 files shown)


adamyeats (commented):
@AndreZiviani Let's hold off on this one for a while. This seems like a natural follow-on  from #2079:  I think the stateless bracket-form approach in #2079 should land first, since this PR's lazy-cache design would inherit the race we already fixed. Let's park this for the time being, and we'll circle back again once that PR is landed. 👍

AndreZiviani:
hey @adamyeats I've rebased my PRs on the #2079 , please take a look when you can

SpencerTorres (approved):
Thanks for submitting this. I can see how this is useful even if we do recommend maps for OTel. My only nit at the moment is the comments. There are a lot of comments in this PR that seem excessive and are not really documentation but rather re-iterating what the code is already saying.

@adamyeats Let me know if you have any thoughts

AndreZiviani:
Thanks @SpencerTorres , I will adjust the comments

> I can see how this is useful even if we do recommend maps for OTel.

Based on the clickstack blog posts I believe that will change soon to JSON and text filter, it is already an opt-in config on the collector

SpencerTorres:
> Based on the clickstack blog posts I believe that will change soon to JSON and text filter, it is already an opt-in config on the collector

@AndreZiviani Can you send me a link to this blog post so I can understand the exact schema you're referencing? I want to make sure I have the right one. Thanks!

AndreZiviani:
Hey @SpencerTorres I was thinking about these two

https://clickhouse.com/blog/whats-new-clickstack-july-2025

https://clickhouse.com/blog/whats-new-in-clickstack-april-2026

But I've missed this one which looks like going to json was not such a good idea

> We've evaluated alternatives. The JSON type is particularly interesting because it can automatically materialize frequently accessed paths into dedicated subcolumns, reducing the amount of data that must be read at query time. In practice, however, the additional insert overhead is rarely justified for observability workloads where attributes are often sparse, highly variable, and continuously evolving. For now, maps remain the most practical representation for dynamic telemetry attributes.


https://clickhouse.com/blog/making-clickstack-5x-faster-clickhouse-observability

https://clickhouse.com/docs/clickstack/ingesting-data/schema/map-vs-json

SpencerTorres (changes requested):
Thanks for sharing that blog post, the snippet you shared sounds like our current recommendations. I've done a lot of benchmarking recently on the schema and it's actually been easier to improve Map performance than JSON performance for large scale OTel workloads.

I had an agent review this PR specifically for edge cases and it found a few things, let me know if any of these stand out to you:

> 1. `buildJSONPathAccess` casts to `Nullable(String)` unconditionally, so ad-hoc `<` `<=` `>` `>=` on a JSON path compare lexicographically. On 26.7.1, `attrs.num > '10'` matches both 42 and 7, and `< '10'` matches neither. No error, just wrong rows. `jsonStringOperators` in `filterOperatorOptions.ts` omits those four operators, so the query builder already sidesteps this.
> 2. `CHDatasource.ts:1804`: `tableKnown` never becomes true for a table with no Map or JSON column, so the cross-table guard its docblock describes doesn't fire.
> 3. `CHDatasource.ts:1971`: a column with the same name as its table produces `SELECT  FROM x`.
> 4. The new escaping and the cache warm-up aren't covered. Emptying `buildJSONAccessForOuterFilter` or deleting both `ensureAttributeColumnCache` call sites leaves all 1113 tests passing.
> 5. `jsonPath.ts:19` describes the column as a trusted `system.columns` name, which doesn't hold on the minted-key path.

AndreZiviani:
Thanks @SpencerTorres — I'm going back to maps schema but went through the agent's points:

1. **Lexicographic string compare on `<`/`>`/`<=`/`>=`**: this is pre-existing and shared with Map keys (both are cast to `String`), and the query builder already sidesteps it by omitting those operators for JSON (`filterOperatorOptions`). Left the ad-hoc behavior as-is to match Map — happy to add a doc note or restrict the operators if you'd prefer.
2. **`attributeColumnsForTable` guard**: fixed — "known" now keys off `attributeSchemaFetched`, so a fetched table with no Map/JSON column returns an empty set instead of the cross-table union.
3. **`col === table` → `SELECT  FROM x`**: fixed — a single-segment key equal to the table name is no longer stripped to empty (both schema- and query-mode paths).
4. **Coverage**: added tests pinning the outer-filter escaping and the cold-cache warm-up — mutation-checked (they fail if the escaping/warm-up is removed).
5. **Comment**: corrected — it no longer claims the column is a trusted `system.columns` name.

Also extracted the 6h probe window to a named constant. `tsc` + suite green.

AndreZiviani:
~I'm also submitting a security report which also affects this, so hold off on merging for now~

it was considered a non-issue

AndreZiviani:
Follow-up: `buildJSONPathAccess` now backtick-quotes any column that isn't a plain (optionally dotted) identifier, so a typed/unusual ad-hoc key yields valid SQL instead of splicing text into the filter expression — this ties off the "trusted identifier" note from the earlier comment thread. Plain schema columns are unchanged. The Map/default key paths get the same treatment in a separate follow-up PR.

SpencerTorres (approved):

adamyeats:
@AndreZiviani the head of this branch is an ancestor of #2095, so #2095 already contains all four commits here plus five more. Rather than rebase both, it might be worth closing this one and landing #2095 on its own.

Merging both costs more than it saves. A squash merge of this PR rewrites the commits, so #2095 then conflicts across four or five files that would otherwise rebase clean.

Happy to keep both open if you would rather keep the smaller review unit.

AndreZiviani:
Agreed — #2095 contains all four commits here plus five more, and I have just rebased and pushed it, so the argument for keeping both is gone. Not worth paying the squash-rewrite conflict across four or five files to land the smaller unit.

Leaving the close to you so you can confirm nothing here is worth keeping separately.

AndreZiviani:
Closing in favour of **#2095**, which supersedes this PR entirely.

I re-verified containment after rebasing #2095, since the ancestry no longer holds by SHA: `git cherry` marks all four commits here as having patch-equivalents on #2095, every one of the ten files touched is present there, and the net diff between the two branches only adds on top. Nothing here is lost.

What #2095 carries beyond this PR:

- value escaping for `IN` / `NOT IN` lists, closing a filter-injection path
- backtick-quoted JSON column identifiers, so typed and unusual keys stay valid SQL
- a bounded value scan on the key probe
- extra coverage for regex/backslash escaping and JSON path edge cases

The changelog entry for this work now lives on #2095 rather than referencing this PR number.

Review continues on #2095.

PR #2095, opened 2026-07-22
feat(adhoc): support JSON-typed columns and escape filter values

AndreZiviani:
## Type of Change

- [x] 🚀 Feature
- [x] 🐛 Bug Fix

### What is this feature?

Ad-hoc filters work against JSON-typed columns — the OTel JSON schema's `LogAttributes` / `ResourceAttributes` and any other `JSON` column. Sub-paths are discovered for the key dropdown and rendered as backtick-quoted dot access cast to `Nullable(String)`, so every operator works and values read back over the native protocol. Keys are minted in a stateless, self-describing form (building on #2079), so a saved JSON filter applies correctly on a fresh dashboard load rather than only while the column cache happens to be warm.

On the JSON schema the key dropdown offered nothing useful, and a hand-written JSON filter produced SQL that either failed or silently read all-null: JSON sub-paths are `Dynamic`, which `IN` / `NOT IN` reject outright, and an uncast sub-path reads as null over the native protocol.

### Who is this feature for?

Anyone on the OTel Collector's JSON schema, or with `JSON` columns of their own, who wants the same ad-hoc filtering the Map schema already had.

### How to test this feature

1. Point the data source at a table whose attribute columns use the native `JSON` type (the e2e fixture `e2e_test.json_events` works).
2. Add an ad-hoc filter variable to a dashboard and open the key dropdown: JSON sub-paths appear alongside ordinary columns.
3. Pick one, choose a value from the value dropdown, and confirm the panel filters.
4. Reload the dashboard and confirm the saved filter still applies (the key carries its own type).
5. Try `IN` and `NOT IN` with several values, including a value containing `'`.

### What is the bug?

Three separate problems in the ad-hoc filter path, all fixed here:

- **Filter values were not escaped** into the `additional_table_filters` clause. A value with unbalanced parentheses bypassed the filter entirely (no quote required), `NOT IN` was a silent no-op that returned every row, and values containing a backslash matched the wrong rows.
- **Filter keys were emitted verbatim**, so a key such as `ServiceName = 'x' OR 1=1 --` — and keys are URL-settable — spliced raw SQL into the filter predicate. Keys that are not plain or dotted identifiers are now backtick-quoted, so the query fails closed with `UNKNOWN_IDENTIFIER` instead. Grafana assessed this as robustness rather than a security boundary, since the same user can already run arbitrary queries.
- **The value-suggestion query scanned the whole column.** It is now bounded by the dashboard time range, falling back to a recent window when no range is supplied, on the configured OTel logs/traces table.

### Related Issues

Supersedes #2094, which is closed — this branch carries that work as well.

## Please check that:

Rebased onto latest main, which now includes #1936 (ad-hoc table resolution through the ClickHouse parser). The conflicts were in `adHocFilter.ts`, `adHocFilter.test.ts` and `CHDatasource.ts`; resolution keeps #1936's parser-based `apply()` and per-call table resolution, and layers this branch's key/value escaping on top. `escapeKey` now shares the identifier-quoting rule with the JSON path rather than carrying its own copy.

Two corrections to what an earlier version of this description said:

- The value scan honours the dashboard time range, with the recent window only as a fallback.
- The structured `filter.values` branch is not reachable from the UI here, since `plugin.json` declares no `multiValueFilterOperators`; it is covered by unit tests so the behaviour is pinned if that changes.

The e2e spec asserts that ClickHouse accepts the SQL shapes this path emits — the SQL generation itself is covered by unit tests — and its `describe` now says so.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nduytg:
Hi team, thanks for the amazing work on the plugin! We love it! 

We're heavy users of the plugin for our self-hosted ClickHouse fleet, and most queries are instant. However, the only slow one is the label lookup, since it scans through our entire table.

For extra context: we retain 1 month of OTel logs as hot data on EBS volumes, then gradually move them to S3 for cold storage. Every time our users open the label dropdown, it scans the entire table including the historical data on S3, which makes it extremely slow.

The 6h bound on the value-suggestion query in this PR would fix this outright for us. We'd love to see it get merged, and we're happy to share more about our use case too! @AndreZiviani @alyssajoyner

SpencerTorres (changes requested):
Thanks for this PR, I see it stacks on #2094. I should warn you, our ad-hoc filters are notoriously hard to work with, so I appreciate you trying to fix these edge cases.

Given the complexity of our ad-hoc filters, I had my ad-hoc filter agent review for edge cases and it found these points:

> 1. `parseInListItems` splits on every comma with no quote awareness, so an IN element containing a comma gets torn in half and the stray apostrophes are escaped as data. `IN "'gzip, deflate','identity'"` becomes three items instead of two, emits valid SQL, and returns nothing. grafana/scenes leaves `values` undefined for `IN`/`NOT IN`, so every real IN filter takes this path.
> 
> 2. Nothing pins quote or backslash escaping on that path. The security test payload has no `'` and no `\`, so the escaping is a no-op on it. A mutant that escapes only `f.values` elements passes all 1129 tests and still breaks out.
> 
> 3. `getTagValues` destructures only `{ key }`, but `DataSourceGetTagValuesOptions` carries `timeRange`. Values older than 6h now disappear from the dropdown regardless of the dashboard range, with no opt-out or hint.
> 
> 4. The 6h window is hardcoded at a third site now, `:1790` alongside `:1215` and `:1253`, while the sibling bound is already a named constant.
> 
> 5. The new bound test uses `toContain`, so clause placement isn't pinned. Putting the where clause after the limit survives the suite and is a syntax error live.
> 
> 6. CHANGELOG says the trigger is a value containing `'`, but the payload in your description has no quote and breaks out on unbalanced parens. Also worth noting that `NOT IN` was a silent no-op returning every row, and `=` on backslash values returned wrong rows.

Let me know if any of these points stand out for your use case, we could put them in a future PR if needed.

AndreZiviani:
Thanks @SpencerTorres — addressed all six:

1. `parseInListItems` is now quote-aware (splits only on commas outside quotes), so an `IN` value like `'gzip, deflate'` stays a single element.
2. Added an escaping test on the string-parse path (a value containing `'`) — the previous test's payload had none, so it couldn't catch a dropped escape.
3. The value lookup now honors the dashboard `timeRange`, falling back to the recent window only when none is supplied, so in-range values aren't hidden.
4. The 6h window is now a named constant, shared with the key-discovery probe.
5. The bound test now pins `WHERE`-before-`LIMIT` ordering, and I added a `timeRange` test.
6. CHANGELOG reworded to describe the real triggers (unbalanced parens, `NOT IN` returning every row, backslash values), not just quotes.

`tsc` + suite green.

SpencerTorres (approved):

adamyeats:
@AndreZiviani this branch is behind main, and main now requires branches to be up to date before merge. The rebase itself is clean, with no conflicts.

One thing is worth folding into the same push so CI only runs once: `cspell` fails on `uncast` and `statelessly`. Adding both words to `cspell.config.json` clears all ten hits.

This branch already contains everything in #2094, so landing it on its own covers both.

AndreZiviani:
Rebased onto latest main and force-pushed — clean, as you said, no conflicts.

Folded the cspell fix into the same push: `uncast` and `statelessly` added to `cspell.config.json`, which clears all ten hits. `npm run spellcheck` is now clean, along with the full suite and typecheck.

src/data/CHDatasource.ts
```diff
@@ -1733,6 +1813,110 @@ export class Datasource
...
+    const table = source.includes('.') ? source.split('.')[1] : source;
+    if (!table || this.attributeSchemaFetched.has(table)) {
+      return;
+    }
+    const db = source.includes('.') ? source.split('.')[0] : this.getDefaultDatabase();
+    // Mark fetched up front so concurrent value lookups don't double-probe.
+    this.attributeSchemaFetched.add(table);
+    try {
+      this.skipAdHocFilter = true;
+      const rawSql = `SELECT name, type FROM system.columns WHERE database = '${db}' AND table = '${table}'`;
```
adamyeats, line 1900: `db` and `table` are interpolated into the `system.columns` probe without `escapeCHStringLiteral`, which this file already uses for Map keys. In schema mode with the table name shown, `table` is the first dotted segment of the typed key, so a `'` in the key breaks the probe. The catch degrades it safely, but the line is new and the fix is two tokens. Consider escaping both.

src/data/CHDatasource.ts
```diff
@@ -1709,22 +1775,36 @@ export class Datasource
...
+    const valueTable = source.includes('.') ? source.split('.')[1] : source;
+    const valueDb = source.includes('.') ? source.split('.')[0] : this.getDefaultDatabase();
+    const timeColumn = valueDb ? this.getMapKeyProbeTimeColumn(valueDb, valueTable) : undefined;
+    let whereClause = '';
+    if (timeColumn) {
+      const tc = escapeIdentifier(timeColumn);
+      const fromMs = timeRange?.from?.valueOf?.();
+      const toMs = timeRange?.to?.valueOf?.();
+      whereClause =
+        typeof fromMs === 'number' && typeof toMs === 'number'
```
adamyeats, line 1803: nit: `typeof fromMs === 'number'` admits `NaN` from an invalid `TimeRange`, which emits `fromUnixTimestamp(NaN)`. `Number.isFinite` on both bounds is the same length and falls back to the recent window instead.

src/data/adHocFilter.ts
```diff
@@ -169,26 +191,67 @@ function escapeKey(s: string, isJSON = false, mapColumns: ReadonlySet<string> =
...
+function escapeValueBasedOnOperator(s: string, operator: string, values?: string[]): string {
+  if (operator === 'IN' || operator === 'NOT IN') {
+    // Build the list from the structured `values` array when Grafana provides it
+    // (multi-select), otherwise best-effort split of the legacy joined string.
+    // Every element is escaped and re-quoted, so no element can break out of the
+    // filter clause — ClickHouse coerces quoted numerics, so numeric IN still
+    // works. Empty list → `(NULL)` so `IN`/`NOT IN` stay valid SQL.
+    const items = values && values.length > 0 ? values : parseInListItems(s);
+    if (items.length === 0) {
+      return '(NULL)';
```
adamyeats, line 213: nit: the empty `IN` and `NOT IN` list emits `(NULL)`, which is the right ClickHouse spelling, but nothing pins it. Changing it to `()` leaves all 372 tests green. A one-line test per operator would lock it in.

tests/e2e/jsonAdhocFilter.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+
+// ---------------------------------------------------------------------------
+// JSON-typed adhoc filter regression guards
+//
+// Unit tests in src/data/adHocFilter.test.ts cover the SQL shape escapeKey()
+// emits when a dotted key references a JSON column, and unit tests in
+// src/data/CHDatasource.test.ts cover getTagKeys() JSON-path expansion and the
+// fetchTagValuesFromSchema() rewrite. Those tests can't confirm ClickHouse
+// actually accepts the resulting SQL strings — only E2E can.
+//
```
adamyeats, line 66: nit: this spec hand-types SQL into Explore and never creates an ad-hoc filters variable, so it exercises ClickHouse rather than the escaping code in this PR. The unit tests cover that well, so it is fine to keep. The name and header comment could say what it does test.

src/data/CHDatasource.test.ts
```diff
@@ -1024,6 +1024,192 @@ describe('ClickHouseDatasource', () => {
...
+        }
+        return of({ data: [columnsFrame] });
+      });
+
+      const timeRange = { from: { valueOf: () => 1_600_000_000_000 }, to: { valueOf: () => 1_600_086_400_000 } };
+      await ds.getTagValues({ key: 'otel_logs.ResourceAttributes.`k8s`.`pod`', timeRange } as any);
+      const valueSql = spyOnQuery.mock.calls
+        .map((c) => (c[0] as any).targets[0].rawSql ?? '')
+        .find((s: string) => s.includes('select distinct'));
+      expect(valueSql).toContain('where "Timestamp" >= fromUnixTimestamp(1600000000) and "Timestamp" <= fromUnixTimestamp(1600086400) limit 1000');
```
adamyeats, line 1208: nit: `prettier --check` flags this line. Not a CI gate here, so only worth a `prettier --write` if you push again.

adamyeats (dismissed):
The body still describes this as three follow-up commits stacked on #2094, with review deferred until that PR merges. #2094 is closed and this branch now carries the whole JSON ad-hoc feature, so the body no longer matches the diff. It might be worth a refresh: tick Feature, and lift the what and why from the two CHANGELOG entries.

Two details to correct at the same time. The value scan now honours the dashboard time range, with the recent window as fallback. And the structured `filter.values` branch is not reachable from the UI here, since `plugin.json` declares no `multiValueFilterOperators`.

AndreZiviani:
Thanks @adamyeats — rebased onto latest main, which now carries #1936, and all five addressed.

**On the rebase.** #1936 reshaped the same file, so this was more than a replay: conflicts in `adHocFilter.ts`, `adHocFilter.test.ts` and `CHDatasource.ts`. The resolution keeps #1936's parser-based `apply()` and its per-call table resolution untouched, and layers this branch's key/value escaping on top — `escapeMapKeyForOuterFilter` had already widened to `escapeForOuterFilterLiteral` here, and `escapeRegExp` from #1936 sits alongside it. The expectation strings picked up the new `\nsettings` placement.

**Description.** Rewritten from the CHANGELOG entries, with Feature and Bug Fix both ticked, and your two corrections folded in: the value scan honours the dashboard time range with the recent window as fallback, and the structured `filter.values` branch is not UI-reachable since `plugin.json` declares no `multiValueFilterOperators` (kept under unit test so it stays pinned if that changes).

**The nits.**

- `db` and `table` in the `system.columns` probe now go through `escapeCHStringLiteral`, like the Map keys in the same file.
- `Number.isFinite` on both bounds instead of `typeof === 'number'`, so an invalid `TimeRange` falls back to the recent window rather than emitting `fromUnixTimestamp(NaN)`.
- The empty `IN` / `NOT IN` rendering is pinned — `(NULL)` for both operators.
- The e2e `describe` now says what it asserts: that ClickHouse accepts the SQL shapes this path emits, with generation covered by the unit tests.
- `prettier --write` run over the touched files.

**One addition.** Since #1936 had already reshaped this file, I folded in the ad-hoc *key* quoting rather than leaving a third pass over the same code for later: `escapeKey` now routes every identifier through the same rule the JSON path already used, so a key that is not a plain or dotted identifier is backtick-quoted and a crafted key like `ServiceName = 'x' OR 1=1 --` reads as one nonexistent column instead of being spliced into the predicate. Plain and dotted names are unchanged. Grafana looked at this and assessed it as ordinary robustness, not a security boundary, since the same user can already run arbitrary queries — so it is framed that way in the CHANGELOG. Verified by mutation: reverting the call makes the new guard fail.

Suite green (1584 passing) apart from `TLSSSLSettingsSection`, which fails on clean main too.

adamyeats (approved):
Checked the 21 September push against every open thread. All addressed, including the new key quoting. Approving.

src/data/CHDatasource.ts
adamyeats: Addressed in the 21 September push (874f9e6). Resolving.

tests/e2e/jsonAdhocFilter.spec.ts
adamyeats: Addressed in the 21 September push (874f9e6). Resolving.

src/data/CHDatasource.test.ts
adamyeats: Addressed in the 21 September push (874f9e6). Resolving.

src/data/CHDatasource.ts
adamyeats: Addressed in the 21 September push (874f9e6). Resolving.

src/data/adHocFilter.ts
adamyeats: Addressed in the 21 September push (874f9e6). Resolving.

PR #2108, opened 2026-08-06
feat(logs): browse and filter by any log table column

alex-fedotyev:
## Type of Change

- [x] 🚀 Feature

---

### What is this feature?

A new **Columns** setting in the logs section of the data source config lets you pick any table
column to browse and filter by, without editing each query. The chosen columns are folded into the
log labels under their real names, so they appear:

- in the Fields list on the left,
- in the log row details under a **Fields** group next to the **Resource attributes** /
  **Log attributes** groups, each with filter-for and filter-out buttons,
- and as real column filters (a filter-for click resolves in both the main and logs-volume queries).

In **single source** mode the config is schema-aware: the **Columns** field is a picker of the
table's actual columns with a one-click **Add all columns**, and the time, level, and message roles
are dropdowns of the table's columns (you can still type an expression). In **all databases** mode
you get the same result by choosing columns per query in the builder, which already has a Columns
field.

Whether the attribute groups render as separate named sections or flat under a single **Fields**
list depends on the Grafana version's log-details rendering; filtering and the folded columns behave
the same either way. The OTel attribute maps keep their existing grouped rendering, unchanged.

Today the logs view only surfaces the role columns (time, level, message) and the OTel attribute
maps as browsable, filterable fields. Any other column, for example `ServiceName` on an OTel table,
or `status` and `method` on a flat access-log table, is only visible if you hand-edit the query, and
even then it does not appear in the log details with filter buttons. People who explore logs by
clicking rather than writing SQL cannot see or filter by those columns.

This lets an operator expose the columns they care about once, on the data source, and browse and
filter by them across every logs query, the same way attributes already work. It works for OTel and
non-OTel log tables (ClickHouse `system.query_log`, nginx or other flat access logs, and so on).

### Who is this feature for?

Anyone using the ClickHouse data source for logs, especially point-and-click users in Explore who
filter by clicking values, and operators who want a consistent set of fields across dashboards
without editing each panel.

### How to test this feature

1. Open a ClickHouse data source set up for logs (Connections -> Data sources), and in the logs
   config, under **Columns**, add a few columns (in single source mode you can pick from the list or
   click **Add all columns**). Save and test.
2. In Explore, run a logs query against that data source.
3. Confirm the extra columns appear in the Fields list on the left and can be checked to show as
   columns.
4. Open a log row's details and confirm the extra columns appear under a **Fields** group next to
   **Log attributes** / **Resource attributes**, each with filter-for and filter-out buttons.
5. Click **Filter for value** on one and confirm the query gains a real column filter (for example
   `ServiceName = '...'`) and the logs volume histogram still loads.
6. Repeat against a non-OTel table (for example `system.query_log`) to confirm it is not OTel
   specific.

### Notes for reviewers

This supersedes and replaces #2086. Compared to that draft:

- The **Include all columns** toggle is dropped. Per the review discussion on #2086, the explicit
  column list (now just **Columns**) is the better primitive: it works in every mode, needs no
  query-time schema, and does not widen every query on a large table. **Add all columns** covers the
  "I want them all" case as a one-click action instead.
- Column selection lives on the data source config, and in single source mode it is schema-aware
  (real column pickers plus role dropdowns). There is no separate per-query column picker in the
  compact editor.
- The earlier review fixes are carried forward: the fold skips aliased columns and folds under the
  real name, `JSON(...)` is treated as a collection type, the per-row `labels` serialize is done
  once, and the compact editor still initializes when the schema fetch fails. New tests cover the
  **Add all columns** skip rules and the schema-backed config field.
- The single-source config reads the table's columns with the datasource's existing `fetchColumns`
  (the same `DESC TABLE` the query builder uses), resolved via `getDataSourceSrv().get(uid)`. It
  falls back to a typed list when the datasource is unsaved or the fetch fails. `getDataSourceSrv`
  is soft-deprecated in favor of an unstable API, so the one lint is suppressed with a note rather
  than depending on the unstable module.
- The selected-column chip wrap in `ColumnsEditor` (a fixed max-width applied via the react-select
  value-container class) is kept builder-wide, carrying forward the discussion on #2086; it now also
  applies to the config Columns field. Happy to scope it to logs or restyle it if you would prefer.

### Screenshots / Videos

Single source config: the Columns picker, the role dropdowns, and the description.

The one-click "Add all columns" entry and the table's real columns in the dropdown.

Not OTel specific: the same picker against `system.query_log`.

In Explore: the configured columns in the Fields list and a filter-for click on one.

### Notes

- The **Columns** editor is intentionally editable with **Use OTel** on (the `disabled={otelEnabled}` guard was dropped), so extra columns can be added on an OTel source.
- The log-message name heuristic change (auto-detecting `query` and `request` columns) is not part of this PR. It changes role detection for every logs table, so it will land as its own follow-up.


SpencerTorres (approved):
useful feature! My only nit is the excessive comments in some places, perhaps we can handle these in a future PR since they have been accumulating recently

katebrenner:
Hey Alex! Before I approve, can you update documentation? query-editor.md and configure.md - thanks!

alex-fedotyev:
Thanks both!

@katebrenner docs are in now, covering the two files you called out:

- `configure.md` (Logs configuration): added a **Columns** row to the settings table, plus a short paragraph on how it behaves. In Single source mode it's a picker of the selected table's columns with an **Add all columns** shortcut; other modes take a typed list. It works for both OTel and non-OTel log tables.
- `query-editor.md` (Logs query builder): added a note that columns set via the **Columns** setting are projected into every logs query automatically and show up as browsable, filterable fields, linking back to the config section.

I also rebased on main so the branch picks up the autocomplete macros fix (#2153). That clears the E2E failures, which were unrelated to this change.

@SpencerTorres good call on the comment density. I'll trim it in a follow-up so this PR stays focused on the feature.

src/data/utils.ts
```diff
@@ -719,3 +720,102 @@ export const dataFrameHasLogLabelWithName = (frame: DataFrame | undefined, name:
...
+        folded = true;
+      }
+
+      if (folded) {
+        labelsField.values[i] = wasString ? JSON.stringify(obj) : obj;
+      }
+    }
+
+    const foldedNames = new Set(sourceFields.map((f) => f.name));
+    frame.fields = frame.fields.filter((f) => f === labelsField || !foldedNames.has(f.name));
```
adamyeats, line 886: This deletes the folded fields from the frame, and log-context columns are hint-less too, so `getLogContextColumnsFromLogRow` can no longer match them by name. On a datasource with **Auto-Select Columns** on and **Context Columns** = `service`, **Show context** now fails with _Unable to match any context columns_, and it fails with `additionalColumns` empty.

src/data/utils.ts
```diff
@@ -719,3 +720,102 @@ export const dataFrameHasLogLabelWithName = (frame: DataFrame | undefined, name:
...
+    const sourceFields = discovered
+      .map((name) => frame.fields.find((f) => f.name === name && f.name !== labelsFieldName))
+      .filter((f): f is DataFrame['fields'][number] => f !== undefined);
+    if (sourceFields.length === 0) {
+      continue;
+    }
+
+    const rowLen = frame.length ?? frame.fields[0]?.values.length ?? 0;
+
+    let labelsField = frame.fields.find((f) => f.name === labelsFieldName);
```
adamyeats: A table with its own `labels` column is picked as the fold target here, and because `JSON.parse` of a non-JSON value falls back to `{}`, the row's real value is replaced rather than merged. Consider skipping the fold when an existing `labels` field is not the backend-serialised attribute map.

src/components/queryBuilder/compactQueryDefaults.ts
```diff
@@ -10,6 +17,25 @@ import {
...
+  datasource: Datasource,
+  allColumns: readonly TableColumn[],
+  columns: SelectedColumn[],
+  includedColumns: Set<string>
+): void => {
+  for (const columnName of datasource.getAdditionalLogColumns()) {
+    if (includedColumns.has(columnName) || includedColumns.has(columnName.split('[')[0])) {
+      continue;
+    }
+    columns.push({ name: columnName });
```
adamyeats: The pushed column carries no `type`, so `isCollectionColumnType` in the fold can never fire for a configured column and a `Map` or `Array` renders as `[object Object]`. The `allColumns` parameter on line 26 is already here and unused, so looking the type up from it would fix this at the source.

src/data/utils.ts
```diff
@@ -719,3 +720,102 @@ export const dataFrameHasLogLabelWithName = (frame: DataFrame | undefined, name:
...
+        obj = {};
+      }
+
+      let folded = false;
+      for (const src of sourceFields) {
+        const v = src.values[i];
+        if (v === null || v === undefined || v === '') {
+          continue;
+        }
+        obj[src.name] = typeof v === 'string' ? v : String(v);
```
adamyeats, line 876: `String(v)` on a time field gives a raw epoch value, so a configured `created DateTime` column shows as `1710495000000` in the row details and filters on that literal. Consider formatting by field type, or leaving non-string scalars to the field's own display processor.

src/data/utils.ts
```diff
@@ -719,3 +720,102 @@ export const dataFrameHasLogLabelWithName = (frame: DataFrame | undefined, name:
...
+    }
+
+    // The columns the field options selected are the hint-less, non-collection ones. Role columns
+    // (time / body / level / trace id) carry a hint; the attribute maps are collection-typed.
+    // A column aliased to a different name is skipped: a filter-for click would key on the alias,
+    // which the logs-volume query cannot resolve. A column whose alias equals its name (how the
+    // query builder tags a plain column pick) folds like an unaliased one, so columns chosen in the
+    // builder become browsable fields the same way columns configured on the datasource do.
+    const discovered = (builderOptions.columns || [])
+      .filter((c) => c.hint === undefined && !isCollectionColumnType(c.type) && (!c.alias || c.alias === c.name))
```
adamyeats: The fold keys on any hint-less column rather than on the new setting, so it also swallows columns picked by hand in the logs builder, which has had a Columns field for a long time. An existing Table panel on a logs builder query loses those columns and gets a single `labels` blob instead.

src/data/sqlGenerator.ts
```diff
@@ -402,6 +402,16 @@ const generateLogsQuery = (_options: QueryBuilderOptions): string => {
...
   const selectPartsSql = selectParts.join(', ');
 
+  // A logs query with no selected columns would generate `SELECT  FROM table`, which
+  // ClickHouse rejects with a syntax error. This happens transiently in the compact
+  // editor on a cold load: a non-OTel table has no columns until the schema fetch
+  // resolves and the default-column hooks populate the roles / include-all set. Emit
+  // no query in that window so Grafana skips it rather than issuing an invalid one; a
+  // valid query is generated on the next render once the columns arrive.
+  if (selectParts.length === 0) {
+    return '';
```
adamyeats, line 412: Paired with the new `filterQuery` guard this turns a broken logs query into silence: a non-OTel table whose columns miss the naming heuristics renders **No data** with no error and no SQL. `main` at least surfaced the ClickHouse syntax error, so consider showing the user why nothing ran.

src/data/CHDatasource.ts
```diff
@@ -1453,7 +1462,15 @@ export class Datasource
...
-    return !query.hide;
+    if (query.hide) {
+      return false;
+    }
+
+    // Skip queries with no SQL to run. A logs builder query carries an empty rawSql until its
+    // columns resolve (for example a non-OTel table in the compact editor before the schema fetch
+    // returns); running it would send an empty statement that ClickHouse rejects with a 400. Once
+    // the columns arrive the generated SQL is non-empty and the query runs normally.
+    return Boolean(query.rawSql && query.rawSql.trim().length > 0);
```
adamyeats: The comment describes a transient logs-builder state, but the guard applies to every query type, so an empty SQL-editor query is now skipped silently as well. Consider narrowing it to builder logs queries.

src/components/configEditor/LogsConfig.tsx
```diff
@@ -67,6 +78,111 @@ export const LogsConfig = (props: LogsConfigProps) => {
...
+            setFetchedColumns([]);
+          }
+        });
+      /* eslint-enable @typescript-eslint/no-deprecated */
+    }, 400);
+    return () => {
+      ignore = true;
+      clearTimeout(handle);
+    };
+  }, [isSingleTable, props.uid, defaultDatabase, defaultTable]);
```
adamyeats, line 128: A fetch that fails before the connection is saved is never retried, not even by **Save & test**, so the schema-backed controls stay hidden for the rest of the session. Consider keying the effect on something that changes when the datasource is saved.

src/components/configEditor/LogsConfig.tsx
```diff
@@ -67,6 +78,111 @@ export const LogsConfig = (props: LogsConfigProps) => {
...
+  // columns are known, falling back to a typed input before the datasource is saved. The select
+  // still allows a typed value, so expression-based roles keep working.
+  const renderRoleColumn = (
+    hint: ColumnHint,
+    value: string | undefined,
+    onChange: (v: string) => void,
+    columnFilterFn: (c: TableColumn) => boolean,
+    fieldLabels: { label: string; tooltip: string }
+  ) => {
+    if (fetchedColumns.length > 0) {
```
adamyeats: The role fields swap control type when the debounced fetch resolves, so text typed into **Log Message column** in the first second or two is replaced by a `ColumnSelect` mid-edit. Consider rendering the select from the start with a loading state, so the control never changes shape under the cursor.

src/components/configEditor/LogsConfig.tsx
```diff
@@ -67,6 +78,111 @@ export const LogsConfig = (props: LogsConfigProps) => {
...
+  // (time/level/message) are configured separately and projected via their roles, so Add all skips
+  // them to avoid double projection.
+  const roleColumnNames = new Set(
+    [filterTimeColumn, timeColumn, levelColumn, messageColumn].filter((c): c is string => Boolean(c))
+  );
+  const selectedColumns: SelectedColumn[] = (additionalColumns || []).map((name) => ({
+    name,
+    type: fetchedColumns.find((c) => c.name === name)?.type,
+  }));
+  const onSelectedColumnsChange = (cols: SelectedColumn[]) => onAdditionalColumnsChange(cols.map((c) => c.name));
```
adamyeats: `ColumnsEditor` sets `allowCustomValue`, so a pasted `ServiceName ` is stored untrimmed here while the `TagsInput` fallback trims through `onAdditionalColumnsChangeTrimmed`. Worth reusing the trimming callback on both paths.

src/components/configEditor/LogsConfig.tsx
```diff
@@ -67,6 +78,111 @@ export const LogsConfig = (props: LogsConfigProps) => {
   const onContextColumnsChangeTrimmed = (columns: string[]) =>
     onContextColumnsChange(columns.map((c) => c.trim()).filter((c) => c));
 
+  const onAdditionalColumnsChangeTrimmed = (columns: string[]) =>
```
adamyeats, line 81: A column that does not exist on the table breaks every logs query for the datasource with `Unknown expression identifier`, and this free-text path has no validation behind it. Consider checking the list against the schema on **Save & test** whenever the columns are known.

src/components/configEditor/LogsConfig.test.tsx
```diff
@@ -274,16 +315,228 @@ describe('LogsConfig', () => {
...
+          logsConfig={{ defaultTable: 'otel_logs', otelEnabled: false }}
+        />
+      );
+
+      // Let the 400ms debounce elapse and the guarded promise settle, then assert the fallback held:
+      // role fields are still text inputs, the Columns tags input is still present (2 tags inputs),
+      // and no schema-backed combobox rendered.
+      await waitFor(
+        () => {
+          expect(result.getByPlaceholderText(rolePlaceholders[1])).toBeInTheDocument();
```
adamyeats: Both schema-fetch fallback tests assert a condition that already holds at first render, so `waitFor` resolves before the 400 ms debounce fires and neither exercises the path it names. Making the guard return a column list leaves both green, as does removing the `!props.uid` guard for the test at line 403.

src/data/utils.test.ts
```diff
@@ -1263,3 +1264,201 @@ describe('dataFrameHasLogLabelWithName', () => {
...
+
+  it('creates a labels field when the frame has none (non-OTel table)', () => {
+    const ds = withFeature(['method', 'status']);
+    const [req, res] = buildLogsRequestResponse(
+      [
+        { name: 'event_time', hint: ColumnHint.Time },
+        { name: 'method', type: 'String' },
+        { name: 'status', type: 'UInt16' },
+      ],
+      [field('event_time', [1]), field('method', ['GET']), field('status', ['200'])]
```
adamyeats: `status` is described as `UInt16` but the fixture supplies the string `'200'`, and every fold fixture is single-row and string-only, so neither `String(v)` nor the per-row loop is covered. Replacing the coercion with `obj[src.name] = v` leaves all eight tests green.

docs/sources/configure.md
```diff
@@ -210,10 +210,13 @@ The data source includes a dedicated configuration section for log queries. Thes
 | **Filter Time column**   | A lower-precision time column for fast partition-based filtering. Used with the `1.2.9` schema only.                                                                                                                                                        |
 | **Log Level column**     | The column containing the log severity level.                                                                                                                                                                                                               |
 | **Log Message column**   | The column containing the log message body.                                                                                                                                                                                                                 |
+| **Columns**              | Extra columns projected into every logs query; each appears in the Fields list, the expanded log row, and the field filters, and can be shown as a displayed field.                                                                                         |
```
adamyeats: "Every logs query" overstates it: the columns are only injected while the builder is seeding defaults for a new query, so an existing dashboard panel keeps its saved column list. Consider "new logs queries" here and in the matching sentence at `query-editor.md:117`.

src/components/queryBuilder/views/columnNameHeuristics.ts
```diff
@@ -19,7 +19,10 @@ import { ColumnHint, TableColumn } from 'types/queryBuilder';
 export const columnNameHeuristics: Partial<Record<ColumnHint, RegExp>> = {
   // Logs
   [ColumnHint.Time]: /^(timestamp|@timestamp|time|event_?time|log_?time|created_?at)$/i,
-  [ColumnHint.LogMessage]: /^(body|message|msg|log_?message)$/i,
+  // `query` covers ClickHouse / DB query logs (system.query_log) and `request` covers HTTP
+  // access logs (nginx and similar), so those non-OTel sources get a message role without
+  // manual configuration. Both are unambiguous message columns in a logs context.
+  [ColumnHint.LogMessage]: /^(body|message|msg|log_?message|query|request)$/i,
```
adamyeats: nit: `findColumnByNameHeuristic` iterates columns rather than patterns, so on a table declared as `(ts, level, request, message)` the new `request` alias wins over `message`. Worth trying the canonical names first, since this changes role detection for every logs table, not just the new feature.

src/components/queryBuilder/ColumnsEditor.tsx
```diff
@@ -1,17 +1,40 @@
...
 import { styles } from 'styles';
+import { isCollectionColumnType, isDateTimeColumn } from './views/columnNameHeuristics';
+
+// Let the selected-column chips wrap onto multiple lines within a bounded width, instead of forming
+// one long row that fills the builder edge to edge (and clips when many columns are selected, for
+// example logs with "Include all columns" on). The dropdown menu still lists every column. The
+// builder's column width is driven by its widest row, so a bounded value list also keeps the whole
+// builder from stretching to fit a long column list.
+const wrappedColumnValues = css`
+  & [class*='grafana-select-value-container'] {
```
adamyeats, line 17: nit: this couples the builder to a `@grafana/ui` internal class name, so the wrapping would drop silently on a react-select upgrade. Scoping it to the logs and config fields, as you offered in the description, would limit the blast radius.

src/components/queryBuilder/views/LogsQueryBuilder.tsx
```diff
@@ -154,10 +160,11 @@ export const LogsQueryBuilder = (props: LogsQueryBuilderProps) => {
         testIdLink={allSelectors.QueryBuilder.LogsQueryBuilder.columnRolesHelpLink}
       />
       <ColumnsEditor
```
adamyeats, line 162: nit: dropping `disabled={builderState.otelEnabled}` here makes the Columns editor editable with OTel on, which is a behaviour change the description does not mention. Worth calling out in the PR body if it is intended.

src/views/config-v2/tracking.ts
```diff
@@ -80,6 +80,7 @@ export const trackClickhouseConfigV2LogsConfig = (props: {
   messageColumn?: string;
   selectContextColumns?: boolean;
   contextColumns?: string[];
+  additionalColumns?: string[];
```
adamyeats, line 83: nit: `reportInteraction` ships these names verbatim, and one **Add all columns** click on a wide table sends the customer's whole schema in a single event. Consider reporting a count rather than the names.

src/labels.ts
```diff
@@ -343,6 +343,14 @@ export default {
             label: 'Log Message column',
             tooltip: 'Column for log message',
           },
+          additionalColumns: {
+            label: 'Columns',
+            tooltip:
+              'Columns to show as log fields, in addition to the time, level, and message columns above. Each one appears in the Fields list, can be filtered, and can be added as a displayed column. Use "Add all columns" in the dropdown to add every column at once. Leave empty to add none.',
```
adamyeats: nit: `onAddAll` skips DateTime, collection-typed and `__`-prefixed columns, so "every column" overpromises. Worth saying it adds the scalar columns.

src/components/configEditor/LogsConfig.tsx
```diff
@@ -100,38 +216,55 @@ export const LogsConfig = (props: LogsConfigProps) => {
...
+        )}
+        {renderRoleColumn(
+          ColumnHint.LogMessage,
+          messageColumn,
+          onMessageColumnChange,
+          columnFilterString,
+          labels.columns.message
+        )}
+        {fetchedColumns.length > 0 ? (
+          <ColumnsEditor
```
adamyeats, line 255: nit: `ColumnsEditor` hardcodes `InlineFormLabel width={8}` and the generic builder tooltip, so once the schema resolves this row shifts left against the `width={12}` rows above it and loses the tooltip written for it. The fallback branch below keeps both.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Features
 
 - Add Forward OAuth Identity authentication: forward the signed-in Grafana user's OAuth token to ClickHouse Cloud as a JWT so queries are attributed to the real user instead of a shared service account. Requires a verified TLS connection. Alert and other backend queries have no signed-in user and are blocked by default; enable **Allow service account fallback** to let them run with the configured username/password (#1987)
+- Add a **Columns** setting to the logs data source config for browsing and filtering by any table column. The chosen columns appear in the Fields list, the log row details (under a **Fields** group next to the attribute groups), and the field filters, without editing each query. In single source mode the Columns field and the time, level, and message roles are pickers of the table's real columns, with a one-click **Add all columns**; other modes accept a typed list. Works for OTel and non-OTel log tables (#2108)
```
adamyeats: nit: release-please will generate its own entry from the `feat(logs):` commit, so this hand-written bullet becomes a second entry for one change. `#1987` and `#2092` already sit twice in `main`'s changelog for exactly this reason.

adamyeats (changes requested):
@alex-fedotyev consider adding an E2E spec before this merges, since the unit tests cannot catch a regression that only appears once Grafana renders the frame. `configEditor.spec.ts` and `columnAutodetect.spec.ts` already have the scaffolding, and the existing fixtures give you both an OTel and a non-OTel logs table.

alex-fedotyev:
Good call. Added an E2E spec in bc5a7ee.

It drives the Columns setting end to end on both a non-OTel table (`e2e_test.events`) and an OTel table (`e2e_test.otel_logs_v151`): a single-source logs datasource with `logs.additionalColumns` set, run the compact logs query, expand the first log row, and assert the configured column renders as a folded field. It then clicks "filter for value" on that field and checks the query regenerates a WHERE on the real column, which is the part the unit tests cannot reach since the fold only runs once Grafana renders the frame.

It reuses the existing scaffolding you pointed at: the datasource is created through the Grafana API in `beforeAll`/`afterAll` (no provisioning changes), and the log-row opener is the version-agnostic helper from `toggleableFilters.spec.ts`, so it holds across the legacy and new logs-panel DOMs in the CI matrix. I ran it locally against Grafana 12.2 before pushing.

alex-fedotyev:
Thanks for the thorough review, this was genuinely useful. I went through every thread. Pushed as `f07f533` (five commits on top of the E2E spec). Here is where each one landed.

**Correctness (`037306f`)**
- Fold deleting context columns and breaking Show-context: the fold now keys on the configured **Columns** setting only, and never folds a context column, so `getLogContextColumnsFromLogRow` still matches by name.
- Real `labels` column clobbered: the fold leaves an existing `labels` field alone unless it is the backend-serialized attribute map, so a user's own `labels` column is preserved.
- `Map`/`Array` rendering as `[object Object]`: configured columns now carry their schema `type` (looked up in `appendAdditionalLogColumns`), so collection columns are skipped; date/time columns are skipped too, so no more raw epoch.
- Fold swallowing hand-picked builder columns and a Table panel losing its columns: the same scoping to the Columns setting fixes it. This is the "key on the new setting" you suggested. It does mean a column hand-picked in the query builder no longer auto-becomes a browsable field, only the datasource setting does, so flagging that as an intentional narrowing.
- Empty-query silence: the `filterQuery` guard is narrowed to the transient logs-builder case, so an empty or invalid SQL-editor query surfaces its error again instead of being dropped.

**Config editor (`15d558e`)**
- Role field swapping control type mid-edit: the control type is now chosen from whether a schema fetch will run, not whether it has resolved, so the field never changes shape under the cursor.
- Untrimmed paste: the schema-backed Columns path trims like the free-text fallback.
- No validation: a configured column that is not in the fetched schema now shows an inline warning.
- The two schema-fetch tests: reworked to advance past the 400ms debounce before asserting, so they exercise the fetch path (a successful fetch would fail them now) rather than the identical first render.

**Test coverage (`037306f`)**
- The `status` fixture was a string with the coercion untested: it now uses a real number, and there is coverage for the `String()` path plus the context-column, date-column, and real-`labels` cases.

**Wording, telemetry, changelog (`1be15e4`)**
- "every logs query" and "every column": now "new logs queries" and "the table's scalar columns" in the docs and the tooltip.
- reportInteraction shipping column names: it now reports counts, so one Add-all click does not send the whole schema.
- Duplicate changelog entry: dropped the hand-written bullet, since release-please generates it from the commit.

**Label alignment (`f07f533`)**
- ColumnsEditor's width-8 label in the config: it now takes an optional label, tooltip, and width, so the config Columns row lines up with the width-12 rows and keeps its own tooltip. The query builders omit the props and are unchanged.

**Message heuristic (moving to its own PR)**
- The `query`/`request` message detection and the pattern-versus-column ordering: reverted here in `42e3dfa` and moving to a dedicated PR, since it changes role detection for every logs table and reads better as a focused change. I will open it once this merges, since it touches the same file.

**Deferred to a small follow-up, happy to fold in if you would rather:**
- Schema fetch not retried after Save & test on a bad connection: the clean fix keys the effect on a save signal that lives in the parent config editors, so I would rather not widen this PR into them.
- Scoping the chip-wrap to the logs and config fields: it is builder-wide today, and scoping it changes the table and traces builders too, so it reads better on its own.
- Surfacing "no columns detected" instead of a silent empty render: the SQL-editor half is covered by the guard narrowing above; a builder-side notice is a UX addition I would do separately.

**OTel unlock**
- Dropping `disabled={otelEnabled}` on the Columns editor is intentional, so columns can be added with OTel on. I have called it out in the PR description.

The E2E covers the render path across the Grafana matrix. Tell me if the narrowing or any of the three deferrals should go differently.

adamyeats:
Thanks, that is a clear write-up, and the e2e spec is exactly what I had in mind. I have resolved 11 threads, including the OTel narrowing and the tooltip wording.

The three deferrals all read fine to me as separate PRs. No need to widen this one for them.

Four are still open, because the fix does not reach every path:

- `appendAdditionalLogColumns` resolves the type now, but `logsQueryBuilderHooks.ts:54` still passes `[]`. The type stays undefined on the classic builder's new-query path, so both new skips go inert. A `Map` column still folds to `[object Object]`, and a `DateTime` still folds to a raw epoch. `LogsQueryBuilder.tsx:57` already holds `allColumns`, so threading it into the hook closes both threads.
- The `labels` guard misses one shape. A real `labels` String column whose first non-empty value parses as a JSON object is still read as the backend attribute map and overwritten.
- `utils.test.ts`: the `UInt16` fixture is fixed, but the per-row fold loop is still uncovered.
- `LogsConfig.test.tsx`: the two debounce tests are fixed, but the `uid`-absent test at line 403 still has the same timing hole.

The narrowing reads right to me, and calling it out in the description was the right call.

alex-fedotyev:
Thanks Adam, all four are closed in 5655167.

- Map / DateTime on the classic builder path: `useLogDefaultsOnMount` now takes `allColumns` and passes it to `appendAdditionalLogColumns`, so a configured column gets its type on the new-query path. I also added a backstop in the fold that skips a source field by its own frame type (time or object), so the Map to `[object Object]` and DateTime to epoch cases are covered even before the schema resolves.
- labels guard: the fold now skips when the query selects a real column named `labels` (the backend attribute map is synthesized and never selected), so a real `labels` column whose value parses as a JSON object is left intact.
- `utils.test.ts`: added a multi-row fold test that exercises the per-row loop, plus tests for the frame-type and selected-labels guards.
- `LogsConfig.test.tsx`: the uid-absent test now advances past the debounce with fake timers, like the other two, so it exercises the settle window instead of the identical first render.

adamyeats:
@alex-fedotyev all 21 threads are resolved and `56551670` checks out, thanks. The branch is 5 commits behind `main` (clean merge, no conflicts). Are you able to update it from your side? The ruleset needs the update to come from someone other than the approver, and I will approve and merge as soon as CI is green on the new head.

adamyeats:
@alex-fedotyev heads up, `main` moved under this since my last comment: `37bc1280` (SQL editor log volume) landed and the PR now conflicts in one hunk, the `./utils` import block at the top of `src/data/CHDatasource.ts`. Everything else auto-merges. I merged `main` locally with this union and the tree is green (typecheck, 1436 tests, lint):

```ts
import { planSqlLogsVolume } from './logsVolumeSql';
import {
  foldDiscoveredLogFieldsIntoLabels,
  labelsFieldName,
  mapGrafanaFormatToQueryType,
  transformQueryResponseWithTraceAndLogLinks,
} from './utils';
```

Once that is pushed I will approve and merge as soon as CI is green.

alex-fedotyev:
Done, pushed in b33ff46. I merged `main` in using the union you posted for the `./utils` import block in `CHDatasource.ts`; everything else auto-merged. Locally the tree is green: typecheck clean, 1441 tests pass, lint clean, build ok. CI is running on the new head now.

alex-fedotyev:
CI is green on `b33ff46`: full Playwright E2E matrix, Test-and-build, Test docs, and the ClickHouse-version matrix all pass. Ready for the approve and merge whenever you have a moment. Thanks, @adamyeats !

adamyeats (approved):

PR #2111, opened 2026-08-11
fix: trace logs linking with timeframe

### LINKED ISSUE #2106: Trace→logs data link drops the time-range filter, so the generated query ignores the Explore time range
**What happened**:

Clicking the `View logs` data link on a trace opens Explore with a query that has **no time predicate at all**, so the Explore time picker has no effect on it. The query is otherwise correct — it returns the right log rows — but it scans the full retention window instead of the selected range.

This is distinct from the other `View logs` reports:

- #1322 — link returns nothing / request times out. Ours returns correct rows.
- #1895 — link returns zero rows. Ours returns rows.
- #2103 — time filter present but emitted against a non-indexable column. Ours has no time filter to mis-target.

The generated logs query should carry the time-range filter that every other logs query carries, so the time picker constrains it.

1. Configure the data source with OTel logs and traces tables (`otelEnabled: true`, `otelVersion: latest`).
2. Run any query returning a `TraceId`, expand a row, click `View logs`.
3. Inspect the generated query.

The resulting Explore state (verbatim, from the URL Grafana produced):

```
refId          : Trace Logs
editorType     : builder
pluginVersion  : 4.20.0

rawSql:
  SELECT Timestamp as "timestamp", Body as "body", SeverityText as "level", TraceId as "traceID",
         ResourceAttributes, ScopeAttributes, LogAttributes
  FROM "otel"."otel_logs"
  WHERE ( traceID = '<trace-id>' )
  ORDER BY timestamp ASC

builderOptions.filters (count = 1):
  {"type":"string","operator":"=","filterType":"custom","key":"TraceId",
   "hint":"trace_id","condition":"AND","value":"<trace-id>"}

meta: {"minimized":true,"otelEnabled":true,"otelVersion":"latest"}
```

No `Timestamp`, no `TimestampTime`, no `$__fromTime`/`$__toTime`, and no filter carrying a time hint — so regenerating `rawSql` from `builderOptions` cannot produce a bound either.

**Root cause**:

`transformQueryResponseWithTraceAndLogLinks` (`src/data/utils.ts`) builds the `Trace Logs` query with a hand-written single-element `filters` array in both branches:

- when the originating query is a logs query, it spreads `originalQuery.builderOptions` but **overwrites** `filters` with just the `TraceId` equality, discarding the time-range filter that query already had
- otherwise it constructs options from log defaults but writes `filters` directly, never calling `getDefaultLogsFilters()`

`getDefaultLogsFilters()` (`src/components/queryBuilder/defaultQueryOptions.ts`) returns exactly the needed entry:

```ts
{ type: 'datetime', operator: FilterOperator.WithInGrafanaTimeRange,
  filterType: 'custom', key: '', hint: ColumnHint.FilterTime, condition: 'AND' }
```

and `getFilters()` in `src/data/sqlGenerator.ts` renders it as `>= $__fromTime AND <= $__toTime`, falling back from `FilterTime` to `Time` when `FilterTime` is unmapped (as it intentionally is for schema 1.3.0). So the mechanism works everywhere else; this one construction site skips it.

Neither data link sets a `range` on its `internal` object either, so there is no bound from the Grafana side to compensate.

**Impact**:

On a production `otel_logs` table — 2.66B rows, 786 active parts, 90-day retention, `PARTITION BY toDate(Timestamp)` — measured from a cold cache:

| Query | Result |
| --- | --- |
| as generated (no time predicate) | **> 5s** |
| `+ Timestamp >= now() - INTERVAL 7 DAY` | **< 1s** |
| `+ Timestamp >= now() - INTERVAL 1 HOUR` | **< 1s** |

`EXPLAIN ESTIMATE` with a bound present prunes to **1 part / 1 mark**, so the predicate is fully effective the moment it exists. Cost scales with parts consulted — 102 parts under a second, 276 over two, 786 over four — which is why an unbounded lookup degrades with retention rather than with result size.

**Suggested fix**:

Either preserve the originating query's time-range filter instead of overwriting `filters`, or seed the new options from `getDefaultLogsFilters()` and append the `TraceId` equality. Both restore time-picker semantics and make the query prune.

A larger alternative would be to extend the `<table>_trace_id_ts` two-step lookup from `generateTraceIdQuery` to `generateLogsQuery` — the machinery is already gated purely on the companion table existing — but that is not needed to fix this.

**Environment**:

- Grafana version: Grafana Cloud
- Plugin version: 4.20.0
- ClickHouse version: 26.2.1.558 (ClickHouse Cloud)
- Schema: `opentelemetry-collector-contrib` clickhouseexporter v0.157.0, `create_schema: true`

karl-power:
## Type of Change

- [ ] 🚀 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧹 Refactor / Chore

---

### What is the bug?

The `View logs` link built in `transformQueryResponseWithTraceAndLogLinks` (`src/data/utils.ts`) had three defects:

1. **No time bound (#2106).** Both construction sites emitted only the TraceId filter — the logs-origin branch overwrote the origin query's time filter, the defaults branch never added one — so the Explore time picker had no effect and the query scanned the full retention window.
2. **First auto-run failed on Grafana ≤12.** Explore mode shipped `rawSql: ''`, but Explore runs the split pane's query before the editor regenerates SQL, producing ClickHouse `code: 62, Empty query`. (Grafana ≥13 regenerates first, masking this.)
3. **No logs volume histogram on first open.** `getSupplementaryLogsVolumeQuery` requires `mode: BuilderMode.List`, which the link query never set.

### Fix

- Both branches include the time-range filter: the logs-origin branch preserves the origin's own time filter(s) (default fallback) and drops all other filters, including stale TraceId filters; the defaults branch seeds `[...getDefaultLogsFilters(), traceIdFilter]`. The existing `FilterTime → Time` fallback covers both OTel log schemas; unresolvable time filters are still dropped silently.
- `rawSql` is always pre-generated, with two substitutions on the serialized SQL only (`builderOptions` keeps the macro/raw forms, so the editor regenerates the live time-picker binding once the pane opens):
  - `bindTimeRangeMacros()` replaces `$__fromTime`/`$__toTime` with concrete `toDateTime(<unix>)` bounds from the request range — Grafana hides links whose serialized query contains unresolvable `$`-tokens, and these macros are backend-only.
  - `quoteTraceIdValueToken()` wraps `${__value.raw}` in quotes, since `escapeValue()` passes `$`-values through unquoted and Grafana ≤12 executes the substituted SQL verbatim (`code: 47`). Mirrors `generateTraceQuery`'s explicit quoting.
- Both branches set `mode: BuilderMode.List` so the logs volume/sample supplementary queries work on first open.

### How to reproduce

See #2106

### Related Issues

Closes #2106

## Please check that:


SpencerTorres (approved):

katebrenner (approved):

katebrenner:
repro'd locally, LGTM!

PR #2122, opened 2026-08-19
test(e2e): add shared E2E helpers

adamyeats:
## Type of Change

- [x] 🧹 Refactor / Chore

---

### What changed?

Adds `tests/e2e/helpers/` with the five pieces every spec currently keeps its own copy of:

| Helper | Replaces |
| --- | --- |
| `explore.ts` | the Explore URL builder, which existed in 15 slightly different versions |
| `sqlEditor.ts` | the SQL editor entry, in 12 versions |
| `queryResponse.ts` | the query-response body wrapper, in 10 versions, now typed |
| `env.ts` | the env and datasource preamble, in 15 versions |
| `builder.ts` | the shared builder interactions |

Also deletes `tests/fixtures/property-prices.sql`, which nothing in the repo references, and guards the bare `logs` gitignore pattern so it cannot match the `logs` spec directory the follow-up introduces.

### Why?

The copies had drifted apart, and one had lost its cloud fallback entirely. The typed response accessors also remove the need for the `no-explicit-any` casts the inline versions carried.

**Nothing imports these yet.** The follow-up regroups the specs into thematic directories and adopts them, which keeps that diff to the specs alone rather than mixing new shared code with 17 moved files.

---

Second of four PRs splitting #2044, which is 56 files and has not attracted a review in a month. #2121 was the config-only cron flakiness fix. This one adds the helpers, the next adopts them, and the last adds the new coverage.

Test-only, no user-facing change. `tsc --noEmit`, prettier and cspell are clean, and `playwright test --list` still collects 84 tests in 18 files, unchanged.


tests/e2e/helpers/env.ts
```diff
@@ -0,0 +1,41 @@
+import { test } from '@grafana/plugin-e2e';
+
+// Shared environment constants for the E2E suite: plugin identifiers,
+// datasource uid values for the local and Cloud runs, and the fixture time window.
+
+export const PLUGIN_TYPE = 'grafana-clickhouse-datasource';
```
aangelisc: nit: This can be imported from `plugin.json`

tests/e2e/helpers/env.ts
```diff
@@ -0,0 +1,41 @@
...
+// GRAFANA_URL is set only by the Cloud cron workflow (see .github/workflows/cron.yml).
+// Its presence indicates the local provisioning file and seed fixtures do not apply.
+export const isCloudRun = !!process.env.GRAFANA_URL;
+
+// CLOUD_DEFAULT_UID points at `[managed_data_source] - ClickHouse Native (PDC)` on the
+// shared Cloud dev instance. The infra team uses a stable `clickhouse-{protocol}-ds-m`
+// naming convention, but if the datasource is ever re-provisioned and Cloud E2E starts
+// failing with datasource-not-found errors, log into the instance, copy the current uid
+// from the /connections/datasources/edit/<uid> URL, and update this constant (or set
+// DS_E2E_UID in the workflow as a quick override).
```
aangelisc: nit: verbose comment

tests/e2e/helpers/explore.ts
```diff
@@ -0,0 +1,58 @@
...
+ * attaches View trace / View logs links to builder queries unless
+ * datasource-level trace defaults are configured, and the local e2e
+ * datasource has none). rawSql is passed alongside so the query runs
+ * deterministically on load without depending on editor re-generation.
+ * queryType is restored via the query's top-level queryType field (used by
+ * SQL mode). Tests that need the Query Builder UI without pre-built
+ * builderOptions should call switchToBuilderMode after
+ * page.goto(exploreUrl(...)).
+ */
+export function exploreUrl(opts: ExploreUrlOptions = {}): string {
```
aangelisc, line 34: Worth extracting to `plugin-e2e`? I could see this being useful elsewhere. Not blocking on this just worth considering.

tests/e2e/helpers/queryResponse.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Wraps explorePage.waitForQueryDataResponse, reading the response body
+ * inside the predicate while the CDP buffer is still live. Only resolves for
+ * an OK response whose body carries a frames array for the given refId.
+ *
+ * TODO: patch @grafana/plugin-e2e so waitForQueryDataResponse exposes the
+ * body directly, removing the need for this workaround.
+ */
+export async function waitForQueryDataResponseWithBody(explorePage: ExplorePage, refId = 'A') {
```
aangelisc, line 37: Also worth extracting most of the functions here.

aangelisc (dismissed):
Generally LGTM, my previous comments about extracting shared functions still stand. I'd also ask if we could go easier on the LLM generated comments as they're overly verbose 😊

tests/e2e/helpers/env.ts
adamyeats: `PLUGIN_TYPE` now comes from `src/plugin.json` (05dbaf08).

tests/e2e/helpers/env.ts
adamyeats: Trimmed in 05dbaf08. The other helpers get the same cut in 45b41b9b: comments now carry only the constraint or workaround the code does not show.

adamyeats:
Helper comments trimmed to the why in 45b41b9b. #2123 and #2124 get the same pass once they are rebased onto this. The extraction asks are tracked in #2183.

tests/e2e/helpers/explore.ts
adamyeats: The pane-state shape is Grafana-generic, so the builder belongs upstream. #2183 tracks proposing it to plugin-e2e. The TODO here is removed in 45b41b9b so it does not rot.

tests/e2e/helpers/queryResponse.ts
adamyeats: The upstream half is done: grafana/plugin-tools#2854 added `waitForQueryDataResponseWithBody` to `GrafanaPage`, shipped in plugin-e2e 3.12.0. The local bump and delegate are tracked in #2183, so the dependency change can land on its own. The typed accessors stay local because plugin-e2e does not type the `/api/ds/query` body.

aangelisc (approved):

PR #2123, opened 2026-08-19
test(e2e): group specs into thematic directories and adopt the shared helpers

adamyeats:
## Type of Change

- [x] 🧹 Refactor / Chore

---

### What changed?

Every spec moves out of the flat `tests/e2e/` root into a directory named for what it covers (`config/`, `editor/`, `builder/`, `filters/`, `traces/`, `logs/`, `types/`, `annotations/`), and adopts the helpers added in #2122 in place of its private copy. Playwright discovers `testDir` recursively, so the config needs no change.

Fixtures move to `tests/fixtures/`, since they are no longer specific to the E2E suite, and `docker-compose.yml` follows the path. Compact query mode splits out of `queryEditor.spec.ts` into its own spec.

**Net −950 lines.** Please review with rename detection on, since GitHub shows the heavier rewrites as delete-plus-add rather than moves.

### Why?

The private copies had drifted: the Explore URL builder existed in 15 slightly different versions and one spec had lost its cloud fallback entirely. Grouping also makes it possible to run a single theme.

---

## Test count: 84 → 83

Worth being explicit, since the number goes down:

- **7 retitled**, not removed. The redundant `(fixture)` prefix and the `should` verb are gone.
- **2 added.** The config editor spec gains a protocol health-check case and a default-port case. They arrive here rather than in the coverage PR only because that file moves in this one.
- **3 removed.** `SQL query returns rows from fixture data`, `SQL query with WHERE filter returns matching rows` and `Aggregate SQL query returns a count from fixture data` are generic smokes subsumed by the `builderQueries`, `macros` and `panelRendering` suites in the follow-up. **If this merges without that follow-up, those three cases are uncovered.** The stack is ordered so that does not happen.

---

Third of four PRs splitting #2044, which is 56 files and has not attracted a review in a month. Stacked on #2122, which adds the helpers this PR adopts.

Two deliberate differences from #2044:

- The moved comments use US spellings, so the cspell dictionary needs one new entry (`spinbutton`, an ARIA role) rather than twenty, six of which were UK spellings.
- The dictionary is not re-sorted, keeping that file's diff to a single line.

Test-only, no user-facing change. `tsc --noEmit`, `npm run spellcheck` and prettier are all clean.


fabrizio-grafana:
@adamyeats stack check—which admittedly I have never seen—is failing. Moving this back to In progress meanwhile.

adamyeats:
@fabrizio-grafana That banner is GitHub's merge gate for the stack, not a CI check, so it did not need to block a review. At the time it had three causes: #2122 had four unresolved review threads, #2122 was three commits behind main (this repo requires branches to be up to date before merging), and this PR had no review yet. The last one is circular: the stack reads as unmergeable partly because this PR is unreviewed. Moving back to PR Needs Review.

PR #2135, opened 2026-08-19
Accept human-friendly units in trace Duration filter

### LINKED ISSUE #2133: Log and traces timestamp filter is bare nanoseconds
**What happened**:

When searching for traces using the traces query builder I added a duration filter and was surprised to figure out through trial and error that it accepts only bare nanosecond time units.

I expected a duration or interval field to take human-friendly units or fractional seconds, or at least a clear field label identifying it as nanoseconds.

* Visit Explore view with a clickhouse data source pointing to a DB containing otel trace data
* Query Type -> Traces
* "+ Filter"
* `AND`, `Duration`, `>`, `1000000000`

You will now see traces with a duration longer than 1 second.

**Screenshots**

Image

It's hard to argue that this is a true behaviour bug, it's a UX bug/issue. But there's no other way to report issues due to https://github.com/grafana/clickhouse-datasource/issues/2132 and I wanted an issue in place to associate a pending PR with.

**Environment**:

- Grafana version: Any 12.x
- Plugin version: 4.20.0
- OS Grafana is installed on: N/A
- User OS & Browser: N/A
- Others: N/A

ringerc:
## Type of Change

- [x] 🚀 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧹 Refactor / Chore

---

> [!NOTE]
> LLM assistance was used to prepare the branch pointed to by this PR

## Feature

Accept human-friendly units in trace Duration filter

### What is this feature?

In the Trace Query Builder, filtering on the `Duration` column accepts only bare numeric input, which is interpreted as a time in nanoseconds. There is no UI hint about the unit, so users have to discover through trial and error that `> 1000000000` means "> 1 second".

Improve usability / UX.

### Who is this feature for?

Anyone who queries traces.

### How to test this feature

Before patch:

1. Open Explore against a ClickHouse datasource with OTel trace data.
2. Query Type -> **Traces**.
3. **+ Filter** -> `AND`, `Duration`, `>`, `1000000000`.
4. Note that the input gives no indication that the value is nanoseconds and rejects any other form.

After patch:

1. Open Explore against a ClickHouse datasource with OTel trace data.
2. Query Type -> **Traces**.
3. **+ Filter** -> `AND`, `Duration`, `>`, `1s`.
4. Query finds traces with duration greater than 1 second
5. UI shows available units; try other units
6. Run with bare 1000000000 value; note it falls back to nanoseconds for backwards compatibility.

For a quick test with some canned data:

1. check out this branch
2. `npm run build`
3. patch the datasource config to use otel traces schema by default:
  ```
  yq -i '
    (.datasources[] | select(.name == "ClickHouse")) |= (
      .allowUiUpdates = true |
      .jsonData.traces.otelEnabled = true |
      .jsonData.traces.otelVersion = "latest"
    )
  ' provisioning/datasources/clickhouse.yml
  ```
4. `npm run server` to stand up the test harness
5. save `seed_otel_traces.sh` to the cwd from [seed_otel_traces.sh](https://github.com/user-attachments/files/31244082/seed_otel_traces.sh)
6. Run `CH_DB='default' bash seed_otel_traces.sh` to populate the table with some traces
7. Visit `http://localhost:3000`, log in, go to "Explore"
8. crtl-shift-R / cmd-shift-R to flush any cached client javascript
9. Select query builder, query type -> traces
10. open the "columns" view, double check that "use otel" is toggled on, and if it isn't, toggle it on
11. "run query" with default settings; you will see 2 traces with durations 100ms and 1.2s
12. "+Filter", `AND`, `Duration`, `>`, `1s` - note the unit, write literally `1s` in the field
13. "run query" again; you will now see only one trace

### Related Issues

Closes #2133

## Screenshots / Videos

image

## Please check that:

- **Only targets `Duration`.** The duration input replaces the plain number input only when the filter targets the configured trace Duration column (matched via `ColumnHint.TraceDurationTime` or the resolved column key). Non-duration numeric filters and other query builders are untouched.
- **No SQL generator changes.** Parsed input is converted to the column's configured `traceDurationUnit` before being written to `NumberFilter.value`, so the existing `WHERE` generation is correct for ns-, µs-, ms-, and s-stored columns.
- **Backwards compatible.**
  - A bare number is still interpreted as nanoseconds, matching today's behaviour.
  - `NumberFilter.rawInput?: string` is additive; saved queries without it fall back to formatting `value` with the stored-unit suffix.
- **UX improvements.** `Input.suffix` shows the current stored unit and an info icon; the tooltip lists accepted forms (`ns`, `us`/`µs`, `ms`, `s`) and notes that bare numbers are treated as nanoseconds.
- **Test cover added.** `tsc --noEmit` is ok; full Jest suite runs ok; ESLint clean on touched files. New unit tests cover the parser/converters; new `FilterEditor` tests cover the render switch, the fallback, and unit conversion for a milliseconds-stored column.


iwysiu (commented):
Hi @ringerc ! Sorry for the delay. It mostly looks good!
- It's failing CI on the cspell step (you can run it with `npm run spellcheck`). If you need to you can add new ignored words in `cspell.config.json`, but I think its mostly regional spellings
- nit, but it may be worth adding a test for the default hinted Duration filter (key: '' + TraceDurationTime) from [here](https://github.com/grafana/clickhouse-datasource/blob/revert-dsabstraction-schema/src/components/queryBuilder/views/traceQueryBuilderHooks.ts#L155)

ringerc:
Hi @iwysiu . I've addressed those comments, thanks for taking a look.

iwysiu (approved):

iwysiu (approved):
Thanks for the work! I'll get this merged.

ringerc:
@iwysiu Looks like CLA-bot needs you to approve it too

Issue #2126, opened 2026-08-19
TimeSeries query builder produces broken queries for clickhouse otel schema metrics tables


ringerc:
**What happened**:

I used the TimeSeries query builder in the clickhouse data source to query an `otel_metrics` table for time-series. I included `ResourceAttributes` and `Attributes` columns in the query's selected columns. The result was a confusing mashed-up timeseries that combined data points from numerous overlapping series into a single displayed series, completely ignoring the `ResourceAttributes` and `Attributes` for the purpose of determining time-series identity.

For this test I am using a clickhouse db populated with an otel-collector running the clickhouse otel exporter to deliver metrics time series data into a table with the clickhouse-documented otel metrics schema.

**What you expected to happen**:

The map-valued `ResourceAttributes` or `Attributes` should be exploded into distinct pairs that uniquely identify each time-series, so the query returns numerous series each of which has only its own data-points.

*or* at least

A notice/warning indicating that some selected columns have unsupported types, and were dropped, so hand-built SQL should be used instead if these need to be used.

**How to reproduce it (as minimally and precisely as possible)**:

Using this docker-compose based repro: [metrics-repro.zip](https://github.com/user-attachments/files/31204095/metrics-repro.zip) which was tested on Ubuntu 26.04 with Docker CE 24.0.5:

* `unzip metrics-repro.zip`
* `cd metrics-repro`
* `docker compose up`
* wait for readiness
* visit `http://localhost:3000/dashboards`
* open the repro dashboard (only one there)
* note the query that uses Attributes and ResourceAttributes directly silently squashes all series into a single misleading series; compare to the workaround panel that stringifies them
[...]

PR #2136, opened 2026-08-20
Warn when time series queries include Map/Array/JSON columns

ringerc:
## Type of Change

- [x] 🐛 Bug Fix

### What is the bug?

Queries from the Time Series query builder silently drop `otel_metrics` query `Attributes` and `ResourceAttributes` so query results incorrectly merge multiple time-series into a single series with the wrong datapoints. See https://github.com/grafana/clickhouse-datasource/issues/2126 .

The cause is that the Time Series query builder silently drops `Map`/`Array`/`Tuple`/`Variant`/`Dynamic`/`JSON` columns from series identity: they arrive as `data.FieldTypeJSON`, and sqlds' `FrameToTimeSeries` only promotes `String` fields to labels. All rows collapse into one misleading series with no warning which is very confusing for the user.

This PR adds a warning notice so the user is informed of the issue. It does not change query building or execution. Future changes could extend `mergeOpenTelemetryLabels` which per its name is currently logs-only to actually promote these maps to output labels.

It provides a hint to tell the user to write `Attributes['key_name']` explicitly in the column-list instead; this will work, albeit in a clunky and less-than-future-proof way. 

### How to reproduce

Before patch:

1. Open Explore against a ClickHouse datasource populated with OTel metrics (e.g. via the otel-collector ClickHouse exporter, stock `otel_metrics` schema).
2. Query Type -> **Time Series**.
3. Select columns `MetricName`, `ResourceAttributes`, `Attributes`, `Value`; optionally filter `MetricName = <known metric>`.
4. Run. Observe a single mashed-up series and no indication that the map columns are being ignored.

After patch:

1. Same steps.
2. The graph is unchanged (this PR does not fix series identity), **but** the panel now shows a warning notice naming the ignored columns and suggesting `col['key']` extraction or the SQL editor.

For a quick test with canned data:

1. Check out this branch
2. `npm run build && mage build:linux` (adjusting per your host os)
3. `npm run server` to stand up the plugin's test harness (Grafana + ClickHouse).
4. Save `seed_otel_metrics.sh` [seed_otel_metrics.sh](https://github.com/user-attachments/files/31246239/seed_otel_metrics.sh) to the cwd.
5. Run `CH_DB=default bash seed_otel_metrics.sh` to create `default.otel_metrics` in the harness's ClickHouse container and populate it with 1000 synthetic points across three distinct `Attributes` sets.
6. Visit `http://localhost:3000`, log in, go to **Explore**.
7. Ctrl-shift-R / cmd-shift-R to flush any cached client JavaScript.
8. Query builder, query type -> **time series**, database `default`, table `otel_metrics`.
9. Add columns `MetricName`, `ResourceAttributes`, `Attributes`, `Value`; filter `MetricName = 'http.server.request.duration'`.
10. Run query. Same misleading single-series graph as before, **plus** a new warning notice listing `ResourceAttributes` and `Attributes` as ignored columns.
11. Switch the query type to **table** and re-run; no notice appears (only time series frames get it).
12. Remove `Attributes` and `ResourceAttributes` from the column list. Instead add custom columns named `Attributes['http.route']`, `Attributes['http.response.status_code']` and `Attributes['missing']`. Re-run the query. Note that the output is now correct.

### Related Issues

Refs #2126.

## Screenshots / Videos

image

Note the added warning. It's not ideal, as it's not particularly prominent, but it's something and would've saved me a lot of time/confusion.

It points the user at the workaround, which is to explicitly qualify the column-list with the desired attributes and use `as` to give the columns readable labels. It's not particularly convenient as there's no real UI discoverability for this and no autocomplete, but it works, as shown below:

image

where the column-list entries `Attributes['http.route'] as http_route` and `Attributes['http.response.status_code'] AS http_response_status_code` are added, resulting in:

Output series look like:

```
{MetricName="http.server.request.duration", http_response_status_code="200", http_route="/a"}
```

which is kind of ugly, but at least correct, though it can't handle varying attribute-sets. The result SQL is

```
SELECT TimeUnix as "time", MetricName, Value, Attributes['http.route'] as http_route, Attributes['http.response.status_code'] AS http_response_status_code FROM "default"."otel_metrics" WHERE ( time >= $__fromTime AND time <= $__toTime ) ORDER BY time ASC LIMIT 10000
```

Omitting the `as` alias works too, it just results in ugly series names.

## Please check that:

- [x] Tests added (`pkg/plugin/driver_test.go`: positive, negative, vis-type gating).
- [ ] Docs — N/A, notice text is self-explanatory.

Intentionally minimal: only attaches a `NoticeSeverityWarning` on `VisTypeGraph` frames that contain `FieldTypeJSON` / `FieldTypeNullableJSON` fields. Verified with `go build`, `go vet`, `go test ./pkg/plugin/`.

PR branch prepared with LLM assistance.


pkg/plugin/driver.go
```diff
@@ -622,6 +626,44 @@ func (h *Clickhouse) MutateResponse(ctx context.Context, res data.Frames) (data.
...
+	}
+	if len(unsupported) == 0 {
+		return
+	}
+
+	text := fmt.Sprintf(
+		"The following selected columns have a Map/Array/Tuple/JSON type and cannot be used to distinguish time series: %s. "+
+			"All rows will collapse into a single series regardless of these columns' values. "+
+			"Extract specific keys instead by typing an expression such as %s['some_key'] as some_key directly into the Columns selector (the 'as some_key' suffix names the resulting series/label), or switch to the SQL editor to hand-write the query.",
+		strings.Join(unsupported, ", "),
```
iwysiu, line 656: I think we want to dedup this list before we return it. We support a couple different formats, and for wide this will lead to multiple of the same column.

It gets kind of in the weeds of the the implementation, but basically [wide timeseries](https://grafana.com/developers/dataplane/timeseries#time-series-wide-format-timeserieswide) have a different column for every value for a label field. The clickhouse datasource fetches data through our `grafana/sqlds` library, which converts frames with string or bool columns to wide frames automatically.

pkg/plugin/driver.go
```diff
@@ -622,6 +626,44 @@ func (h *Clickhouse) MutateResponse(ctx context.Context, res data.Frames) (data.
...
+		}
+	}
+	if len(unsupported) == 0 {
+		return
+	}
+
+	text := fmt.Sprintf(
+		"The following selected columns have a Map/Array/Tuple/JSON type and cannot be used to distinguish time series: %s. "+
+			"All rows will collapse into a single series regardless of these columns' values. "+
+			"Extract specific keys instead by typing an expression such as %s['some_key'] as some_key directly into the Columns selector (the 'as some_key' suffix names the resulting series/label), or switch to the SQL editor to hand-write the query.",
```
iwysiu, line 655: nit: this is just phrasing but technically the `%s['some_key']` isn't the correct syntax for arrays or tuples

pkg/plugin/driver.go
```diff
@@ -622,6 +626,44 @@ func (h *Clickhouse) MutateResponse(ctx context.Context, res data.Frames) (data.
 	return res, nil
 }
 
+// warnUnsupportedTimeSeriesFields attaches a warning notice to a time-series
+// frame when it contains columns that cannot participate in series identity.
+//
+// ClickHouse Map/Array/Tuple/Variant/Dynamic/JSON columns are surfaced by the
+// converter registry as data.FieldTypeJSON. sqlds' FrameToTimeSeries only
```
iwysiu, line 633: nit: This is actually beacuse of sqlds running `LongToWide` not `FrameToTimeSeries`

pkg/plugin/driver.go
```diff
@@ -622,6 +626,44 @@ func (h *Clickhouse) MutateResponse(ctx context.Context, res data.Frames) (data.
...
+// JSON-shaped columns are silently ignored for series identity: all rows collapse
+// into a single series even though the user selected columns intended to
+// distinguish them. This is particularly confusing for OTel metrics-schema
+// tables where ResourceAttributes/Attributes are Map(...) columns.
+//
+// See https://github.com/grafana/clickhouse-datasource/issues/2126.
+func warnUnsupportedTimeSeriesFields(frame *data.Frame) {
+	var unsupported []string
+	for _, field := range frame.Fields {
+		if field.Type() == data.FieldTypeJSON || field.Type() == data.FieldTypeNullableJSON {
```
iwysiu, line 644: You have a golint failure that would be fixed by swapping to 
```suggestion
		if field.Type().JSON() {
```

iwysiu (commented):
Hi @ringerc , thanks for your work and your patience! This mostly looks good, a couple comments.

pkg/plugin/driver.go
ringerc: I pushed a change to accumulate to a map then use the sorted key-set. I can switch to iterating over it and keeping a 'seen' map if desired, but the map approach is simpler.

pkg/plugin/driver.go
ringerc: Amended to explicitly cite map columns. I'm not inclined to embed a whole language tutorial with indexing guidance for each different type, hopefully it's enough.

pkg/plugin/driver.go
ringerc: Thanks. That's my mistake. Clarified in an update.

pkg/plugin/driver.go
ringerc: Fixed, thanks

ringerc:
Thanks @iwysiu . Hopefully these corrections are sufficient. Thanks for taking the time to have a look. This issue threw me pretty badly when I was trying to use the timeseries query, and a hint would've helped me a lot.

pkg/plugin/driver_test.go
```diff
@@ -97,6 +97,91 @@ func TestMergeOpenTelemetryLabels(t *testing.T) {
...
+		require.Len(t, frame.Meta.Notices, 1)
+		notice := frame.Meta.Notices[0]
+		assert.Equal(t, data.NoticeSeverityWarning, notice.Severity)
+		assert.Contains(t, notice.Text, "ResourceAttributes")
+		assert.Contains(t, notice.Text, "Attributes")
+		// Column names are reported in sorted order; "Attributes" sorts before "ResourceAttributes".
+		assert.Contains(t, notice.Text, "Attributes['some_key'] as some_key")
+	})
+
+	t.Run("dedups repeated JSON-shaped column names from wide-format frames", func(t *testing.T) {
```
iwysiu, line 125: The linter is catching on "dedups" as not a word
```suggestion
	t.Run("deduplicates repeated JSON-shaped column names from wide-format frames", func(t *testing.T) {
```

iwysiu (commented):
@ringerc the changes look good! Just one more linting issue and I can get this merged in

PR #2141, opened 2026-08-26
feat(explore): show log volume for SQL editor queries

### LINKED ISSUE #2112: Support aggregated logs volume for SQL editor queries (follow-up to #1364)
## Summary

In Explore, a Logs query written in the SQL editor gets a logs volume histogram that only covers the rows the query returns, not the selected time range. With a typical `ORDER BY <time> DESC LIMIT 1000`, the histogram collapses to the span of the newest 1000 rows, which is a few seconds under steady ingest, and it does not change when you widen the time range. The same query built in the Query Builder shows a correct full-range histogram.

I would like the SQL editor to get the same full-range aggregated logs volume when the data source has enough configuration to build it safely, and to keep the current fallback otherwise.

## Current behavior

`getSupportedSupplementaryQueryTypes` returns an empty array when any target uses the SQL editor (`editorType !== EditorType.Builder`), so Grafana is told the data source has no logs volume query for that request and draws its built-in limited histogram from the returned log rows. Because those rows are capped by the query `LIMIT`, the histogram only ever covers the newest N rows.

This is intended today. It came from #1364 and shipped in #1475 (enabled by `grafana/grafana#115348`, which made supplementary-query support request-aware). #1475 chose the fallback on purpose so the SQL editor stops issuing a broken volume query. #1364 also called out the alternative that was not built: "This could be supported by making use of the default columns for the data source configuration. If they are not set, then the logs volume query can be skipped."

## Proposal

Build the aggregated volume query for SQL editor Logs queries when it can be done safely, using the configured defaults that already exist for this purpose:

- Reuse `getDefaultLogsTable()` and `getDefaultLogsColumns()` (already used by `getSupplementaryLogsSampleQuery`) to resolve the timestamp and log level columns without the builder options.
- Preserve the user `WHERE` so the histogram matches the log list. The predicate can be read from the raw SQL with the existing AST helpers in `src/data/ast.ts` (`sqlToStatement`), or the top-level clause boundaries can be located with the ClickHouse lexer in `src/ch-parser/lexer.ts` (#1778). Drop the user `ORDER BY` and `LIMIT` for the aggregation.
- Emit the same time-bucketed `count()` and per-level `sum(multiSearchAny(...))` query the builder path already generates in `getSupplementaryLogsVolumeQuery`, over the full range.

This applies to both the compact and classic editors. Both funnel through the same `EditorTypeSwitcher` and set `editorType: EditorType.SQL`, so the gate and the fix are identical; there is nothing compact-specific.

## Fallback so a wrong count never renders

Keep the current limited row-based histogram (return `[]` for that target) whenever a faithful aggregation cannot be built, including:

- No default logs table or columns configured, or the SQL `FROM` is not the configured logs table.
- The SQL is not a single plain `SELECT` (CTEs, `UNION`, `JOIN`, subquery in `FROM`).
- The query already aggregates (its own `GROUP BY` or aggregate functions).
- The `WHERE` references a SELECT-list alias that would be dropped (ClickHouse allows alias references in `WHERE`).
- The SQL fails to parse (the pgsql-based parser is not full ClickHouse dialect; `sqlToStatement` already returns empty on parse failure).

The rule: only replace the fallback when the aggregation is provably consistent with the log list, otherwise keep today's behavior.

## Alternatives considered

- Wrapping the user SQL as a subquery and counting over time buckets. This still has to strip the inner `ORDER BY` and `LIMIT` (otherwise it counts at most `LIMIT` rows) and expose a timestamp to bucket on, so it needs the same parsing and is more fragile than reusing the default-column config.
- Hiding the histogram entirely for the SQL editor (the other option raised in #1364). That avoids the misleading chart but drops a useful feature for the common OTel case where the default columns are known.

## Repro

1. Explore with a ClickHouse OTel logs source.
2. SQL editor, Logs query type: `SELECT Timestamp, SeverityText, Body FROM otel.otel_logs WHERE ( Timestamp >= $__fromTime AND Timestamp <= $__toTime ) ORDER BY Timestamp DESC LIMIT 1000`.
3. Set the range to Last 12 hours under steady ingest and run.
4. The logs volume histogram only spans the newest 1000 rows (a few seconds) and does not change when the range widens. The same query in the Query Builder shows the full 12 hours.

Tested on 4.20.0; the gate is unchanged on `main`. Happy to work on this if the approach sounds right.

SpencerTorres:
## Type of Change

- [x] 🚀 Feature

---

### What is this feature?

Aggregated log volume for Logs queries written in the SQL editor. The plugin wraps the user's SQL as a derived table, strips the trailing `ORDER BY`/`LIMIT`, and aggregates over the full time range, instead of letting Grafana draw its row-based histogram.

Closes #2112

In Explore, a SQL editor Logs query got a histogram capped by the query's `LIMIT` that didn't follow the selected time range.

### Who is this feature for?

Anyone querying logs via the SQL editor, including users who start in the Query Builder and click "Edit as SQL".

### How to test this feature

1. Configure a data source with an OTel logs source (default log table + columns set under **Logs**).
2. In Explore, query logs from the SQL editor, either by clicking "Edit as SQL" from the Query Builder, or by typing SQL directly. Confirm the volume histogram covers the whole time range rather than stopping at the `LIMIT`.
3. Change the time range (1h / 12h / 7d) and confirm the histogram follows it.
4. Add a `WHERE` clause and confirm the histogram reflects it.
5. Append `SETTINGS max_threads = 4` and confirm it falls back to the row-based histogram rather than showing a wrong count. The browser console explains why.

---

## Screenshots / Videos

SQL query with logs histogram view, unfiltered

> SQL query with logs histogram view, unfiltered

SQL query with logs histogram view, filters preserved

> SQL query with logs histogram view, filters preserved
---

## Please check that:

**Approach:** the user's SQL is wrapped, not rewritten, so their `FROM`, `WHERE`, joins, CTEs, macros and variables carry through verbatim. New `scanTopLevelSelect` (`src/ch-parser/spans.ts`) only locates the droppable tail and the projected columns; it never interprets the `FROM`. Anything not positively recognized declines and falls back.

**Known limitation:** the configured logs timestamp must be a `DateTime`-family column. `toStartOfInterval` rejects `Date`/`Date32` below day granularity and rejects `UInt32`/`String` at any granularity. Shared with the Query Builder path.

**Verification:** 1338 unit tests. The emitted SQL was also executed against ClickHouse 26.7 and against a running Grafana's `/api/ds/query`, with histogram totals diffed against independently computed ground truth across several time ranges and filter shapes at 250k rows.


src/ch-parser/spans.ts
```diff
@@ -0,0 +1,361 @@
...
+      }
+    }
+
+    if (tailIndex !== -1) {
+      continue;
+    }
+
+    // Only in clause position: requiring the expected argument stops `WHERE format = 1`,
+    // `WHERE limit = 1` and `... AS format` from truncating the statement.
+    if (token.type === TokenType.BareWord && TAIL_KEYWORDS.has(token.text.toUpperCase()) && next) {
```
adamyeats: The tail only latches when a Number follows LIMIT, so `LIMIT (500)` and `LIMIT ${maxRows}` survive into the head. I ran `... WHERE (Timestamp >= $__fromTime AND Timestamp <= $__toTime) LIMIT (500)` in Explore and the histogram reported 500 lines in a single hour instead of roughly 72,000 across six, with no error and no fallback. Consider declining when a LIMIT is present but its argument is not recognised.

src/ch-parser/spans.ts
```diff
@@ -0,0 +1,361 @@
...
+      const expectsNumber = !token.matchKeyword('FORMAT');
+      const argumentFits = expectsNumber ? next.type === TokenType.Number : next.type === TokenType.BareWord;
```
adamyeats: For FORMAT any bareword counts as the format name, and keywords lex as barewords, so a column called `format` truncates the statement mid-clause. `WHERE format LIKE 'json%'` emits `FROM (\nSELECT ... WHERE\n) AS src` and ClickHouse answers `Code: 62 Syntax error`, with no fallback because the plan reported success. It might be worth requiring nothing after `FORMAT <name>` except a semicolon.

src/ch-parser/spans.ts
```diff
@@ -0,0 +1,361 @@
...
+  token: Token;
+  depth: number;
+}
+
+/** Splits the SELECT list on depth-`baseDepth` commas so commas inside calls do not split items. */
+function parseSelectItems(tokens: DepthToken[], baseDepth: number): SelectItem[] {
+  const groups: DepthToken[][] = [];
+  let current: DepthToken[] = [];
+  for (const entry of tokens) {
+    if (entry.token.type === TokenType.Comma && entry.depth === baseDepth) {
```
adamyeats, line 192: `depth` only tracks round brackets, so an array literal in the select list splits into phantom items. `SELECT Timestamp, Body, [ServiceName, SeverityText, ScopeName] AS labels FROM demo.otel_logs` reports a `SeverityText` item the derived table never exposes, and `planSqlLogsVolume` still returns ok with `src."SeverityText"` in the aggregate.

src/ch-parser/spans.ts
```diff
@@ -0,0 +1,361 @@
...
+      // An alias is fine, but a second identifier after it is not.
+      continue;
+    }
+    if (
+      token.matchKeyword('WHERE') ||
+      token.matchKeyword('PREWHERE') ||
+      token.matchKeyword('GROUP') ||
+      token.matchKeyword('HAVING') ||
+      token.matchKeyword('ORDER') ||
+      token.matchKeyword('LIMIT') ||
```
adamyeats: nit: OFFSET and FORMAT are missing from the terminator list, so `SELECT * FROM otel_logs OFFSET 5` loses its FROM target and declines as `WildcardOverUnknownTable` even though the tail was already stripped.

src/ch-parser/spans.ts
```diff
@@ -0,0 +1,361 @@
...
+
+  // Only a bare reference, optionally with FINAL or an alias, counts. Anything else — a JOIN, a
+  // comma join, ARRAY JOIN, SAMPLE — means the row set is not simply that table's.
+  for (let j = i; j < topLevel.length; j++) {
+    const { token } = topLevel[j];
+    if (token.matchKeyword('FINAL')) {
+      continue;
+    }
+    if (token.matchKeyword('AS') || isIdentifier(token)) {
+      // An alias is fine, but a second identifier after it is not.
```
adamyeats: nit: the loop continues on every identifier, so `FROM otel_logs x y` is accepted and the restriction the comment describes is not implemented.

src/ch-parser/spans.ts
```diff
@@ -0,0 +1,361 @@
...
+  from?: FromTarget;
+}
+
+export type ScanResult = { ok: true; select: TopLevelSelect } | { ok: false; problem: SelectShapeProblem };
+
+/** Keywords that begin the trailing clauses we drop. */
+const TAIL_KEYWORDS = new Set(['LIMIT', 'OFFSET', 'FORMAT']);
+
+/** `"a""b"` -> `a"b`, `` `x` `` -> `x`. */
+export function unquoteIdentifier(text: string): string {
```
adamyeats: The lexer accepts backslash escapes inside quoted identifiers, but this only undoes the doubled-quote form, so `AS "a\\"b"` round-trips to an alias that `escapeIdentifier` re-quotes into invalid SQL.

src/data/logsVolumeSql.ts
```diff
@@ -0,0 +1,168 @@
...
+  // collides with one of ours (a level column aliased `time`, or `info`) cannot be shadowed by
+  // the outer SELECT's aliases.
+  const qualifiedTime = `${SOURCE_ALIAS}.${escapeIdentifier(timeName)}`;
+  const aggregates = levelName
+    ? buildLogLevelAggregateExpressions(`${SOURCE_ALIAS}.${escapeIdentifier(levelName)}`).map(
+        ({ alias, expression }) => `sum(${expression}) AS "${alias}"`
+      )
+    : [`count(*) AS "${DEFAULT_LOGS_ALIAS}"`];
+
+  const bucket = `toStartOfInterval(${qualifiedTime}, INTERVAL 1 ${getTimeFieldRoundingInterval(scopedVars)})`;
```
adamyeats: The plugin's own `$__timeInterval` macro wraps the column in `toDateTime()` for exactly this reason. With a `Date32` timestamp the panel shows "Failed to load log volume" and `Code: 43 Illegal interval kind for argument data type Date`, where main drew the row-based histogram. `ColumnHint.FilterTime` is documented as a lower-precision column, so this is reachable through the supported config.

src/data/logsVolumeSql.ts
```diff
@@ -0,0 +1,168 @@
...
+/**
+ * Whether the scanned FROM is the data source's configured logs table. Table and database names
+ * are case sensitive in ClickHouse, so this compares exactly, and an unqualified reference is
+ * accepted only when the configured table is itself unqualified or its database matches.
+ */
+function isConfiguredLogsTable(
+  from: FromTarget | undefined,
+  logsTable?: { database?: string; table?: string }
+): boolean {
+  if (!from || !logsTable?.table || from.table !== logsTable.table) {
```
adamyeats, line 49: `parseFromTarget` never reads the WITH clause, so `WITH otel_logs AS (SELECT 1 AS x) SELECT * FROM otel_logs` returns a FromTarget that passes this check, and the wildcard branch then references columns the CTE does not produce.

src/data/CHDatasource.ts
```diff
@@ -308,11 +311,59 @@ export class Datasource
...
+      }
+    });
+
+    // A target wants a histogram if it declares itself as logs, or if it is SQL that aggregates
+    // cleanly — the SQL editor defaults queryType to Table, so a hand-written logs query often
+    // does not declare itself. Anything that declares logs but cannot be built vetoes the whole
+    // request, since Grafana merges every volume frame into one chart.
+    const wantsVolume = probed.filter(
+      ({ target, built }) =>
+        this.isDeclaredLogsQuery(target) || (target.editorType === EditorType.SQL && built !== undefined)
```
adamyeats: Any SQL target that happens to aggregate is admitted regardless of its declared query type. With a second query explicitly set to Table, the histogram gained a `logs` series of 67.8K and roughly doubled. Consider treating an explicitly chosen Table or Time Series type as a veto, since `SqlEditor` writes queryType and format whenever the switcher is touched.

src/data/CHDatasource.ts
```diff
@@ -308,11 +311,59 @@ export class Datasource
...
+    );
+    if (wantsVolume.length > 0 && wantsVolume.every(({ built }) => built !== undefined)) {
+      types.push(SupplementaryQueryType.LogsVolume);
+    }
+
+    if (
+      visibleTargets.some((t) => this.isDeclaredLogsQuery(t)) &&
```
adamyeats: Explore only renders the logs sample when there is no logs result. The shipped 13.2.1 bundle computes it as `!!(E.dataProvider !== void 0 && !x && (u || h))` with `x` = logsResult, so requiring a declared Logs target offers the sample exactly when it cannot render, and withholds it from the builder Table queries the panel exists for.

src/data/CHDatasource.ts
```diff
@@ -431,11 +489,48 @@ export class Datasource
...
+    logsVolumeRequest: DataQueryRequest<CHQuery>,
+    query: CHQuery
+  ): CHQuery | undefined {
+    const plan = planSqlLogsVolume(query.rawSql, logsVolumeRequest.scopedVars, this.getDefaultLogsColumns(), {
+      database: this.getDefaultLogsDatabase(),
+      table: this.getDefaultLogsTable(),
+    });
+    if (!plan.ok) {
+      // Once per query text rather than per render, so it cannot spam.
+      if (this.lastDeclinedLogsVolumeSql !== query.rawSql) {
```
adamyeats: nit: one slot on the datasource instance is shared across every pane and panel, so two declining queries overwrite each other and warn on every probe. A Set keyed on `rawSql` would hold the "once per query text" promise.

src/data/CHDatasource.ts
```diff
@@ -308,11 +311,59 @@ export class Datasource
...
+
+    const types: SupplementaryQueryType[] = [];
+    // Hidden targets are not rendered, so they must not influence the decision or be counted.
+    const visibleTargets = dsRequest.targets.filter((t) => !t.hide);
+
+    const probed = visibleTargets.map((target) => {
+      try {
+        return { target, built: this.getSupplementaryLogsVolumeQuery(dsRequest, target) };
+      } catch (error) {
+        console.warn('[clickhouse] logs volume feasibility check failed, falling back', error);
```
adamyeats, line 331: nit: no test drives a throwing target, so this guard is unexercised by the suite.

src/data/logs.ts
```diff
@@ -83,6 +87,45 @@ export const LOG_LEVEL_TO_IN_CLAUSE: LogLevelToInClause = (() => {
...
+  const levelExpression = `ifNull(toString(${quotedLevelColumn}), '')`;
+
+  let level: keyof LogLevelToInClause;
+  for (level in LOG_LEVEL_TO_IN_CLAUSE) {
+    aggregates.push({
+      alias: level,
+      expression:
+        level === 'unknown'
+          ? `${levelExpression} NOT IN (${KNOWN_LOG_LEVEL_IN_CLAUSE})`
+          : `${levelExpression} IN (${LOG_LEVEL_TO_IN_CLAUSE[level]})`,
```
adamyeats: Swapping `multiSearchAny` for exact IN moves values that used to match by substring into `unknown`. Against ClickHouse, `'ERROR: connection refused'` and `'level=error'` were counted as error before and are unknown now, and `emergency` leaves `critical` because the list only carries `emerg`. The CHANGELOG frames this purely as a fix, so it might be worth calling the reattribution out.

src/data/logs.ts
```diff
@@ -83,6 +87,45 @@ export const LOG_LEVEL_TO_IN_CLAUSE: LogLevelToInClause = (() => {
...
+  const aggregates: Array<{ alias: keyof LogLevelToInClause; expression: string }> = [];
+  const levelExpression = `ifNull(toString(${quotedLevelColumn}), '')`;
+
+  let level: keyof LogLevelToInClause;
+  for (level in LOG_LEVEL_TO_IN_CLAUSE) {
+    aggregates.push({
+      alias: level,
+      expression:
+        level === 'unknown'
+          ? `${levelExpression} NOT IN (${KNOWN_LOG_LEVEL_IN_CLAUSE})`
```
adamyeats: Grafana's `LogLevel` enum maps `notice` to info and `emergency` to critical, so the complement now positively draws those rows in the grey unknown band while the log lines below stay green and red. Adding `notice` to the info list and `emergency` to the critical list would keep the two views in agreement.

src/data/CHDatasource.test.ts
```diff
@@ -1522,19 +1546,109 @@ describe('ClickHouseDatasource', () => {
...
+            { ...sqlLogsTarget('SELECT anything FROM nowhere SETTINGS x = 1'), hide: true },
+          ],
+        };
+        expect(datasource.getSupportedSupplementaryQueryTypes(dsRequest)).toEqual([
+          SupplementaryQueryType.LogsVolume,
+          SupplementaryQueryType.LogsSample,
+        ]);
+      });
+
+      it('should ignore non-logs targets when deciding volume support', async () => {
```
adamyeats: This passes because `SELECT count() FROM whatever` does not aggregate, not because the target declares Table. A target whose first column is the configured timestamp is admitted despite the Table type, which is the case worth pinning here.

src/data/CHDatasource.test.ts
```diff
@@ -1697,6 +1813,95 @@ describe('ClickHouseDatasource', () => {
...
+      });
+
+      it('should wrap the user SQL as a derived table and drop the trailing ORDER BY and LIMIT', () => {
+        const result = datasource.getSupplementaryLogsVolumeQuery(
+          request,
+          sqlLogsTarget("SELECT created_at, level FROM logs WHERE level = 'ERROR' ORDER BY created_at DESC LIMIT 100")
+        );
+
+        expect(result).toBeDefined();
+        expect(result?.editorType).toBe(EditorType.SQL);
```
adamyeats, line 1879: nit: `format: 0` is what shapes the response into the stacked series Explore draws, and nothing in this block asserts it. The sibling builder-path test does assert format.

src/ch-parser/spans.test.ts
```diff
@@ -0,0 +1,693 @@
...
+      expect(scanOk('SELECT t.* FROM t').items).toEqual([{}]);
+    });
+
+    it('does not report `* EXCEPT (...)` as a plain wildcard', () => {
+      // NOTE: the scanner currently declines this statement outright (EXCEPT is matched by text
+      // at depth 0, so it is reported as a set operation) rather than returning an item with no
+      // names. Either way the caller must not believe it selected every column, which is the
+      // property asserted here.
+      const result = scanTopLevelSelect('SELECT * EXCEPT (Body) FROM t');
+      const items = result.ok ? result.select.items : [];
```
adamyeats: The `: []` fallback makes the next assertion pass whenever the scan declines. The comment above also says the scanner declines `SELECT * EXCEPT`, which the test at line 681 contradicts.

src/data/logsVolumeSql.test.ts
```diff
@@ -0,0 +1,451 @@
...
+
+  it('is unaffected for an explicit projection over an unknown table', () => {
+    // No wildcard, so only names the projection provides are referenced and the table is
+    // irrelevant.
+    expect(planSqlLogsVolume('SELECT Timestamp, SeverityText FROM anything', HOURLY, columns()).ok).toBe(true);
+  });
+});
+
+describe('planSqlLogsVolume statement macros', () => {
+  it.each(['$__columns(Timestamp)', '$__rateColumns(a, b)', '$__perSecondColumns(a, b)', '$__lttb(a, b)'])(
```
adamyeats: nit: `STATEMENT_MACROS` lists five macros, and `$__increaseColumns` is not a substring of any of these four, so it is genuinely uncovered.

src/data/logsVolumeSql.test.ts
```diff
@@ -0,0 +1,451 @@
...
+
+    it('declines when no timestamp column is configured', () => {
+      expectDeclined(planSqlLogsVolume('SELECT Timestamp FROM t', HOURLY, new Map()), 'NoTimeColumnConfigured');
+    });
+
+    // The volume request snaps the interval to a coarser bucket, so an interval macro would
+    // expand to a different value here than in the query the user sees.
+    it.each([
+      'SELECT Timestamp AS timestamp FROM t WHERE Duration > $__interval',
+      'SELECT Timestamp AS timestamp FROM t WHERE DurationMs > $__interval_ms',
```
adamyeats: nit: the check is `head.includes(macro)`, so this case is satisfied by the `$__interval` entry alone, and `$__interval_s` has no case at all.

src/data/logsVolumeSql.test.ts
```diff
@@ -0,0 +1,451 @@
...
+  });
+
+  describe('bucket granularity', () => {
+    const head = 'SELECT Timestamp FROM t';
+
+    it.each([
+      { bucket: 'SECOND', intervalMs: SECOND },
+      { bucket: 'MINUTE', intervalMs: SECOND + 1 },
+      { bucket: 'HOUR', intervalMs: MINUTE + 1 },
+      { bucket: 'DAY', intervalMs: HOUR + 1 },
```
adamyeats, line 252: nit: the branches compare with `>`, and the table only uses `+1` values, so exactly `MINUTE` and exactly `HOUR` are untested. The newly exported `getTimeFieldRoundingInterval` has no direct test either.

tests/e2e/fixtures/demo_otel_logs.sql
```diff
@@ -0,0 +1,108 @@
+-- Demo data for manual testing of the logs experience, in its own `demo` database so it never
```
adamyeats: The compose loader runs every `/data/*.sql`, so this executes on every e2e run across the whole matrix and inserts 252,000 rows, and no spec references it. Given the PR generates raw ClickHouse SQL that nothing in the suite ever executes against a server, an e2e spec that uses this fixture would earn its place.

tests/e2e/fixtures/demo_otel_logs.sql
```diff
@@ -0,0 +1,108 @@
...
+-- collides with the e2e fixtures that tests assert against.
+--
+-- The schema matches the opentelemetry-collector-contrib clickhouseexporter v0.151.0 layout,
+-- which is what the plugin's `latest` OTel version expects (src/otel.ts). Turning on the OTel
+-- toggle in the data source's Logs settings therefore configures every column automatically.
+--
+-- Rows are generated relative to load time, so the data always covers the last 24 hours, at
+-- roughly 170 lines per minute — dense enough that minute-wide histogram buckets read as real
+-- traffic rather than ones and twos. A burst of errors around four hours ago gives the
+-- histogram a recognisable shape.
```
adamyeats: nit: US spelling in code comments for grafana-org repos, so `recognizable`. CI will not catch it, because the spellcheck glob does not include `.sql`.

CHANGELOG.md
```diff
@@ -4,11 +4,14 @@
 
 ### Features
 
+- Show the full time range in the log volume histogram for Logs queries written in the SQL editor, instead of a histogram capped by the query's `LIMIT`. Requires the default log table and columns to be configured under **Logs** in the data source settings; queries the plugin cannot aggregate safely keep the previous behavior (#XXXX)
```
adamyeats: Three entries still carry the literal `(#XXXX)` placeholder, on lines 7, 13 and 14.

docs/sources/query-editor.md
```diff
@@ -155,10 +155,18 @@ JSON-type attribute columns require ClickHouse 26 or later and the OTel Collecto
...
+
+Log volume works for queries written in either the **query builder** or the **SQL editor**. For a SQL editor query, the plugin aggregates over your query as a subquery, so your filters, joins, CTEs, macros, and template variables are preserved, and it removes your `ORDER BY` and `LIMIT` so the histogram covers the whole selected time range instead of just the rows your query returns. Log level series require a log level column configured under **Logs** in the data source settings and projected by your query.
+
+When the plugin cannot guarantee a correct count it declines, and Grafana draws its row-based histogram instead. This happens when no default log columns are configured, when your query does not select the configured timestamp column first or selects it as an expression, when a `SELECT *` query reads a table other than the configured logs table, or when the query uses a construct that changes which rows exist, such as `SETTINGS`, `LIMIT n BY`, `WITH FILL`, `WITH TOTALS`, `UNION`, an interval macro, or multiple statements.
```
adamyeats: nit: the second sentence runs to 69 words with four stacked `when` clauses. `troubleshooting.md` already presents the same content as a bullet list, which reads much better.

adamyeats (changes requested):
@SpencerTorres could we add an E2E spec for the SQL editor path before this lands? Nothing in the suite runs the generated SQL against a server, and the fixture plus the `traceLogsLink.spec.ts` pattern are already in place.

Two notes that do not sit on changed lines, so they could not go inline:

- `src/data/logs.ts:78` (`unknown: ['unknown'],`): only the key is used now, so this value list is dead and adding to it later would silently do nothing.
- `docs/sources/_index.md:44`: the requirements bullet still says SQL editor log volume needs Grafana 12.4.0, which the two pages this PR rewrites now contradict.

adamyeats (commented):
Could we drop the `CHANGELOG.md` hunk? release-please generates the changelog from the squash commit subject, so this PR will produce a single generated line, `* show log volume for SQL editor queries (#2141)`, and the five manual entries here would sit alongside it as duplicates.

nit: the old `should ignore non-logs targets when deciding volume support` test still passes because `SELECT count() FROM whatever` does not aggregate, rather than because the target declares Table. The two new tests cover that case properly, so this one could go.

teeohhem:
@adamyeats Looks like this is ready for another review :) TY

adamyeats (approved):
All 24 threads addressed and verified against the updated branch, including the E2E spec, the LIMIT/format/Date32 fixes, and the CHANGELOG removal. Merged locally onto current main with no conflicts, 499 unit tests and tsc pass.

PR #2145, opened 2026-08-27
feat(dashboards): let Query Analysis read any query_log source

AndreZiviani:
## Type of Change

- [x] 🚀 Feature

---

### What is this feature?

The bundled **ClickHouse - Query Analysis** dashboard no longer hardcodes `system.query_log`. It reads the source from a new `query_log` constant variable that holds the fully qualified name and defaults to `system.query_log`, so the shipped behaviour is unchanged. Editing that one variable in dashboard settings repoints every panel *and* every filter variable (`Query kind`, `Query status`, `User`).

The "All" filters were also reworked to use the `$__conditionalAll` macro:

```sql
FROM ${query_log}
WHERE
  $__conditionalAll(type IN (${type:singlequote}), $type)
  AND $__conditionalAll(initial_user IN (${user:singlequote}), $user)
  AND $__conditionalAll(query_kind IN (${query_kind:singlequote}), $query_kind)
  AND $__timeFilter(event_time)
```

Variables are interpolated with `:singlequote` in the condition and passed as a plain `$var` in the macro's second argument, per [template-variables.md](https://github.com/grafana/clickhouse-datasource/blob/main/docs/sources/template-variables.md).

Three drive-by fixes in the same queries:

- **Top users** panel had no time filter, so it counted the whole retained log regardless of the picker.
- **Query requests by user** subquery had `ORDER BY count() as c DESC` (alias definition in `ORDER BY`).
- SQL was reflowed multi-line; it was previously unreadable single-line strings.

`system.query_log` is local to whichever node Grafana's connection lands on, and it is truncated by the node's retention/rotation settings. Anyone with a cluster-wide or rotation-aware view over the query log (e.g. `clusterAllReplicas`-backed, or a merge over rotated tables) had to fork the dashboard to point it at that view — a fork that then has to be re-applied on every plugin release.

Separately, with `IN ($var)` the "All" selection expands into a full literal `IN` list of every value, which is both slower and breaks as soon as a value appears that was not in the variable's option list. `$__conditionalAll` drops the clause entirely instead.

### Who is this feature for?

Operators running multi-node/clustered ClickHouse who keep a curated query-log view, and anyone who currently sees partial data in this dashboard because the log rotated or lives on another replica.

### How to test this feature

1. Import (or re-import) the **ClickHouse - Query Analysis** dashboard and confirm every panel still works unchanged against `system.query_log`.
2. Set the **Query kind** / **Query status** / **User** variables to `All` and inspect a panel's query in the panel inspector — the clause should be `1=1`, not an expanded `IN (...)` list. Then select a single value and confirm it is quoted (`type IN ('QueryFinish')`).
3. Create a view, e.g. `CREATE VIEW default.query_log_all AS SELECT * FROM clusterAllReplicas(default, system.query_log)`, then set **Dashboard settings → Variables → Query log table** to `default.query_log_all`. Every panel and the three filter drop-downs should now read from the view.

---

## Behaviour changes worth calling out

- **"Top query types over time"** gains an `initial_user` filter in its top-5 hash subquery that main did not have, so the top-5 set now follows the User selection instead of being computed cluster-wide and then filtered.
- **"Query performance by type over time"** regains the outer `type` filter. On main it applied the Query status only inside the hash subquery, so the outer average still covered every type for those hashes.
- **"Memory usage over time"** and **"Read vs Write Rows"** now honour the Query status, User and query-kind variables; previously they read the whole `query_log` regardless. "Read vs Write Rows" also bucketed in its `SELECT` while grouping by the raw `event_time`, so the bucketing did nothing and the frame carried repeated timestamps — it now buckets with `$__timeInterval` and groups by that.

## Please check that:

`src/dashboards/__tests__/dashboards.test.ts` now also guards this dashboard directly: `query_log` stays a constant defaulting to the local system table, every panel reads through `${query_log}`, and each panel constrains `type`, `initial_user` and `query_kind` — either from the variable or, as "Query requests by user" does, with its own predicate. Both guards were checked by mutation (hard-coding one source, and dropping the filters from one panel) rather than only by passing. `docs/sources/dashboards.md` documents only the four OTel dashboards, so there is no page describing Query Analysis to update — happy to add one if you'd like it covered.

- The variable is a **constant** (`hide: 2`), not a query variable, so the dashboard makes no extra `system.tables` lookup on load and no new drop-down appears in the nav bar. It is changed in dashboard settings.
- It is named `query_log` rather than `database` because it holds `db.table`: a premade view is not necessarily named `query_log` in its own database.
- No screenshots: the rendered dashboard is visually identical with the default value, and the queries were verified by reading the interpolated SQL rather than against a live cluster.
- All commits are signed.
- Rebased onto #2083 (`fix/dashboards-multivalue-quoting`) as suggested, so the `${type:singlequote}` conversions are the base and this PR's `$__conditionalAll(... IN (${type:singlequote}), $type)` rewrites supersede them. The diff will shrink to this PR's own commits once #2083 lands.

### DIFF
--- a/src/dashboards/query-analysis.json
+++ b/src/dashboards/query-analysis.json
@@ -138,7 +138,7 @@
             }
           },
           "queryType": "sql",
-          "rawSql": "SELECT count() as \"Total queries\" FROM system.query_log WHERE type in ($type) AND initial_user IN ($user) AND query_kind IN ($query_kind) AND $__timeFilter(event_time) ",
+          "rawSql": "SELECT count() as \"Total queries\"\nFROM ${query_log}\nWHERE\n  $__conditionalAll(type IN (${type:singlequote}), $type)\n  AND $__conditionalAll(initial_user IN (${user:singlequote}), $user)\n  AND $__conditionalAll(query_kind IN (${query_kind:singlequote}), $query_kind)\n  AND $__timeFilter(event_time)",
           "refId": "A"
         }
       ],
@@ -220,7 +220,7 @@
             }
           },
           "queryType": "sql",
-          "rawSql": "SELECT avg(memory_usage) as \"Avg query memory\", $__timeInterval(query_start_time) as time FROM system.query_log WHERE type in ($type) AND initial_user IN ($user) AND query_kind IN ($query_kind) AND $__timeFilter(event_time) GROUP BY time ORDER BY time",
+          "rawSql": "SELECT\n  avg(memory_usage) as \"Avg query memory\",\n  $__timeInterval(query_start_time) as time\nFROM ${query_log}\nWHERE\n  $__conditionalAll(type IN (${type:singlequote}), $type)\n  AND $__conditionalAll(initial_user IN (${user:singlequote}), $user)\n  AND $__conditionalAll(query_kind IN (${query_kind:singlequote}), $query_kind)\n  AND $__timeFilter(event_time)\nGROUP BY time\nORDER BY time",
           "refId": "A"
         }
       ],
@@ -308,7 +308,7 @@
             }
           },
           "queryType": "sql",
-          "rawSql": "SELECT query_duration_ms as \"Query time\" FROM system.query_log WHERE type in ($type) AND initial_user IN ($user) AND query_kind IN ($query_kind) AND $__timeFilter(event_time) LIMIT 1000;",
+          "rawSql": "SELECT query_duration_ms as \"Query time\"\nFROM ${query_log}\nWHERE\n  $__conditionalAll(type IN (${type:singlequote}), $type)\n  AND $__conditionalAll(initial_user IN (${user:singlequote}), $user)\n  AND $__conditionalAll(query_kind IN (${query_kind:singlequote}), $query_kind)\n  AND $__timeFilter(event_time)\nLIMIT 1000",
           "refId": "A"
         }
       ],
@@ -380,7 +380,7 @@
             }
           },
           "queryType": "sql",
-          "rawSql": "SELECT initial_user, count() as c FROM system.query_log WHERE type in ($type) AND initial_user IN ($user) AND query_kind IN ($query_kind) GROUP BY initial_user LIMIT 100;\n",
+          "rawSql": "SELECT\n  initial_user,\n  count() as c\nFROM ${query_log}\nWHERE\n  $__conditionalAll(type IN (${type:singlequote}), $type)\n  AND $__conditionalAll(initial_user IN (${user:singlequote}), $user)\n  AND $__conditionalAll(query_kind IN (${query_kind:singlequote}), $query_kind)\n  AND $__timeFilter(event_time)\nGROUP BY initial_user\nLIMIT 100",
           "refId": "A"
         }
       ],
@@ -463,7 +463,7 @@
             }
           },
           "queryType": "sql",
-          "rawSql": "SELECT avg(query_duration_ms) as \"Avg query time\", $__timeInterval(query_start_time) as time FROM system.query_log WHERE type in ($type) AND initial_user IN ($user) AND query_kind IN ($query_kind) AND $__timeFilter(event_time) GROUP BY time ORDER BY time",
+          "rawSql": "SELECT\n  avg(query_duration_ms) as \"Avg query time\",\n  $__timeInterval(query_start_time) as time\nFROM ${query_log}\nWHERE\n  $__conditionalAll(type IN (${type:singlequote}), $type)\n  AND $__conditionalAll(initial_user IN (${user:singlequote}), $user)\n  AND $__conditionalAll(query_kind IN (${query_kind:singlequote}), $query_kind)\n  AND $__timeFilter(event_time)\nGROUP BY time\nORDER BY time",
           "refId": "A"
         }
       ],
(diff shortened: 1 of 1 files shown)


src/dashboards/query-analysis.json
```diff
@@ -670,7 +670,7 @@
...
+          "rawSql": "SELECT\n  $__timeInterval(query_start_time) as time,\n  any(normalizeQuery(query)) AS normalized_query,\n  avg(query_duration_ms) as avg_query_duration\nFROM ${query_log}\nWHERE type != 'QueryStart'\n  AND $__conditionalAll(initial_user IN (${user:singlequote}), $user)\n  AND $__conditionalAll(query_kind IN (${query_kind:singlequote}), $query_kind)\n  AND $__timeFilter(event_time)\n  AND normalized_query_hash IN (\n    SELECT normalized_query_hash\n    FROM ${query_log}\n    WHERE\n      $__conditionalAll(type IN (${type:singlequote}), $type)\n      AND $__conditionalAll(initial_user IN (${user:singlequote}), $user)\n      AND $__conditionalAll(query_kind IN (${query_kind:singlequote}), $query_kind)\n      AND $__timeFilter(event_time)\n    GROUP BY normalized_query_hash\n    ORDER BY avg(query_duration_ms) DESC\n    LIMIT 10\n  )\nGROUP BY normalized_query_hash, time\nORDER BY time",
```
adamyeats, line 673: The outer `WHERE` on "Query performance by type over time" has lost its type filter. On main it read `AND type in ($type)` alongside the user and query-kind filters. Here `$type` survives only inside the `normalized_query_hash` subquery, so with one type selected the outer aggregation still averages every type for those ten hashes. Consider restoring `$__conditionalAll(type IN (${type:singlequote}), $type)` on the outer query.

src/dashboards/query-analysis.json
```diff
@@ -562,7 +562,7 @@
...
+          "rawSql": "SELECT\n  $__timeInterval(query_start_time) as time,\n  any(normalizeQuery(query)) AS normalized_query,\n  count() as c\nFROM ${query_log}\nWHERE type != 'QueryStart'\n  AND $__conditionalAll(initial_user IN (${user:singlequote}), $user)\n  AND $__conditionalAll(query_kind IN (${query_kind:singlequote}), $query_kind)\n  AND $__timeFilter(event_time)\n  AND normalized_query_hash IN (\n    SELECT normalized_query_hash\n    FROM ${query_log}\n    WHERE\n      $__conditionalAll(type IN (${type:singlequote}), $type)\n      AND $__conditionalAll(initial_user IN (${user:singlequote}), $user)\n      AND $__conditionalAll(query_kind IN (${query_kind:singlequote}), $query_kind)\n      AND $__timeFilter(event_time)\n    GROUP BY normalized_query_hash\n    ORDER BY count() DESC\n    LIMIT 5\n  )\nGROUP BY normalized_query_hash, time\nORDER BY time",
```
adamyeats, line 565: nit: the top-5 hash subquery on "Top query types over time" gains an `initial_user` filter that main did not have. It reads as the right call, but it is a behaviour change worth a line in the body.

src/dashboards/query-analysis.json
```diff
@@ -1018,7 +1018,7 @@
             }
           },
           "queryType": "sql",
-          "rawSql": "SELECT toStartOfInterval(toDateTime(event_time), INTERVAL 60 second),  sum(read_rows) read_rows, sum(written_rows) written_rows FROM system.query_log WHERE $__timeFilter(event_time) GROUP BY event_time ORDER BY event_time ASC",
+          "rawSql": "SELECT\n  toStartOfInterval(toDateTime(event_time), INTERVAL 60 second),\n  sum(read_rows) read_rows,\n  sum(written_rows) written_rows\nFROM ${query_log}\nWHERE $__timeFilter(event_time)\nGROUP BY event_time\nORDER BY event_time ASC",
```
adamyeats, line 1021: nit: this panel selects `toStartOfInterval(toDateTime(event_time), 60s)` but groups and orders by the raw `event_time`, so the bucketing does nothing. Pre-existing, but since the line was reflowed here it might be worth grouping by the bucket.

adamyeats (changes requested):
This PR rewrites the same nine `rawSql` strings in `query-analysis.json` that #2083 quotes, so the two conflict. #2083 is close to landing, so it might be worth rebasing onto it once merged and letting `${type:singlequote}` supersede the quoting there.

Separately, the template ticks tests added, but the diff touches no test file. The existing datasource-reference guard runs over this file, but nothing asserts the new behaviour.

AndreZiviani:
Thanks @adamyeats — all three addressed, and the branch now sits on #2083.

- **Stacking:** rebased onto `fix/dashboards-multivalue-quoting`, so the `${type:singlequote}` conversions are the base and this PR's `$__conditionalAll(... IN (${type:singlequote}), $type)` rewrites supersede them, as you suggested. The diff shrinks to this PR's own commits once #2083 lands.
- **Outer type filter:** restored on "Query performance by type over time" — the outer aggregation was averaging every type for the ten selected hashes.
- **Read vs Write Rows:** now buckets with `$__timeInterval(event_time)` and groups by that, instead of bucketing in the `SELECT` and grouping by the raw second. It also honours the three variables, as does "Memory usage over time", which additionally no longer hands Grafana a reverse-sorted time field.
- **`initial_user` in the top-5 subquery:** kept, and now called out in the description as a behaviour change.

**Tests:** the template's tick was wrong before — there are now guards for it. `query_log` stays a constant defaulting to `system.query_log`, every panel reads through `${query_log}`, and each panel constrains `type`, `initial_user` and `query_kind`, either from the variable or with its own predicate ("Query requests by user" picks its own top ten, as it does on main). Both guards were checked by mutation — hard-coding one source, and stripping the filters from one panel — rather than only by passing.

Dashboards suite green (51/51).

adamyeats:
@AndreZiviani #2083 is on main now, so the #2083 commits this branch carries can go. After a rebase onto main the branch conflicts in `src/dashboards/__tests__/dashboards.test.ts` and `src/dashboards/query-analysis.json`, both from your own side of the stack. Your three fixes check out, so I will re-review as soon as CI is green.

AndreZiviani:
Rebased onto main. The three #2083 commits this branch carried are gone — I replayed only this PR's own two commits with `--onto`, since #2083 landed as a squash and its individual patches would not have been recognised as already applied.

The diff is now just this PR: `query-analysis.json` and its guards in `dashboards.test.ts`, 90 insertions. Main's `${type:singlequote}` conversions are the base, and this PR's `$__conditionalAll(type IN (${type:singlequote}), $type)` rewrites sit on top of them, as you suggested.

Re-checked after the rebase: dashboards suite green (51/51), and both guards still fail under mutation — hard-coding one panel's source to `system.query_log`, and stripping the variable filters from one panel.

adamyeats (approved):
Checked the rebase against all three threads. The branch now carries only its own two commits, every panel reads through the query_log variable and constrains type, user and query kind, and the outer type filter and the bucketing fix are in place. Approving.

src/dashboards/query-analysis.json
adamyeats: Addressed in the 21 September push (d8150e2). Resolving.

src/dashboards/query-analysis.json
adamyeats: Addressed in the 21 September push (d8150e2). Resolving.

src/dashboards/query-analysis.json
adamyeats: Addressed in the 21 September push (d8150e2). Resolving.

Issue #1258, opened 2025-05-01
PDC is failing


alyssajoyner:
PDC is currently failing when attempting to establish a connection to ClickHouse.

Slach:
any news about this bug?

dafydd-t:
We have some good descriptions of the observed behaviour in https://github.com/grafana/pdc-agent/issues/212. https://github.com/grafana/pdc-agent/issues/152 is another bug report of this behaviour.

PR #2146, opened 2026-08-27
fix(pdc): apply TLS to native protocol connections made through PDC

adamyeats:
## Type of Change

- [x] 🐛 Bug Fix

---

### What is the bug?

A data source that uses the native protocol through Private data source connect ignores its TLS settings.

`clickhouse-go` applies `Options.TLS` only on its own dial path. It skips that path when `Options.DialContext` is set, which is how the proxy dialer works. The connection is dialled without TLS. A ClickHouse instance that requires TLS on its native port then refuses it.

The dialer now completes the TLS handshake itself for the native protocol. The HTTP protocol stays unwrapped, because `net/http` applies `Transport.TLSClientConfig` over a custom dialer and a second handshake breaks it.

Three decisions a reviewer cannot infer from the diff:

- The TLS config is cloned, because `clickhouse.Options.TLS` shares the same value.
- `tls.Client` does not derive the SNI name from the dial address. `tls.Dial` does. When `ServerName` is empty, the dialer fills it in. An explicit `ServerName` takes precedence.
- `proxyDialContext` takes the base dialer as a parameter. This seam makes the behaviour testable without a real SOCKS proxy and certificate material.

### How to reproduce

1. Enable TLS on the native port of a ClickHouse instance (`tcp_port_secure`, 9440).
2. Configure a secure SOCKS proxy that can reach that instance.
3. Add a ClickHouse data source.
4. Set Protocol to Native. Enable TLS and Private data source connect.
5. Select **Save & test**.

Before this change the test fails. After it the handshake completes.

### Related Issues

Related to #1258 for the TLS case. That issue collects several causes, so this PR does not close it.

---

## Please check that:

### DIFF
--- a/pkg/plugin/driver.go
+++ b/pkg/plugin/driver.go
@@ -69,7 +69,13 @@ func getTLSConfig(settings Settings) (*tls.Config, error) {
 }
 
 // getPDCDialContext returns a dialer function for creating a connection to PDC if a secure SOCKS proxy is enabled.
-func getPDCDialContext(settings Settings) (func(context.Context, string) (net.Conn, error), error) {
+//
+// clickhouse-go applies Options.TLS only on its own dial path, and skips that
+// path once Options.DialContext is set. The native protocol therefore needs
+// this dialer to complete the TLS handshake itself. HTTP does not: net/http
+// applies Transport.TLSClientConfig over the dialed connection, and a second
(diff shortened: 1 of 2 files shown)


mattvella07 (approved):
LGTM

aangelisc (approved):

PR #2149, opened 2026-08-27
feat(query): add a per-query min interval override

AndreZiviani:
## Type of Change

- [x] 🚀 Feature

### What is this feature?

A **Min interval** field in the query editor, directly under **Query type**, that raises the interval for one query. Leaving it empty keeps the interval Grafana derives, so existing queries are unaffected.

It is a floor, matching the semantics of the panel-level **Min interval**: the effective interval is `max(derived, configured)`. Setting a value finer than what Grafana picked does nothing.

Explore has no panel query options, so a query there always buckets at whatever interval Grafana derives from the time range and pane width. There is no way to say "never bucket finer than 1m" for a single query — the closest lever is the panel-level **Min interval**, which Explore does not offer, and which is per-panel rather than per-query anyway.

### Who is this feature for?

Anyone exploring high-cardinality time series in Explore, where the derived interval produces buckets far finer than the question being asked, and anyone who wants two queries in the same panel to bucket differently.

### How to test this feature

1. In Explore, pick a ClickHouse data source and a time range of a few hours.
2. In the SQL editor, run `SELECT $__timeInterval(timestamp) AS time, count() FROM <db>.<table> WHERE $__timeFilter(timestamp) GROUP BY time ORDER BY time`.
3. Open the query inspector and note the `INTERVAL n second` Grafana derived.
4. Set **Min interval** to `5m` and re-run: the inspector shows `INTERVAL 300 second`, and the graph buckets accordingly.
5. Set it to `1s` and re-run: nothing changes, because the field never lowers the interval.
6. Type `1mo` (or `60`, or `1h30m`): the field goes red, the value is not committed, and the query keeps running at the derived interval.
7. Repeat step 4 in the query builder with a Time series query — the same field sits under **Query type** and the generated `$__timeInterval` follows it.

## Screenshots

**SQL editor — `Min interval` set to `5m`**, buckets held at 5 minutes on a range where Grafana would have picked a finer interval:

sql-editor-min-interval-5m

**Query builder — the same field under Query Type**, with the generated `$__timeInterval(timestamp)` in the SQL preview:

query-builder-min-interval-5m

**Invalid duration** — flagged inline and never committed, so the query keeps running at the interval Grafana derived:

sql-editor-invalid-duration

## How it works

The floor is applied on the backend, in `Clickhouse.MutateQuery`, which runs before `sqlutil.Query` is built (that struct drops unknown query fields, so the interpolator cannot see this one). Because it lands on `backend.DataQuery.Interval`, it reaches `$__timeInterval`, `$__timeInterval_ms`, `$__interval_s`, `$__timeGroup` and the statement macros in every query path — Explore, dashboard panels and alert rules alike.

`$__interval` and `$__interval_ms` are the exception: they are substituted from scoped vars in the frontend before the backend ever sees the SQL, so `applyTemplateVariables` mirrors the same floor onto those two, per target. The Explore log volume histogram takes its bucket from `scopedVars.__interval_ms` as well, so the floor is folded in there per target too.

### One grammar on both sides

Both sides accept exactly one form: a single integer plus a unit of `ms`, `s`, `m`, `h`, `d` or `w`, capped at 365d, with the same explicit surrounding-whitespace set trimmed on each side.

That is deliberately narrower than `rangeUtil.intervalToMs`, which disagrees with `gtime.ParseDuration` on everything the grammar excludes:

| Input | `rangeUtil.intervalToMs` | `gtime.ParseDuration` |
|---|---|---|
| `60` | 60s (unit-less branch) | error |
| `5minutes` | 5m (regex is unanchored) | error |
| `1.5m` | 1m (`parseInt` truncates) | 90s |
| `1h30m` | 1h (first unit only) | 90m |
| `1M` | 30d | 365.25/12 d |

Left unchecked, a query using both `$__interval` and `$__timeInterval` would bucket two columns at two different resolutions, or the editor would accept a value the backend silently drops. The 365d cap also rules out values that overflow `time.Duration` on the backend, which would skip the clamp there while the frontend still applied it. Anything outside the grammar is flagged inline in the editor and never committed.

The field is stored as `minInterval` on the query model, matching the label and the semantics; the backend decodes it leniently, so a hand-written dashboard carrying a number degrades to no floor instead of failing the whole decode.

## Please check that:

The field is rendered in the classic SQL editor and the classic builder (under the **Query type** row), and in the compact single-table builder it appears in the **Advanced** strip alongside Order by / Limit, where there is no Query type row.

Tests: 18 backend subtests over the clamp (raising, never lowering, unit handling, every rejected form, the numeric/null/object decode, and the shared whitespace set), plus frontend coverage of the parser, the scoped-var mirroring, the per-target logs volume floor, commit-on-blur/Enter, and that a min-interval edit preserves a hand-authored `format` and re-runs the query.

Verified against a live ClickHouse through a local Grafana, reading `executedQueryString`:

```
intervalMs=30000, minInterval='5m'            -> INTERVAL 300 second
intervalMs=30000, minInterval=' 2h '          -> INTERVAL 7200 second
intervalMs=900000, minInterval='1s'           -> INTERVAL 900 second   (never lowers)
intervalMs=30000, minInterval='1h30m'         -> INTERVAL 30 second    (rejected)
intervalMs=30000, minInterval='60'            -> INTERVAL 30 second    (rejected)
intervalMs=30000, minInterval='5minutes'      -> INTERVAL 30 second    (rejected)
intervalMs=30000, minInterval='1.5m'          -> INTERVAL 30 second    (rejected)
intervalMs=30000, minInterval='1M'            -> INTERVAL 30 second    (rejected)
intervalMs=30000, minInterval='366d'          -> INTERVAL 30 second    (rejected)
intervalMs=30000, minInterval='9999999999d'   -> INTERVAL 30 second    (rejected)
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/plugin/driver.go
+++ b/pkg/plugin/driver.go
@@ -545,6 +545,49 @@ func interpolateMacros(_ context.Context, query *sqlutil.Query, _ json.RawMessag
 	return sql, nil
 }
 
+// minIntervalPattern is the grammar for a per-query min interval: one integer
+// plus one unit, anchored. It matches parseMinIntervalMs in
+// src/data/queryInterval.ts exactly — gtime.ParseDuration is deliberately not
+// used, because it accepts forms the frontend reads differently (compound
+// "1h30m", fractional "1.5m") and defines M/y with a different length, which
+// would let $__timeInterval and $__interval bucket one query two ways.
+var minIntervalPattern = regexp.MustCompile(`^(\d+)(ms|s|m|h|d|w)$`)
+
+// maxMinInterval bounds the floor. Larger values emit nonsense SQL and can
+// overflow time.Duration, which would silently drop the clamp here while the
+// frontend still applied it to $__interval.
+const maxMinInterval = 365 * 24 * time.Hour
+
+var minIntervalUnits = map[string]time.Duration{
+	"ms": time.Millisecond,
+	"s":  time.Second,
+	"m":  time.Minute,
+	"h":  time.Hour,
+	"d":  24 * time.Hour,
+	"w":  7 * 24 * time.Hour,
+}
+
+// parseMinInterval returns the per-query interval floor, or 0 when the value is
+// absent or outside the shared grammar (in which case no floor is applied).
+func parseMinInterval(value string) time.Duration {
+	matches := minIntervalPattern.FindStringSubmatch(strings.TrimSpace(value))
+	if matches == nil {
+		return 0
+	}
+
+	count, err := strconv.ParseInt(matches[1], 10, 64)
+	if err != nil {
+		return 0
+	}
+
+	unit := minIntervalUnits[matches[2]]
+	if count <= 0 || count > int64(maxMinInterval/unit) {
+		return 0
+	}
+
+	return time.Duration(count) * unit
+}
+
 func (h *Clickhouse) MutateQuery(ctx context.Context, req backend.DataQuery) (context.Context, backend.DataQuery) {
 	ctx, span := tracing.DefaultTracer().Start(ctx, "clickhouse mutate_query", trace.WithAttributes(
 		attribute.String("db.system", "clickhouse"),
@@ -582,10 +625,20 @@ func (h *Clickhouse) MutateQuery(ctx context.Context, req backend.DataQuery) (co
 			TimeZone string `json:"timezone"`
 		} `json:"meta"`
 		Format int `json:"format"`
+		// Per-query minimum interval. Raises the interval Grafana derived from
+		// the time range and panel width, so $__timeInterval and friends bucket
+		// no finer than this. Explore has no built-in equivalent.
+		TimeInterval string `json:"timeInterval"`
 	}
 
 	if err := json.Unmarshal(req.JSON, &dataQuery); err != nil {
 		return ctx, req
+	}
+
+	// Clamped here rather than in the interpolator because sqlutil.Query drops
+	// unknown query fields, and MutateQuery runs before it is built.
+	if minInterval := parseMinInterval(dataQuery.TimeInterval); minInterval > req.Interval {
+		req.Interval = minInterval
 	}
 
 	if dataQuery.Meta.TimeZone == "" {
--- a/src/components/MinIntervalEditor.tsx
+++ b/src/components/MinIntervalEditor.tsx
@@ -1,0 +1,69 @@
+import React, { useRef, useState } from 'react';
+import { InlineField, InlineFormLabel, Input } from '@grafana/ui';
+import labels from 'labels';
+import { selectors } from 'selectors';
+import { parseMinIntervalMs } from 'data/queryInterval';
+
+interface MinIntervalEditorProps {
+  minInterval?: string;
+  onMinIntervalChange: (minInterval: string) => void;
+}
+
+/**
+ * Draft state for a min interval input: local while typing, committed on blur or
+ * Enter, re-synced when Grafana replaces the query model underneath us (query
+ * history, pane switches). Shared so the inline and compact layouts differ only
+ * in their surrounding markup, never in behavior.
+ */
+export const useMinIntervalDraft = (minInterval: string | undefined, onCommit: (minInterval: string) => void) => {
+  const propValue = minInterval || '';
+  const [draft, setDraft] = useState<string>(propValue);
+  const lastPropValue = useRef<string>(propValue);
+  if (lastPropValue.current !== propValue) {
+    lastPropValue.current = propValue;
+    setDraft(propValue);
+  }
+
+  const invalid = draft.trim() !== '' && parseMinIntervalMs(draft) === undefined;
+
+  const commit = () => {
+    const trimmed = draft.trim();
+    if (!invalid && trimmed !== propValue) {
+      onCommit(trimmed);
+    }
+  };
+
+  return { draft, setDraft, invalid, commit };
+};
+
+export const MinIntervalEditor = (props: MinIntervalEditorProps) => {
+  const { draft, setDraft, invalid, commit } = useMinIntervalDraft(props.minInterval, props.onMinIntervalChange);
+  const { label, tooltip, placeholder, error } = labels.components.MinIntervalEditor;
+
+  return (
+    <InlineField
+      label={
+        <InlineFormLabel width={8} className="query-keyword" tooltip={tooltip}>
+          {label}
+        </InlineFormLabel>
+      }
+      invalid={invalid}
+      error={error}
+    >
+      <Input
+        data-testid={selectors.components.QueryEditor.MinIntervalEditor.input}
+        width={10}
+        value={draft}
+        invalid={invalid}
+        placeholder={placeholder}
+        onChange={(e) => setDraft(e.currentTarget.value)}
+        onBlur={commit}
+        onKeyDown={(e) => {
+          if (e.key === 'Enter') {
+            commit();
+          }
+        }}
+      />
+    </InlineField>
+  );
+};
--- a/src/data/CHDatasource.ts
+++ b/src/data/CHDatasource.ts
@@ -62,6 +62,7 @@ import {
   splitLogsVolumeFrames,
   TIME_FIELD_ALIAS,
 } from './logs';
+import { applyMinIntervalToScopedVars } from './queryInterval';
 import { escapeIdentifier, generateSql, getColumnByHint, logAliasToColumnHints } from './sqlGenerator';
 import { labelsFieldName, transformQueryResponseWithTraceAndLogLinks } from './utils';
 import { CHVariableSupport } from './CHVariableSupport';
@@ -528,7 +529,7 @@ export class Datasource
 
     // resolve template variables
     rawQuery = this.applyConditionalAll(rawQuery, templateSrvVariables);
-    rawQuery = this.replace(rawQuery, scoped) || '';
+    rawQuery = this.replace(rawQuery, applyMinIntervalToScopedVars(scoped, query.timeInterval)) || '';
 
     if (!this.skipAdHocFilter) {
       if (this.adHocFiltersStatus === AdHocFilterStatus.disabled && filters.length > 0) {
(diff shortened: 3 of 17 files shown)


pkg/plugin/driver.go
```diff
@@ -545,6 +545,49 @@ func interpolateMacros(_ context.Context, query *sqlutil.Query, _ json.RawMessag
 	return sql, nil
 }
 
+// minIntervalPattern is the grammar for a per-query min interval: one integer
+// plus one unit, anchored. It matches parseMinIntervalMs in
+// src/data/queryInterval.ts exactly — gtime.ParseDuration is deliberately not
```
adamyeats, line 550: `cspell` flags `gtime` here and `unparseable` at `driver_test.go:448`, and the spellcheck step in `CI / Test and build plugin` will fail on them. Both files pass on main. Either add the two words to `cspell.config.json` or reword them.

pkg/plugin/driver.go
```diff
@@ -582,12 +625,22 @@ func (h *Clickhouse) MutateQuery(ctx context.Context, req backend.DataQuery) (co
...
 		} `json:"meta"`
 		Format int `json:"format"`
+		// Per-query minimum interval. Raises the interval Grafana derived from
+		// the time range and panel width, so $__timeInterval and friends bucket
+		// no finer than this. Explore has no built-in equivalent.
+		TimeInterval string `json:"timeInterval"`
 	}
 
 	if err := json.Unmarshal(req.JSON, &dataQuery); err != nil {
 		return ctx, req
```
adamyeats, line 635: If `timeInterval` is not a string in the saved JSON, for example `60` from a hand-written dashboard, `json.Unmarshal` fails and this early return also skips the timezone handling below it. Consider decoding the field leniently, so a bad min interval degrades to no floor rather than dropping an unrelated feature.

src/data/queryInterval.ts
```diff
@@ -0,0 +1,76 @@
...
+ */
+export const MAX_MIN_INTERVAL_MS = 365 * UNIT_MS.d;
+
+/**
+ * Parses a per-query min interval ("10s", "1m", "1d") into milliseconds.
+ * Returns undefined for anything the backend would not apply, which means
+ * "no floor".
+ */
+export const parseMinIntervalMs = (minInterval?: string): number | undefined => {
+  const matches = minInterval?.trim().match(MIN_INTERVAL_PATTERN);
```
adamyeats, line 39: Same input on the frontend: `minInterval?.trim()` throws inside `applyTemplateVariables` when the saved value is a number. A `typeof minInterval === 'string'` guard keeps the query running at the derived interval.

src/types/sql.ts
```diff
@@ -14,6 +14,12 @@ export interface CHQueryBase extends DataQuery {
   editorType: EditorType;
   rawSql: string;
 
+  /**
+   * Per-query lower bound for the interval ("10s", "1m"). Overrides nothing when
+   * unset — the interval Grafana derives from the time range is used as-is.
+   */
+  timeInterval?: string;
```
adamyeats, line 21: The wire name is `timeInterval` while the UI, props and title all say min interval. This is the one name that cannot change after release without a `migrateCHQuery` step, so it is worth deciding now. `minInterval` would read as the floor it is.

src/components/MinIntervalEditor.tsx
```diff
@@ -0,0 +1,69 @@
...
+    <InlineField
+      label={
+        <InlineFormLabel width={8} className="query-keyword" tooltip={tooltip}>
+          {label}
+        </InlineFormLabel>
+      }
+      invalid={invalid}
+      error={error}
+    >
+      <Input
```
adamyeats, line 53: nit: the input has no accessible name. `InlineField` gets a `ReactNode` label so no `htmlFor` is generated, and the `Input` has no `id` or `aria-label`. An `aria-label` matching the visible label fixes it. The compact variant could expose the error through `aria-invalid` rather than only `title`.

src/data/CHDatasource.ts
```diff
@@ -528,7 +529,7 @@ export class Datasource
 
     // resolve template variables
     rawQuery = this.applyConditionalAll(rawQuery, templateSrvVariables);
-    rawQuery = this.replace(rawQuery, scoped) || '';
+    rawQuery = this.replace(rawQuery, applyMinIntervalToScopedVars(scoped, query.timeInterval)) || '';
```
adamyeats, line 532: nit: the logs-volume supplementary query near line 370 buckets from the un-floored `scopedVars.__interval_ms`, so a per-query min interval is ignored by the histogram while the panel-level one is honoured. Low impact, but worth a decision on whether the supplementary request should apply the floor per target too.

pkg/plugin/driver.go
```diff
@@ -545,6 +545,49 @@ func interpolateMacros(_ context.Context, query *sqlutil.Query, _ json.RawMessag
...
+	"s":  time.Second,
+	"m":  time.Minute,
+	"h":  time.Hour,
+	"d":  24 * time.Hour,
+	"w":  7 * 24 * time.Hour,
+}
+
+// parseMinInterval returns the per-query interval floor, or 0 when the value is
+// absent or outside the shared grammar (in which case no floor is applied).
+func parseMinInterval(value string) time.Duration {
```
adamyeats, line 572: nit: grammar parity is exact except for two whitespace code points. Go's `strings.TrimSpace` strips U+0085 but not U+FEFF, and JS `trim()` does the reverse, so a leading BOM is accepted by the editor and ignored by the backend. Negligible, noted because the body makes the parity guarantee explicit.

adamyeats (changes requested):
A new query-editor field with its own grammar and floor semantics ships with no docs change. The element table in `docs/sources/query-editor.md` lists Editor type, Run Query and Query type. It might be worth a row for Min interval: the accepted forms, that it never lowers the interval, and which macros it reaches in dashboards versus alert rules.

nit: the body does not follow the repo template (Type of Change, How to test, checklist) and says 15 backend subtests where the table has 14.

AndreZiviani:
Thanks @adamyeats — all seven addressed, rebased onto latest main, and the description rewritten to the repo template (the subtest count was wrong; there are 18 now).

- **cspell:** `gtime` added to the word list, since it names a real package and the comment explaining why we do *not* use it is worth keeping. `unparseable` was my invention, so it is reworded — no dictionary growth for that one.
- **Unmarshal early return:** the field is decoded as `json.RawMessage` and read separately, so a number degrades to no floor instead of failing the whole decode and taking the timezone handling with it. There is a regression test for exactly that: a numeric `minInterval` alongside a timezone.
- **Frontend throw:** `parseMinIntervalMs` guards `typeof !== 'string'` up front. Same reasoning — a saved number must cost the floor, not the query.
- **Wire name:** renamed to `minInterval`. You are right that this is the one that cannot change later, and it now matches the label, the props and the semantics.
- **Accessibility:** the input carries `aria-label`, `aria-invalid` and `aria-errormessage`, and the compact variant exposes the same error rather than relying on `title` alone.
- **Logs volume:** the floor is now folded in per target. Both volume builders take their bucket from `scopedVars.__interval_ms`, which is request-level, so the per-query value was being dropped there; there is a test asserting three targets in one request get their own buckets (none, raised, and a floor below the derived interval).
- **BOM:** good catch, and it mattered more than it looked — the parity claim is explicit. Both sides now trim the same explicit set (`\t\n\v\f\r`, space, U+0085, U+00A0, U+FEFF) instead of `trim()` / `TrimSpace`, with cases on each side.

**Docs:** added a **Min interval** row to the element table in `docs/sources/query-editor.md` covering the accepted forms, that it never lowers the interval, which macros it reaches, and that it applies in Explore, panels and alert rules alike.

Full suite green apart from `TLSSSLSettingsSection › shows CA certificate input when tlsAuthWithCACert is true`, which fails on clean main too (checked with this branch stashed).

adamyeats (dismissed):
Checked the 21 September push against all seven threads. All addressed, and the wire name is settled before release. Approving.

pkg/plugin/driver.go
adamyeats: Addressed in the 21 September push (8bd63d1). Resolving.

src/types/sql.ts
adamyeats: Addressed in the 21 September push (8bd63d1). Resolving.

pkg/plugin/driver.go
adamyeats: Addressed in the 21 September push (8bd63d1). Resolving.

src/components/MinIntervalEditor.tsx
adamyeats: Addressed in the 21 September push (8bd63d1). Resolving.

src/data/CHDatasource.ts
adamyeats: Addressed in the 21 September push (8bd63d1). Resolving.

pkg/plugin/driver.go
adamyeats: Addressed in the 21 September push (8bd63d1). Resolving.

src/data/queryInterval.ts
adamyeats: Addressed in the 21 September push (8bd63d1). Resolving.

adamyeats:
@AndreZiviani thanks for the push. All seven threads are resolved and the PR is approved. One thing still blocks the merge: CI's spellcheck flags `ufeff` at `pkg/plugin/driver.go:577` and `pkg/plugin/driver_test.go:449`. Local cspell can pass where CI fails, so it will not show on your side. Adding `ufeff` to the word list in `cspell.config.json` is the smallest fix. Once that is pushed I will re-approve and merge.

AndreZiviani:
Thanks @adamyeats — `ufeff` added to `cspell.config.json` and pushed, rebased onto latest main (which now has #2095 and #2083).

You were right that it would not show locally: the word only appears in the Go escape `"\u0085 ﻿"` and the test that covers the BOM case, and my local cspell cannot run at all on this machine's Node version.

`tsc`, `go build` and the min-interval Go tests are green.

adamyeats (approved):
Checked the rebase: the only change beyond replaying onto main is the cspell word list entry. Approving again.

PR #2151, opened 2026-08-30
feat(jwt): add allowCleartextJWTForwarding for proxy-terminated TLS

ringerc:
## Type of Change

- [x] 🚀 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧹 Refactor / Chore

---

### What is this feature?

A new datasource setting to deliberately bypass the restriction on JWT forwarding on cleartext connections, for use when a sidecar or proxy handles TLS termination and origination on behalf of Grafana.

The setting is **Allow cleartext JWT forwarding** (`allowCleartextJWTForwarding`). It is off by default, and only appears in the config editor when **Forward OAuth Identity** is enabled.

Today `buildClickHouseOptions` rejects the connection outright if **Forward OAuth Identity** is on and no TLS config was built:

```
JWT authentication requires a secure (TLS) connection
```

This is intended to protect against MitM replay attacks, which is reasonable. But it can't detect when there's a sidecar or co-located reverse proxy handling termination so it blocks a perfectly safe connection in those cases. For example, Istio Envoy is typically used to originate and terminate TLS for workloads, so the workload sees an "insecure" cleartext connection but there's no real exposure risk.

### Who is this feature for?

Operators running Grafana in a service mesh, or behind any local egress proxy that does TLS origination on the plugin's behalf, and who want per-user ClickHouse identities (row policies, quotas, query log attribution) via JWT pass-through.

### How to test this feature

Run the added tests:

1. `go test ./pkg/plugin/...`
2. Look at `TestBuildClickHouseOptionsJWTAllowsCleartextWhenOptedIn`. With `Secure = false` and the new setting on, `buildClickHouseOptions` succeeds, `opts.GetJWT` still returns the forwarded token, and `opts.TLS` is nil.
3. Look at `TestBuildClickHouseOptionsJWTCleartextOptInDoesNotAllowSkipTLSVerify`. With the new setting on *and* `InsecureSkipVerify` on, the connection is still rejected.
4. `TestBuildClickHouseOptionsJWTRequiresTLS` is untouched and still passes, confirming the default behaviour has not moved.

Frontend:

1. `npx jest src/views/CHConfigEditor.test.tsx src/views/config-v2/DatabaseCredentialsSection.test.tsx`
2. In a running Grafana, open a ClickHouse data source. The new toggle should not be visible. Turn on **Forward OAuth Identity** and it appears, below **Allow service account fallback**. It is present in both the classic config editor and the config-v2 editor.

End to end (simplified):

1. Deploy Grafana, ClickHouse and Istio into a `kind` or `micro8ks`. Configure Grafana for OAuth against a placeholder service like OAuth2-Mock-Server
2. Configure both ClickHouse and Grafana for Istio sidecar injection, with Istio policies requiring mTLS
3. Configure ClickHouse for JWT authentication
4. Configure a data source with `protocol: http`, **Secure connection** off,  **Forward OAuth Identity** on, pointed at the sidecar's local listener.
5. Run a query. Before this change it will fail with  "JWT authentication requires a secure (TLS) connection". With **Allow cleartext JWT forwarding** on, the query runs and the token reaches ClickHouse via the sidecar.
6. Turn the setting back off and confirm the query fails again.

---

## Please check that:

This change _does_ allow a misguided user to knowingly weaken a security safety-catch, but they must do so deliberately.

Mitigations:

- It is off by default, so nothing changes for anyone who does not go looking for it;
- It is only reachable once **Forward OAuth Identity** is already on, so it cannot be switched on by accident while poking at TLS settings; and
- The docs section (`Delegating TLS to a proxy`) states plainly that the setting disables a check and does not make the connection secure, and that tokens go out in the clear if the assumed proxy is not there.

No warning log was added. This is deliberate. The only place to emit one is per connection, which would be far too noisy.

Proxy detection isn't possible for outbound TLS origination. The plugin can't detect an incoming proxied request because Grafana sanitizes incoming requests to strip unknown headers - and in some contexts (like grafana alerts or connection tests) there are no client headers at all. The presence of an inbound proxy wouldn't prove anything about an outbound proxy anyway.

The setting is exposed in the UI. It could be made provisioning-only instead if desired.

On a side note, this doesn't change the way the Clickhouse plugin munges forwarded OAuth so it can be consumed by the native Clickhouse server's JWT auth either. So it won't help users whose sidecars/proxies want to check the JWT or manipulate the request based on it. But it does allow Clickhouse JWT auth to be used when Grafana delegates TLS to a sidecar.


docs/sources/configure.md
```diff
@@ -131,6 +131,7 @@ After adding the data source, configure the following settings.
 | **Username**          | ClickHouse user name. Use a [read-only user](#clickhouse-user-and-permissions).                                                                         |
 | **Password**          | ClickHouse user password.                                                                                                                               |
 | **Forward OAuth Identity** | Forward the logged-in Grafana user's OAuth token to ClickHouse as a JWT instead of authenticating with the configured username and password. ClickHouse Cloud only; requires a [secure (TLS) connection](#tls-settings). See [Forward OAuth Identity](#forward-oauth-identity). |
```
iwysiu, line 133: Does the cleartext JWT forwarding change the secure connection requirements?

docs/sources/configure.md
```diff
@@ -357,6 +370,7 @@ datasources:
       # forwardGrafanaHeaders: <bool>
       # oauthPassThru: <bool>  # forward the user's OAuth token as a JWT (ClickHouse Cloud only); requires secure: true
```
iwysiu, line 371: Similarly, is it still true that this requires secure:true? Same on line 416

iwysiu (commented):
Hi @ringerc ! Thanks for the the PR, I just a question about the docs. Also I think you have an unsigned commit.

docs/sources/configure.md
ringerc: Good catch. I'll amend this to say a TLS-protected connection is strongly recommended.

docs/sources/configure.md
ringerc: Good point. No longer enforced, to allow the use of TLS-originating proxies. Will amend to strongly recommended here too.

ringerc:
@iwysiu I've addressed those issues - thanks for taking a look. Sorry about the force push, I made an error in rebasing and I had to clean it up.

iwysiu (dismissed):
Hi @ringerc , the docs look good! I'll get this merged.

iwysiu:
Spellcheck issues again 😞 It's upset about "cleartext"

iwysiu:
As an update: Talked with the team, and we don't think we want this to be available in Grafana Cloud, so we want to make that we're gating that correctly. I'll get back to you once I have more information

PR #2152, opened 2026-08-31
fix(headers): fix keepCookies cookie-forwarding on http protocol

### LINKED ISSUE #2139: Clickhouse datasource does not respect Grafana's cookie or OAuth forwarding
**What happened**:

When using the Clickhouse HTTP protocol the Clickhouse datasource plugin plugin does not forward HTTP headers and cookies that Grafana is configured to forward.

Specifically:

- `Cookie` entries listed under the data source's `keepCookies`  option are not forwarded to ClickHouse; and
- `Authorization` and `X-Id-Token` are not sent to ClickHouse when `oauthPassThru: true` is set on the data source.

Setting `forwardGrafanaHeaders: true` on the data source does not fix this on its own, either alone or in combination with the two options above.

The Grafana Prometheus data source, given the equivalent `keepCookies` / `oauthPassThru` configuration, does forward these headers.

This breaks reverse-proxy setups that place an authorising HTTP proxy in front of ClickHouse and require a per-user cookie or bearer token on every request. The plugin's first outbound call to ClickHouse is an anonymous `queryHello` probe with none of those headers on it, so the proxy rejects it with 403 and the data source fails to connect.

With `keepCookies: [<cookie-name>]` set on the data source and the client request to Grafana carrying that cookie, the outbound HTTP request from the plugin to ClickHouse should include the cookie.

Same for `Authorization` and `X-Id-Token` when `oauthPassThru: true` is set.

[Due to an underlying Grafana limitation](https://github.com/grafana/grafana/issues/87364) general-purpose HTTP header forwarding cannot currently work anyway - but Grafana's cookie-forwarding and its OAuth `authorization_token` forwarding _should_ both work with ClickHouse when suitably enabled.

See attached self-contained repro test case: [repro-2139-v2.zip](https://github.com/user-attachments/files/31622127/repro-2139-v2.zip). It has an embedded README. Test case is LLM-assisted. TL;DR

```sh
unzip repro-2139-v2.zip
cd repro-2139-v2
docker compose up --build --abort-on-container-exit --exit-code-from verifier
```

then check to see log

```
repro-2139-verifier-1  | BUG A REPRODUCED: Cookie x_authorized_project_ids was NOT forwarded to the CH backend under forwardGrafanaHeaders + keepCookies (oauthPassThru=false). See pkg/plugin/driver.go Connect() -- the grafana-header-grafting path is gated on oauthPassThru only.
```

Alternately, explicit manual steps would be too long to list, but here's the general idea:

1. Put an HTTP proxy in front of ClickHouse that returns 403 on any request without `Cookie: my_auth_cookie=...`.
2. Configure a ClickHouse data source pointing at that proxy:

   ```yaml
   jsonData:
     protocol: http
     host: <the proxy>
     port: 8123
     username: grafana_ds
     keepCookies:
       - my_auth_cookie
     forwardGrafanaHeaders: true
   ```

3. Log in to Grafana with a browser session that carries a valid
   `my_auth_cookie`.
4. Run any query against the data source.

Observed: every `/api/ds/query` request fails with `failed to query server hello: [HTTP 403] "missing or invalid cookie"`. The proxy logs an inbound request from the plugin with no `Cookie` header on it.

The when using a matching setup with the Prometheus data source, cookie forwarding works.

**Screenshots**

N/A

Suspected cause: the plugin's `Connect` runs `db.PingContext` on every call, including the anonymous bootstrap call sqlds makes at data source instantiation with a nil per-query message. That probe carries no user context, so no cookie can attach to it, and the authorising proxy rejects it. `forwardGrafanaHeaders` and `oauthPassThru` only affect per-query dispatch, not this bootstrap call.

Use of a cookie to carry an opaque bearer token is an interim workaround for grafana/grafana#87364 where Grafana itself does not support an allow-list of additional headers to forward from inbound HTTP through to datasources for propagation. Grafana sanitizes them before the datasource plugin can see them on the request, so it's not currently possible for datasources to forward arbitrary headers, only the limited subset that Grafana injects (`Cookie` if cookie allow-list non-empty, `Authorization` and `X-Id-Token` when `oauthPassThru: true`, `X-Grafana-Id`, `X-Grafana-User` if `forwardGrafanaHeaders: true`, and Grafana tracing/dashboard headers). As Grafana _does_ support allow-lists for cookie forwarding, a cookie injected by a reverse-proxy in front of grafana can be used to propagate state through datasources.

Fixing this issue in the plugin will be required to allow the clickhouse plugin to properly support forwarding of client HTTP headers when Grafana upstream gains that ability in a fix to [grafana/grafana#87364](https://github.com/grafana/grafana/issues/87364). With both the plugin fix and a fix to the upstream issue, header forwarding will then work correctly.

It's also important to fix bearer token forwarding per this PR. That allows proxies between Grafana and Clickhouse to consume user identity, group/role membership and embedded custom claims in the OAuth `authorization_token` forwarded by Grafana when `oauthPassThru: true`. The proxy can then use that to modify the Clickhouse request to (e.g.) enforce row-security constraints or specific user/role bindings. (Unfortunately Clickhouse's native JWT auth validation does not support deriving settings from a token, but a proxy in front can do it). The clickhouse plugin does have JWT-forwarding support but it's focused on Clickhouse as a JWT consumer and doesn't match the behaviour of other Grafana plugins' OAuth forwarding.

**Environment**:

- Grafana version: 11.2.0
- Plugin version: 4.20.0
- OS Grafana is installed on: Alpine (Docker), Ubuntu 26.04 host
- User OS & Browser: N/A
- Others:

ringerc:
## Type of Change

- [x] 🐛 Bug Fix

### What is the bug?

Fixes: https://github.com/grafana/clickhouse-datasource/issues/2139 (cookie forwarding part only).

On the HTTP protocol, the plugin doesn't forward two headers that Grafana already makes available to it:

- `Cookie` (subset): populated from the data source's `keepCookies` list by Grafana's cookies middleware. Not forwarded at all, even with `keepCookies` set. Fixed by this PR.
- `Authorization` and `X-Id-Token`, populated when `oauthPassThru: true`. These do reach the plugin, but `Authorization` is diverted into ClickHouse's own JWT auth instead of forwarded as a plain header (see "OAuth forwarding" below). This PR doesn't change that part.

The cause is that `sqlds.DriverSettings.ForwardHeaders` was only set when `forwardGrafanaHeaders` or `oauthPassThru` was on. If a datasource just only set `keepCookies` no headers could be were forwarded.

Additionally, `Connect` pings ClickHouse unconditionally, including on the header-less call sqlds makes at data source startup. If a proxy in front of ClickHouse requires the cookie on every request it will reject these pings, so the data source never starts even once the first bug is fixed. This PR adds a new setting, `skipConnectionPings`, that an operator can turn on when they know their data source requires user header context (such as cookies) to permit a request to pass.

Note that the equivalent Prometheus data source forwards its cookie subset without any extra configuration.

#### Cookie forwarding

See https://github.com/grafana/clickhouse-datasource/issues/2139 for a self-contained test case. Manual steps:

Put a proxy in front of ClickHouse that requires a cookie named `my_auth_cookie` and returns 403 without it. Create a ClickHouse user `grafana_ds` with no other auth. Define a data source:

```yaml
jsonData:
  protocol: http
  host: <the proxy above>
  port: 8123
  username: grafana_ds
  keepCookies:
    - my_auth_cookie
  skipConnectionPings: true # the proxy would otherwise reject the header-less bootstrap ping
```

Grafana anonymous auth is fine for this test:

```
[auth.anonymous]
enabled = true
org_name = Main Org.
org_role = Viewer
```

Load a dashboard using this data source with a client request carrying `Cookie: my_auth_cookie=magicpermit`.

**Before this PR**: the data source fails to start, or every query fails with `[HTTP 403] "missing or invalid cookie"`.

**After this PR**: the cookie reaches ClickHouse, matching Prometheus data source behaviour.

#### OAuth forwarding

Not changed by this PR. See "OAuth forwarding" below.

### Related Issues

Fixes #2139 (cookie forwarding only)

Issue #2139 covers two things: (a) `Cookie` entries under `keepCookies` not forwarded, and (b) `Authorization`/`X-Id-Token` not usable the way Prometheus uses them under `oauthPassThru: true`. This PR resolves (a) only. Part (b) is a separate, pre-existing difference from Prometheus (see "OAuth forwarding" below) that this PR doesn't touch. It probably makes sense to split off #2139 into a separate bug for supporting prometheus-datasource-like OAuth forwarding that proxies can consume.

---

## Screenshots / Videos

Not applicable, this is a back-end fix. `skipConnectionPings` is not exposed on the UI, it's a provisioning-only setting since it's for fairly specialist use cases where auth-filtering reverse proxies are in use.

Tested with a real ClickHouse instance behind a header-capturing proxy, comparing the commit before this fix to this fix, for a `keepCookies`-only data source:

| case | before | after |
|---|---|---|
| cookie forwarding | fails, cookie not forwarded | passes |
| no forwarding configured | passes (unaffected) | passes |
| `forwardGrafanaHeaders` alone | passes (already worked) | passes |

---

## Please check that:

### Grafana limitations

Due to https://github.com/grafana/grafana/issues/87364 Grafana does not support forwarding an arbitrary subset of HTTP headers from client through to a datasource plugin.

If this limitations is lifted (e.g. by https://github.com/grafana/grafana/pull/131747 and https://github.com/grafana/grafana/pull/131748) then this PR will also permit the clickhouse datasource to forward the permitted headers, in much the same way `keepCookies` currently works. In the mean time it fixes `keepCookies`.

### Changes in this PR

1. `Settings()` now sets `sqlds.ForwardHeaders` to true on `protocol: http`, regardless of `forwardGrafanaHeaders` or `oauthPassThru`. The Native protocol is unaffected: clickhouse-go never reads `Options.HttpHeaders` on that transport, so there's nothing to forward there either way.

2. `Connect` skips its own ping when called with no message (sqlds's bootstrap call) if the new `skipConnectionPings` setting is on. Nothing else affects this: `forwardGrafanaHeaders` and `oauthPassThru` alone still ping as before. A `keepCookies`-only data source can't be told apart from a plain one from inside the plugin (`keepCookies` isn't visible to it), so the operator has to opt in explicitly when their proxy needs it. "Save & Test" is unaffected either way, since it pings the cached bootstrap connection through a different code path that never re-enters `Connect`.

3. Header extraction gains a restricted mode: when neither `forwardGrafanaHeaders` nor `oauthPassThru` is set, only `Cookie`, `Authorization`, and `X-Id-Token` are forwarded. Forwarding arbitrary headers still requires `forwardGrafanaHeaders` so there's no change to the actual headers forwarded when `keepCookies` isn't involved.

### Effect

- `keepCookies` alone on `protocol: http` now forwards the named cookies on every query, matching Prometheus data source behaviour.
- `oauthPassThru: true` alone still forwards `Authorization` (into JWT auth, see below) and `X-Id-Token` (as a plain header), unchanged.
- `forwardGrafanaHeaders: true` still forwards arbitrary headers, unchanged.
- Native protocol is unchanged.
- If no forwarding configured no headers are forwarded and ping still runs.

### Backward compatibility

- No behaviour changes unless `keepCookies` is set on the plugin. This used to be silently ignored; with this fix, the listed cookies will be passed through.
- `skipConnectionPings` defaults to off, so it has no BC effect.

### OAuth forwarding (not addressed by this PR)

- **TLS is required for `oauthPassThru`.** The plugin rejects `oauthPassThru: true` without a secure connection, and rejects it combined with skipping TLS verification. Prometheus has no equivalent requirement. A separate PR, https://github.com/grafana/clickhouse-datasource/pull/2151, addresses this.
- **`Authorization` handling differs from Prometheus even with TLS.** Prometheus forwards `Authorization` and `X-Id-Token` as plain headers. This plugin instead takes the bearer token out of `Authorization` and passes it to ClickHouse's own JWT auth, so a proxy expecting a plain `Authorization: Bearer` header the way it would see from Prometheus won't see one here. `X-Id-Token` is unaffected and passes through as a plain header. This is existing behavior, unchanged by this PR. This breaks the use of a proxy between the datasource and clickhouse for extracting custom claims from a JWT to use to stamp a HTTP request with extra parameters (e.g. for automatic row-security setting stamping or dynamic clickhouse role assignment) but it's not a new issue. I didn't want to tackle it in the same scope as this PR.

### DIFF
--- a/docs/sources/configure.md
+++ b/docs/sources/configure.md
@@ -283,6 +283,12 @@ With header forwarding enabled, connections are keyed by the forwarded header se
 
 To forward headers other than the Grafana-set ones (for example, bearer tokens or tenant identifiers), add them as **Custom HTTP headers** in the same **Optional HTTP settings** panel. Custom headers are sent on every query regardless of the **Forward Grafana HTTP headers** toggle.
 
+### Skip connection pings behind a fail-closed proxy
+
+On every query, and once at data source creation, the plugin pings ClickHouse to catch connection errors early. That initial ping carries no forwarded headers.
+
+If ClickHouse sits behind a proxy that requires a forwarded header on every request and rejects requests that lack it, that ping fails even though real per-user queries would succeed. For example, proxy might require the presence of a cookie retained via Grafana's own **Forward cookies** data source option. Set `skipConnectionPings: true` (provisioning only; there is no UI toggle) to disable pings in this case. Real queries are unaffected: they always carry the forwarded headers, so ClickHouse connection errors are still caught on the first query.
+
 ## Forward OAuth Identity
 
 {{< admonition type="note" >}}
@@ -306,6 +312,7 @@ To enable it, turn on **Forward OAuth Identity** in the **Database credentials**
   Enabling **Allow service account fallback** lets alert rules and other backend queries fall back to the configured username and password. Those queries then authenticate as the shared service account and are **not** subject to the per-user [row policies](https://clickhouse.com/docs/en/operations/access-rights/#row-policies), quotas, or query-log attribution that OAuth pass-through enforces for interactive queries. As a result, an alert may read rows a given dashboard user could not see interactively. Scope the configured service account to the least privilege your alert queries require. The plugin emits a backend warning log each time this fallback is exercised.
   {{< /admonition >}}
 - **Connections are keyed per user.** Enabling JWT authentication automatically turns on header forwarding, so each Grafana user opens a separate ClickHouse connection. See [Connection pool implications](#connection-pool-implications) for sizing guidance.
+- **Uses ClickHouse's native JWT authentication, not a raw header pass-through.** The plugin hands the forwarded OAuth token to the ClickHouse client library's JWT auth mechanism, which ClickHouse [authenticates natively as a JWT](https://clickhouse.com/docs/en/operations/external-authenticators/jwt). Unlike other data sources (for example, Prometheus), it does **not** forward the unmodified `Authorization` header as-is over HTTP.
 
 ## Verify the connection
 
@@ -357,6 +364,7 @@ datasources:
       # forwardGrafanaHeaders: <bool>
       # oauthPassThru: <bool>  # forward the user's OAuth token as a JWT (ClickHouse Cloud only); requires secure: true
       # oauthPassThruAllowFallback: <bool>  # allow alerts/backend queries to fall back to username and password
+      # skipConnectionPings: <bool>  # see "Forward Grafana HTTP headers" for when this is needed
       # path: <string>  # HTTP URL path (HTTP protocol only)
       # httpHeaders:     # HTTP protocol only
       #   - name: X-Example-Header
--- a/pkg/plugin/driver.go
+++ b/pkg/plugin/driver.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"net/textproto"
 	"regexp"
 	"strconv"
 	"strings"
@@ -249,7 +250,14 @@ func buildClickHouseOptions(ctx context.Context, settings Settings, message json
 		customSettings["limit"] = settings.RowLimit
 	}
 
-	httpHeaders, err := extractForwardedHeadersFromMessage(message)
+	// When the operator has explicitly opted into arbitrary header
+	// forwarding (forwardGrafanaHeaders) or OAuth pass-through, we surface
+	// everything sqlds marshalled into the message. Otherwise we restrict
+	// the extractor to the SDK's unprefixed whitelist so that Cookie /
+	// Authorization / X-Id-Token can flow (matching Prometheus) without
+	// broadening the arbitrary-http_-headers pass-through contract.
+	restrictToSDKWhitelist := !settings.ForwardGrafanaHeaders && !settings.OAuthPassThru
+	httpHeaders, err := extractForwardedHeadersFromMessage(message, restrictToSDKWhitelist)
 	if err != nil {
 		return nil, err
 	}
@@ -372,12 +380,30 @@ func (h *Clickhouse) Connect(
 	// `sqlds` normally calls `db.PingContext()` to check if the connection is alive,
 	// however, as ClickHouse returns its own non-standard `Exception` type, we need
 	// to handle it here so that we can categorize and surface the error correctly.
-	if err := db.PingContext(ctx); err != nil {
-		if ctx.Err() != nil {
-			return nil, fmt.Errorf("the operation was cancelled during execution: %w", ctx.Err())
-		}
+	//
+	// sqlds calls Connect(ctx, settings, nil) at bootstrap (NewConnector) with no
+	// per-request headers, and again from GetConnectionFromQuery before dispatching
+	// to the per-args Connect that actually carries the user's Cookie / Authorization
+	// / X-Id-Token. For most data sources that nil-message bootstrap ping is exactly
+	// what we want: it catches a bad host/port/credentials eagerly, at data source
+	// creation time. But the result would be misleading if the backend or a proxy in
+	// front of it requires a header or cookie to be forwarded from the client HTTP
+	// context and fails-closed if it's absent; that's why we have the
+	// `skipConnectionPings` option. We can't just add the client http context to the
+	// ping as one may not exist at the time a ping is called for.
+	//
+	// The per-args Connect (message != nil) still pings here. That ping carries
+	// the forwarded headers so genuine wire-level errors on real queries are still
+	// reported.
+	shouldPing := message != nil || !settings.SkipConnectionPings
+	if shouldPing {
+		if err := db.PingContext(ctx); err != nil {
+			if ctx.Err() != nil {
+				return nil, fmt.Errorf("the operation was cancelled during execution: %w", ctx.Err())
+			}
 
-		return nil, wrapCategorizedConnectionError(err)
+			return nil, wrapCategorizedConnectionError(err)
+		}
 	}
 
 	// Honor the (nil-resource-on-error) contract so callers can rely on
@@ -472,7 +498,19 @@ func (h *Clickhouse) Settings(ctx context.Context, config backend.DataSourceInst
 		FillMode: &data.FillMissing{
 			Mode: data.FillModeNull,
 		},
-		ForwardHeaders:  settings.ForwardGrafanaHeaders || settings.OAuthPassThru,
+		// On the HTTP protocol we always route through the sqlds
+		// header-forwarding machinery. This enables the  subset of cookies
+		// listed in `keepCookies` to be propagated as a `Cookie` header
+		// on the outbound HTTP connection.
+		//
+		// The extractor (see extractForwardedHeadersFromMessage)
+		// applies an allow-list filter when forwardGrafanaHeaders is
+		// off so that arbitrary http_-prefixed headers still require
+		// the existing opt-in.
+		//
+		// Useless for the native protocol because clickhouse-go drops
+		// headers anyway.
+		ForwardHeaders:  settings.ForwardGrafanaHeaders || settings.OAuthPassThru || settings.Protocol == "http",
 		RowCapacityHint: settings.RowCapacityHint,
 	}
 }
@@ -670,7 +708,16 @@ func convertFieldToString(field *data.Field) (*data.Field, error) {
 	return newField, nil
 }
 
-func extractForwardedHeadersFromMessage(message json.RawMessage) (map[string]string, error) {
+// extractForwardedHeadersFromMessage decodes the grafana-http-headers blob
+// that sqlds writes into the query message's ConnectionArgs. When
+// restrictToSDKWhitelist is true, only the three unprefixed keys the
+// grafana-plugin-sdk-go http-header helper exposes to backend plugins —
+// Cookie, Authorization, and X-Id-Token — pass through, matching the
+// header-forwarding surface of the Grafana HTTP client (used by the
+// Prometheus plugin). This lets Cookie / OAuth flow on the HTTP protocol
+// without also broadening the pass-through to arbitrary http_-prefixed
+// headers, which remains gated behind forwardGrafanaHeaders.
+func extractForwardedHeadersFromMessage(message json.RawMessage, restrictToSDKWhitelist bool) (map[string]string, error) {
 	// An example of the message we're trying to parse:
 	// {
 	//   "grafana-http-headers": {
@@ -697,6 +744,15 @@ func extractForwardedHeadersFromMessage(message json.RawMessage) (map[string]str
 		}
 
 		for k, v := range fwdHeaders {
+			if restrictToSDKWhitelist {
+				canonical := textproto.CanonicalMIMEHeaderKey(k)
+				if canonical != backend.CookiesHeaderName &&
+					canonical != backend.OAuthIdentityTokenHeaderName &&
+					canonical != backend.OAuthIdentityIDTokenHeaderName {
+					continue
+				}
+			}
+
 			anyHeadersArr, ok := v.([]interface{})
 			if !ok {
 				return nil, fmt.Errorf("couldn't parse header %s as an array", k)
--- a/pkg/plugin/driver_integration_test.go
+++ b/pkg/plugin/driver_integration_test.go
@@ -1387,6 +1387,71 @@ func TestHTTPConnectWithHeaders(t *testing.T) {
 
 		assert.Equal(t, nil, err)
 	})
+
+	// Regression tests for Cookie and other HTTP header forwarding from
+	// client requests.
+	// The ClickHouse plugin used to only forward headers if forwardGrafanaHeaders
+	// or oauthPassThru were enabled, so an operator who configured keepCookies
+	// but not that toggle would see no cookies reach ClickHouse. On protocol: http
+	// the plugin now forwards the SDK's unprefixed whitelist unconditionally so
+	// keepCookies works out of the box.
+	// This aligns it with the Prometheus datasource, which routes queries through
+	// grafana-plugin-sdk-go's HTTP client to get cookie-forwarding automatically.
+	t.Run("should forward the Cookie header out of the box on http protocol (parity with Prometheus keepCookies)", func(t *testing.T) {
+		settings := backend.DataSourceInstanceSettings{JSONData: []byte(fmt.Sprintf(`{"server": "localhost", "port": %s, "username": "%s", "protocol": "http"}`, proxyPort, username)), DecryptedSecureJSONData: secure}
+		// Permissive handler during NewDatasource's ping. We only care
+		// about headers on the actual QueryData call below.
+		proxy.NonproxyHandler = proxyHandlerGenerator(t, map[string]string{})
+		dsInstance, err := NewDatasource(ctx, settings)
+		assert.Equal(t, nil, err)
+
+		ds, ok := dsInstance.(backend.QueryDataHandler)
+		require.True(t, ok, "instance must implement backend.QueryDataHandler")
+
+		cookieReq := &backend.QueryDataRequest{
+			PluginContext: backend.PluginContext{DataSourceInstanceSettings: &settings},
+			Headers: map[string]string{
+				// Cookie is one of the three unprefixed headers the SDK
+				// forwards; this is what Grafana's CookiesMiddleware emits
+				// after keepCookies filtering.
+				"Cookie": "x_authorized_project_ids=proj-a,proj-b",
+			},
+			Queries: []backend.DataQuery{{RefID: "A", JSON: []byte(`{"rawSql": "SELECT 1"}`)}},
+		}
+
+		proxy.NonproxyHandler = proxyHandlerGenerator(t, map[string]string{"Cookie": "x_authorized_project_ids=proj-a,proj-b"})
+		_, err = ds.QueryData(ctx, cookieReq)
+		assert.Equal(t, nil, err)
+	})
+
+	t.Run("should not forward arbitrary http_-prefixed headers when forwardGrafanaHeaders is off, even with a Cookie present", func(t *testing.T) {
+		// Confirms the parity fix does not broaden the existing
+		// arbitrary-header opt-in contract: X-Test must only be emitted when
+		// forwardGrafanaHeaders is true even on the HTTP protocol.
+		settings := backend.DataSourceInstanceSettings{JSONData: []byte(fmt.Sprintf(`{"server": "localhost", "port": %s, "username": "%s", "protocol": "http"}`, proxyPort, username)), DecryptedSecureJSONData: secure}
+		proxy.NonproxyHandler = proxyHandlerGenerator(t, map[string]string{})
+		dsInstance, err := NewDatasource(ctx, settings)
+		assert.Equal(t, nil, err)
+
+		ds, ok := dsInstance.(backend.QueryDataHandler)
+		require.True(t, ok, "instance must implement backend.QueryDataHandler")
+
+		mixedReq := &backend.QueryDataRequest{
+			PluginContext: backend.PluginContext{DataSourceInstanceSettings: &settings},
+			Headers: map[string]string{
+				"Cookie":      "x_authorized_project_ids=proj-a",
+				"http_X-Test": "Hello World!",
+			},
+			Queries: []backend.DataQuery{{RefID: "A", JSON: []byte(`{"rawSql": "SELECT 1"}`)}},
+		}
+
+		proxy.NonproxyHandler = proxyHandlerGenerator(t, map[string]string{
+			"Cookie": "x_authorized_project_ids=proj-a",
+			"X-Test": "",
+		})
+		_, err = ds.QueryData(ctx, mixedReq)
+		assert.Equal(t, nil, err)
+	})
 }
 
 // TestQueryDataMacroExpansion exercises the macropro-driven macro pipeline end to
--- a/pkg/plugin/driver_test.go
+++ b/pkg/plugin/driver_test.go
@@ -534,26 +534,95 @@ func TestSettingsForwardHeadersWithJWT(t *testing.T) {
 		ds := h.Settings(t.Context(), config)
 		assert.False(t, ds.ForwardHeaders)
 	})
+
+	t.Run("ForwardHeaders is true on the http protocol so keepCookies flows without forwardGrafanaHeaders", func(t *testing.T) {
+		// Cookie parity with the Prometheus data source: on protocol: http
+		// we route through the sqlds header-forwarding machinery so that
+		// Cookie / Authorization / X-Id-Token (populated by Grafana's
+		// CookiesMiddleware from keepCookies, or by oauthPassThru) reach
+		// the ClickHouse HTTP endpoint out of the box.
+		// buildClickHouseOptions still restricts what actually flows to
+		// the SDK's unprefixed whitelist so arbitrary http_-headers stay
+		// opt-in behind forwardGrafanaHeaders.
+		config := backend.DataSourceInstanceSettings{
+			JSONData:                []byte(`{"host": "test", "port": 443, "protocol": "http"}`),
+			DecryptedSecureJSONData: map[string]string{},
+		}
+		ds := h.Settings(t.Context(), config)
+		assert.True(t, ds.ForwardHeaders)
+	})
+
+	t.Run("ForwardHeaders remains false on the native protocol when no opt-in is set", func(t *testing.T) {
+		// clickhouse-go silently drops opts.HttpHeaders on Native, so we
+		// leave ForwardHeaders off to avoid per-user connection cache
+		// pressure with no wire-level benefit.
+		config := backend.DataSourceInstanceSettings{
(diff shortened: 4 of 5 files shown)


pkg/plugin/driver.go
```diff
@@ -472,7 +498,19 @@ func (h *Clickhouse) Settings(ctx context.Context, config backend.DataSourceInst
...
+		// on the outbound HTTP connection.
+		//
+		// The extractor (see extractForwardedHeadersFromMessage)
+		// applies an allow-list filter when forwardGrafanaHeaders is
+		// off so that arbitrary http_-prefixed headers still require
+		// the existing opt-in.
+		//
+		// Useless for the native protocol because clickhouse-go drops
+		// headers anyway.
+		ForwardHeaders:  settings.ForwardGrafanaHeaders || settings.OAuthPassThru || settings.Protocol == "http",
```
adamyeats: With this on for every HTTP data source, sqlds opens, pings and caches a separate `*sql.DB` per distinct forwarded header set. Grafana puts per-panel and per-dashboard headers in that set and the cache never evicts, so consider keeping this behind an explicit opt-in.

pkg/plugin/driver.go
```diff
@@ -249,7 +250,14 @@ func buildClickHouseOptions(ctx context.Context, settings Settings, message json
...
 	}
 
-	httpHeaders, err := extractForwardedHeadersFromMessage(message)
+	// When the operator has explicitly opted into arbitrary header
+	// forwarding (forwardGrafanaHeaders) or OAuth pass-through, we surface
+	// everything sqlds marshalled into the message. Otherwise we restrict
+	// the extractor to the SDK's unprefixed whitelist so that Cookie /
+	// Authorization / X-Id-Token can flow (matching Prometheus) without
+	// broadening the arbitrary-http_-headers pass-through contract.
+	restrictToSDKWhitelist := !settings.ForwardGrafanaHeaders && !settings.OAuthPassThru
```
adamyeats, line 265: sqlds hashes the whole header set into the cache key before it calls Connect, so this filter bounds the wire headers but not the number of pools. The plugin cannot filter earlier either, because sqlds reads the headers before MutateQueryData, so the comment above overstates what this protects.

pkg/plugin/settings.go
```diff
@@ -52,8 +52,16 @@ type Settings struct {
 	// Health checks and schema introspection always fall back regardless of
 	// this setting, since no user token is ever available for them.
 	OAuthPassThruAllowFallback bool `json:"oauthPassThruAllowFallback,omitempty"`
-	CustomSettings        []CustomSetting   `json:"customSettings"`
-	ProxyOptions          *proxy.Options
+	// SkipConnectionPings disables the bootstrap/health-check ping that
```
adamyeats: Save & test does not go through this ping. sqlds pings the bootstrap pool itself in CheckHealth, so it is worth trimming this to the bootstrap ping only.

docs/sources/configure.md
```diff
@@ -283,6 +283,12 @@ With header forwarding enabled, connections are keyed by the forwarded header se
 
 To forward headers other than the Grafana-set ones (for example, bearer tokens or tenant identifiers), add them as **Custom HTTP headers** in the same **Optional HTTP settings** panel. Custom headers are sent on every query regardless of the **Forward Grafana HTTP headers** toggle.
 
+### Skip connection pings behind a fail-closed proxy
+
+On every query, and once at data source creation, the plugin pings ClickHouse to catch connection errors early. That initial ping carries no forwarded headers.
```
adamyeats: The plugin pings once at creation and once per new forwarded header set, not on every query. Consider rewording so operators do not expect a ping per query.

docs/sources/configure.md
```diff
@@ -283,6 +283,12 @@ With header forwarding enabled, connections are keyed by the forwarded header se
...
+
+If ClickHouse sits behind a proxy that requires a forwarded header on every request and rejects requests that lack it, that ping fails even though real per-user queries would succeed. For example, proxy might require the presence of a cookie retained via Grafana's own **Forward cookies** data source option. Set `skipConnectionPings: true` (provisioning only; there is no UI toggle) to disable pings in this case. Real queries are unaffected: they always carry the forwarded headers, so ClickHouse connection errors are still caught on the first query.
```
adamyeats: Save & test still pings through sqlds and ignores this setting, so behind a fail-closed proxy it keeps failing after provisioning succeeds. It is worth saying that here so operators do not read the failed test as a broken data source.

adamyeats: nit: `proxy might require` needs an article: `a proxy might require`.

pkg/plugin/driver.go
```diff
@@ -372,12 +380,30 @@ func (h *Clickhouse) Connect(
 	// `sqlds` normally calls `db.PingContext()` to check if the connection is alive,
 	// however, as ClickHouse returns its own non-standard `Exception` type, we need
 	// to handle it here so that we can categorize and surface the error correctly.
-	if err := db.PingContext(ctx); err != nil {
-		if ctx.Err() != nil {
-			return nil, fmt.Errorf("the operation was cancelled during execution: %w", ctx.Err())
-		}
+	//
+	// sqlds calls Connect(ctx, settings, nil) at bootstrap (NewConnector) with no
+	// per-request headers, and again from GetConnectionFromQuery before dispatching
```
adamyeats: GetConnectionFromQuery never calls Connect with a nil message. It loads the bootstrap entry from the cache and calls the per-args Connect once on a miss, so consider trimming this to the NewConnector call.

pkg/plugin/driver.go
```diff
@@ -670,7 +708,16 @@ func convertFieldToString(field *data.Field) (*data.Field, error) {
 	return newField, nil
 }
 
-func extractForwardedHeadersFromMessage(message json.RawMessage) (map[string]string, error) {
+// extractForwardedHeadersFromMessage decodes the grafana-http-headers blob
+// that sqlds writes into the query message's ConnectionArgs. When
+// restrictToSDKWhitelist is true, only the three unprefixed keys the
+// grafana-plugin-sdk-go http-header helper exposes to backend plugins —
+// Cookie, Authorization, and X-Id-Token — pass through, matching the
```
adamyeats: Grafana only attaches Authorization and X-Id-Token when oauthPassThru is on, and that flag turns restricted mode off. In practice only Cookie can match here, so consider saying so to save the next reader a search for the OAuth path.

docs/sources/configure.md
```diff
@@ -283,6 +283,12 @@ With header forwarding enabled, connections are keyed by the forwarded header se
 
 To forward headers other than the Grafana-set ones (for example, bearer tokens or tenant identifiers), add them as **Custom HTTP headers** in the same **Optional HTTP settings** panel. Custom headers are sent on every query regardless of the **Forward Grafana HTTP headers** toggle.
 
+### Skip connection pings behind a fail-closed proxy
```
adamyeats, line 293: After this change Cookie reaches ClickHouse on the HTTP protocol with the toggle off, and the header table above (line 265) does not mention it. Consider documenting the new default there, since it is the fix the PR ships.

docs/sources/configure.md
```diff
@@ -283,6 +283,12 @@ With header forwarding enabled, connections are keyed by the forwarded header se
 
 To forward headers other than the Grafana-set ones (for example, bearer tokens or tenant identifiers), add them as **Custom HTTP headers** in the same **Optional HTTP settings** panel. Custom headers are sent on every query regardless of the **Forward Grafana HTTP headers** toggle.
```
adamyeats, line 291: If the blanket ForwardHeaders change stays, the Connection pool implications paragraph above (line 280) needs to cover every HTTP data source and say the key is the whole forwarded header set, not the user.

pkg/plugin/settings.go
```diff
@@ -249,6 +257,17 @@ func LoadSettings(ctx context.Context, config backend.DataSourceInstanceSettings
 		}
 	}
 
+	if jsonData["skipConnectionPings"] != nil {
```
adamyeats, line 283: nit: the other boolean flags each have a LoadSettings test for the bool, string and default cases in settings_test.go. Consider adding the same for skipConnectionPings, in particular the string branch.

pkg/plugin/driver_test.go
```diff
@@ -832,3 +901,35 @@ func TestMissingTableMacroIsDownstream(t *testing.T) {
...
+// or header is missing. This relies solely on Grafana's own keepCookies
+// forwarding. The operator must opt in explicitly with skipConnectionPings
+// to allow this as there's no way to auto-detect it.
+func TestConnectSkipsBootstrapPingWithSkipConnectionPingsOptIn(t *testing.T) {
+	h := &Clickhouse{}
+	settings := backend.DataSourceInstanceSettings{
+		JSONData:                []byte(`{"host": "127.0.0.1", "port": 1, "protocol": "http", "skipConnectionPings": true, "dialTimeout": "1", "queryTimeout": "1"}`),
+		DecryptedSecureJSONData: map[string]string{},
+	}
+	db, err := h.Connect(t.Context(), settings, nil)
```
adamyeats, line 983: nit: this covers the nil-message half of the gate only. A second call with a non-nil message against 127.0.0.1:1 that requires an error protects the `message != nil ||` term.

pkg/plugin/driver_test.go
```diff
@@ -534,27 +534,96 @@ func TestSettingsForwardHeadersWithJWT(t *testing.T) {
...
+		assert.NoError(t, err)
+		assert.Equal(t, "x_authorized_project_ids=proj-a,proj-b", headers["Cookie"])
+	})
+
+	// Restricted mode (forwardGrafanaHeaders off, oauthPassThru off) is what
+	// buildClickHouseOptions passes to give Cookie / OAuth pass-through
+	// parity with the Grafana HTTP client (as used by the Prometheus
+	// datasource) without also broadening arbitrary http_-prefixed
+	// forwarding, which remains opt-in via forwardGrafanaHeaders.
+	t.Run("restricted mode passes only the SDK whitelist and drops everything else", func(t *testing.T) {
```
adamyeats, line 632: nit: consider a sibling case with oauthPassThru on that asserts a non-whitelisted header survives. The OAuthPassThru half of the condition at driver.go:259 has no test on the header outcome.

pkg/plugin/driver_integration_test.go
```diff
@@ -1387,6 +1387,71 @@ func TestHTTPConnectWithHeaders(t *testing.T) {
...
+				// forwards; this is what Grafana's CookiesMiddleware emits
+				// after keepCookies filtering.
+				"Cookie": "x_authorized_project_ids=proj-a,proj-b",
+			},
+			Queries: []backend.DataQuery{{RefID: "A", JSON: []byte(`{"rawSql": "SELECT 1"}`)}},
+		}
+
+		proxy.NonproxyHandler = proxyHandlerGenerator(t, map[string]string{"Cookie": "x_authorized_project_ids=proj-a,proj-b"})
+		_, err = ds.QueryData(ctx, cookieReq)
+		assert.Equal(t, nil, err)
```
adamyeats: nit: sqlds returns a nil error from QueryData unless DriverSettings.Errors is set, so this assertion cannot fail. Consider checking `resp.Responses["A"].Error` instead, since these two subtests are the only proof of the headline behaviour.

pkg/plugin/driver_test.go
```diff
@@ -534,27 +534,96 @@ func TestSettingsForwardHeadersWithJWT(t *testing.T) {
...
 	})
 
 	t.Run("returns empty map when message is nil", func(t *testing.T) {
-		headers, err := extractForwardedHeadersFromMessage(nil)
+		headers, err := extractForwardedHeadersFromMessage(nil, false)
 		assert.NoError(t, err)
 		assert.Empty(t, headers)
 	})
+
+	// Grafana's plugin SDK filters the headers plugins can see so most client
```
adamyeats: nit: the SDK does no filtering here. Grafana core middlewares build the header map, and the set is wider than these three (org, dashboard and panel headers too).

pkg/plugin/driver_test.go
```diff
@@ -832,3 +901,35 @@ func TestMissingTableMacroIsDownstream(t *testing.T) {
...
+	settings := backend.DataSourceInstanceSettings{
+		JSONData:                []byte(`{"host": "127.0.0.1", "port": 1, "protocol": "http", "forwardGrafanaHeaders": true, "dialTimeout": "1", "queryTimeout": "1"}`),
+		DecryptedSecureJSONData: map[string]string{},
+	}
+	_, err := h.Connect(t.Context(), settings, nil)
+	require.Error(t, err, "forwardGrafanaHeaders alone, without skipConnectionPings, must still probe the wire on Connect")
+}
+
+// TestConnectSkipsBootstrapPingWithSkipConnectionPingsOptIn covers the case
+// where the downstream is a a data source that fails-closed if a required cookie
```
adamyeats: nit: `a a data source` has a doubled article.

pkg/plugin/driver.go
```diff
@@ -472,7 +498,19 @@ func (h *Clickhouse) Settings(ctx context.Context, config backend.DataSourceInst
 		FillMode: &data.FillMissing{
 			Mode: data.FillModeNull,
 		},
-		ForwardHeaders:  settings.ForwardGrafanaHeaders || settings.OAuthPassThru,
+		// On the HTTP protocol we always route through the sqlds
+		// header-forwarding machinery. This enables the  subset of cookies
```
adamyeats: nit: double space in `the  subset`.

docs/sources/configure.md
```diff
@@ -357,6 +364,7 @@ datasources:
       # forwardGrafanaHeaders: <bool>
       # oauthPassThru: <bool>  # forward the user's OAuth token as a JWT (ClickHouse Cloud only); requires secure: true
       # oauthPassThruAllowFallback: <bool>  # allow alerts/backend queries to fall back to username and password
+      # skipConnectionPings: <bool>  # see "Forward Grafana HTTP headers" for when this is needed
```
adamyeats: nit: the explanation lives in the new `Skip connection pings behind a fail-closed proxy` subsection, so consider pointing there. The setting is independent of the forward-headers toggle.

adamyeats (changes requested):
nit: the title lint requires a lowercase subject and the body ticks Bug Fix, so `feat(headers): Fix …` fails the check and lands under Features in the changelog. Consider `fix(headers): forward keepCookies cookies on the http protocol`.

pkg/plugin/driver.go
ringerc: I don't understand this claim. The comment doesn't claim to limit the number of pools or cached datasources.

>  so the comment above overstates what this protects.

How? It's making a claim about what data is visible, not caching side-effects.

pkg/plugin/driver.go
ringerc: Valid. Unfortunate the caching can't be scoped to the subset of headers we care about. Will look into it.

pkg/plugin/settings.go
ringerc: Unclear what this means. The comment doesn't claim that save-and-test goes through this. Is this asking for a comment that explicitly says so, a param renaming, something else?

docs/sources/configure.md
ringerc: Thanks for the correction. That's probably a misunderstanding arising from the test harness. Will address.

docs/sources/configure.md
ringerc: I thought I'd addressed that. I'll look again, as I was seeking to have save-and-test return an unknown/indeterminate result (which it supports AFAICS) in this case.

pkg/plugin/driver.go
ringerc: This is deliberate - https://github.com/grafana/grafana/pull/131747 seeks to widen this to allow other headers. But I'll look at wording adjustment here.

docs/sources/configure.md
ringerc: Assuming "the toggle" means `skipConnectionPings: true` - I don't see why this section would mention that. It's only relevant for when a proxy is used to filter access to CH based on the http headers (including cookie).

If this claims that there's a defaults behaviour change, then that's a separate issue I need to look into. The comment is not very clear.

ringerc:
Thanks for the independent check run @adamyeats . I'll go over those. Some of the comments are difficult to understand though.

ringerc:
I've rebased and pushed fixes with re-signed commits.

pkg/plugin/driver.go
ringerc: (Claude, on behalf of @ringerc) Fixed: `ForwardHeaders` on the http protocol now only turns on when Grafana's `keepCookies` is configured for the data source, or the new `forceCookieForwarding` provisioning-only override is set — not unconditionally for every http data source. This avoids growing sqlds's non-evicting pool cache for data sources that never asked for cookie forwarding.

pkg/plugin/driver.go
ringerc: (Claude, on behalf of @ringerc) Fixed: reworded the comment to note this filter only bounds what reaches the wire, not the sqlds connection-pool cache key, since sqlds hashes the full forwarded header set into that key before `Connect` is ever called.

pkg/plugin/settings.go
ringerc: (Claude, on behalf of @ringerc) Fixed: the `SkipConnectionPings` doc comment now states explicitly that it does not affect Save & test, since sqlds pings the bootstrap pool directly in `CheckHealth`.

docs/sources/configure.md
ringerc: (Claude, on behalf of @ringerc) Fixed: reworded to "once at data source creation, and again on the first query for each new forwarded header set" instead of implying a ping on every query.

docs/sources/configure.md
ringerc: (Claude, on behalf of @ringerc) Fixed: added an admonition noting Save & test still fails behind a fail-closed proxy even with `skipConnectionPings: true`, and that this doesn't indicate a broken data source.

ringerc: (Claude, on behalf of @ringerc) Fixed.

pkg/plugin/driver.go
ringerc: (Claude, on behalf of @ringerc) Fixed: trimmed the comment so it says `GetConnectionFromQuery` loads the bootstrap entry from cache and calls the per-args `Connect` once on a miss, rather than implying a nil-message call there.

pkg/plugin/driver.go
ringerc: (Claude, on behalf of @ringerc) Fixed: noted that Authorization/X-Id-Token only ever appear when `oauthPassThru` is set (which already selects unrestricted mode), so in practice restricted mode only ever sees Cookie flow through this path.

docs/sources/configure.md
ringerc: (Claude, on behalf of @ringerc) Fixed: added a note above the header table documenting that Cookie now forwards on the http protocol when `keepCookies` lists at least one cookie, regardless of the toggle, and mentioning the new `forceCookieForwarding` override.

docs/sources/configure.md
ringerc: (Claude, on behalf of @ringerc) Fixed: reworded to cover the whole forwarded header set (not just per-user), and explained why Cookie forwarding is gated behind `keepCookies`/`forceCookieForwarding` rather than unconditional for every http data source.

pkg/plugin/settings.go
ringerc: (Claude, on behalf of @ringerc) Added `TestLoadSettingsSkipConnectionPings` (bool/string/default cases), plus matching `LoadSettings` tests for the new `keepCookies` detection and `forceCookieForwarding`.

pkg/plugin/driver_test.go
ringerc: (Claude, on behalf of @ringerc) Added `TestConnectPingsOnPerArgsConnectRegardlessOfSkipOptIn`, covering the `message != nil` half of the gate with `skipConnectionPings` set.

pkg/plugin/driver_test.go
ringerc: (Claude, on behalf of @ringerc) Added a sibling case asserting a non-whitelisted header (`X-Grafana-Org-Id`) survives when `oauthPassThru` selects unrestricted mode.

pkg/plugin/driver_integration_test.go
ringerc: (Claude, on behalf of @ringerc) Fixed: both subtests now also assert `resp.Responses["A"].Error` is nil, in addition to the `QueryData` err check.

pkg/plugin/driver_test.go
ringerc: (Claude, on behalf of @ringerc) Fixed: reworded to say Grafana core middlewares (not the SDK) build the header set, and that it's wider than the three keys this test covers.

pkg/plugin/driver_test.go
ringerc: (Claude, on behalf of @ringerc) Fixed.

pkg/plugin/driver.go
ringerc: (Claude, on behalf of @ringerc) Fixed.

docs/sources/configure.md
ringerc: (Claude, on behalf of @ringerc) Fixed: the provisioning example now points to the "Skip connection pings behind a fail-closed proxy" section.

PR #2168, opened 2026-09-16
test: bind the proxy on an ephemeral port and inherit the Cloud expect timeout

codyconfer:
## Type of Change

- [ ] 🚀 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧹 Refactor / Chore

---
#### increase cloud run timeout for sql Autocomplete

_Playwright starts each test with a blank browser context, each test re-downloads the plugin bundle and Monaco assets, this process can last longer on a cloud run, so cloud runs get a higher timeout_

#### remove driver integration proxy

_Driver integration test started a mock proxy server in the background, then immediately tried to poll it causing a race between "server starting" and "server status".
The old fix just polled longer, which is still a gamble. This change set grabs the port first, then starts serving on it — so there's nothing to poll for.
This change set also ensures the server closes._

## Please check that:


tests/e2e/sqlAutocomplete.spec.ts
```diff
@@ -10,6 +10,11 @@ const CLOUD_DEFAULT_UID = 'clickhouse-native-ds-m';
 const LOCAL_DEFAULT_UID = 'clickhouse-e2e';
 const DATASOURCE_UID = process.env.DS_E2E_UID || (isCloudRun ? CLOUD_DEFAULT_UID : LOCAL_DEFAULT_UID);
 
+// Playwright starts each test with a blank browser context
+// each test re-downloads the plugin bundle and Monaco assets
+// this process can last longer on a cloud run
+const SUGGEST_WIDGET_TIMEOUT = isCloudRun ? 15_000 : 5_000;
```
adamyeats: `playwright.config.ts` already sets `expect.timeout` to 30s on Cloud for this same cause, but `waitFor` reads `actionTimeout` and never sees it. Consider `await expect(widget).toBeVisible()` instead, which picks up the tuned config values and removes this constant.

tests/e2e/sqlAutocomplete.spec.ts
```diff
@@ -42,7 +47,7 @@ async function focusEditorAndType(page: Page, text: string) {
 // a deterministic target across local fixture and Cloud cron runs.
 async function captureMacroLabels(page: Page): Promise<string[]> {
   const widget = page.locator('.monaco-editor .suggest-widget.visible');
-  await widget.waitFor({ timeout: 5000 });
+  await widget.waitFor({ timeout: SUGGEST_WIDGET_TIMEOUT });
```
adamyeats: This line conflicts with my #2123, which moves the spec to `tests/e2e/editor/` and sets the same wait to a flat 15s. Whichever lands second needs a rebase, so it might be worth agreeing on one value now.

adamyeats (changes requested):

tests/e2e/sqlAutocomplete.spec.ts
codyconfer: I switched it to this

adamyeats (approved):

Issue #2172, opened 2026-09-16
Ad-hoc filter follow-ups from #1936: CTE FROM, UNION branches, trailing comment


alex-fedotyev:
Follow-ups from #1936 (ad-hoc filter table detection). Each is a pre-existing behavior that #1936 surfaces or intentionally leaves for a focused change. None is a regression from #1936: each is a no-op today or the known multi-table-application limitation.

### 1. Resolve a FROM that references a CTE

`getTable` returns the CTE alias for a query that selects directly from a CTE:

```sql
WITH lookup AS (SELECT id FROM dim_services) SELECT * FROM lookup
```

resolves `lookup`. ClickHouse ignores an `additional_table_filters` entry keyed on a CTE, so the filter is a no-op (the same result as before #1936). Resolving the alias to the CTE body's physical table would let the filter apply.

### 2. Key every branch of a UNION

For `SELECT a FROM tbl_a UNION ALL SELECT a FROM tbl_b`, only the first branch's table is keyed, so half the result set obeys the filter. This is the same class as applying filters to every table in a join, which #1936 left out of scope.

### 3. Append the settings clause after a newline

`apply()` appends ` settings additional_table_filters={...}` on the same line. A query ending in a trailing line comment (`-- ...`) then swallows the appended clause, so the filter is silently inert. Appending after a newline avoids it.

Separately, the `changelog` label instruction in `pull_request_template.md` is stale since release-please was adopted (no workflow reads it), and worth removing.

Refs #1936.

alex-fedotyev:
Draft #2173 covers items 1 (resolve a CTE-referencing FROM) and 3 (append the settings clause on a new line). Item 2 (keying every UNION branch) remains open here, since it needs the same per-table column resolution as the join case.

alex-fedotyev:
Rescoped after #1936. Items 1 (a CTE-referencing FROM) and 3 (settings after a newline) are handled in #1936 directly, and a top-level UNION there now resolves to no table so a filter is never half-applied to one branch. What remains here is the enhancement:

- Key every branch of a UNION (per table, with column resolution), the same multi-table work as applying a filter across a JOIN.

Also open: the stale `changelog` label instruction in `pull_request_template.md`.

adamyeats:
#1936 has merged. Items 1 and 3 above landed in it, so the open scope is item 2 plus three leftovers from the final review.

[...]

PR #2173, opened 2026-09-16
fix(adhoc): resolve CTE references and append settings on a new line

alex-fedotyev:
## Summary

Follow-up to #1936 that addresses two of the deferred items in #2172.

- **CTE references.** `getTable()` now resolves a `FROM` that references a CTE to the CTE's underlying table, instead of the alias. ClickHouse ignores an `additional_table_filters` entry keyed on a CTE, so the filter was a no-op before. The parser records each WITH-clause alias and its subquery body, and resolution follows the alias to that body's table.
- **Trailing comments.** `AdHocFilter.apply()` appends the settings clause on a new line, so a query ending in a trailing line comment (`-- ...`) no longer swallows the clause and silently drops the filter.

The remaining #2172 item, keying every branch of a `UNION`, is left out here: it is the same multi-table-with-column-resolution work that #1936 scoped out of the join case.

> [!NOTE]
> Stacked on #1936. The diff includes that PR's commit until it merges, so this stays a draft; I will rebase it onto `main` once #1936 lands.

## Test plan

- [x] `npx jest` (full suite green)
- [x] `npx tsc --noEmit`
- [x] `npx eslint .`
- [x] `npx prettier --check`
- [x] `npm run spellcheck`

New tests: CTE resolution in `src/data/ast.test.ts` and `src/ch-parser/parser.test.ts` (alias to table, a CTE reference with a WHERE clause, the referenced CTE among several, and a table-less CTE), plus `apply()` cases for the trailing-comment newline and CTE resolution end to end.


alex-fedotyev:
Folding into #1936. The CTE-alias deferral turned out to introduce a regression on the tag-source path (a `$clickhouse_adhoc_query` CTE query pinned the shared target, so other panels came back unfiltered), so CTE resolution and the settings-after-newline fix now live in #1936 directly. The remaining UNION work stays in #2172. Closing.

PR #2177, opened 2026-09-17
fix(adhoc): follow-up review fixes for ad-hoc table detection

alex-fedotyev:
## Summary

Follow-ups from #1936, tracked in #2172. Each is small and independent.

- **Semicolon before a trailing comment.** `apply()` stripped only a final `;`, so `SELECT count() FROM t; -- note` with an active filter became `t; -- note` followed by the settings line, which ClickHouse rejects as `Multi-statements are not allowed` (it ran, unfiltered, before #1936). It now strips a `;` followed only by whitespace and comments; a `;` inside a literal (`splitByChar(';', col)`) is left intact.
- **Unqualified `format` / `prewhere`.** `RESERVED_FROM_KEYWORDS` refused both as the first FROM segment, so `SELECT * FROM format` returned no table. pgsql-ast-parser did not reserve these two, so they resolved before #1936; dropping them restores parity. `db.<keyword>` and quoting are unaffected.
- **`${` as a JOIN target.** `typingVariable` read `root.from`, so only the first FROM offered variable completions. The cursor now reads the nearest FROM/JOIN, so a `$`-prefixed JOIN target offers `${table}` / `${database}` too.
- **INTERSECT / EXCEPT.** A top-level UNION already resolves to no table; INTERSECT and EXCEPT now do the same, so a filter is never keyed to only the first branch. Per-branch keying across a set operation remains the open item in #2172.

Refs #2172

## Test plan

- [x] `npx jest` (full suite green)
- [x] `npx tsc --noEmit`
- [x] `npx eslint .`
- [x] `npx prettier --check`
- [x] `npm run spellcheck`
- [x] Ran a `main`-vs-branch differential harness over an enumerated corpus (`getTable` + `getSuggestions`); the only deltas are the four intended changes above, with nothing else altered.

New tests cover each fix: the trailing-`;`-and-comment strip, `format` / `prewhere` resolution, INTERSECT / EXCEPT resolving to no table, the set-operation flag, and variable completions for a JOIN target.


src/ch-parser/parser.ts
```diff
@@ -303,9 +295,9 @@ export function parseSelectQueryNode(parser: QueryNodeParser): SelectQueryNode |
       node.children!.push({ type: QueryNodeType.Default, token: parser.next(), clause: ClauseType.OrderBy });
     } else if (token.matchKeyword('LIMIT')) {
       node.children!.push({ type: QueryNodeType.Default, token, clause: ClauseType.Limit });
-    } else if (token.matchKeyword('UNION')) {
+    } else if (token.matchKeyword('UNION') || token.matchKeyword('INTERSECT') || token.matchKeyword('EXCEPT')) {
```
adamyeats, line 298: ClickHouse also uses `EXCEPT` as a column modifier before the FROM, as in `SELECT * EXCEPT (secret) FROM otel_logs` or `COLUMNS('^Span') EXCEPT (SpanId)`, and this branch flags those as a set operation, so `getTable` returns no table and the ad-hoc filter is silently skipped on queries that were filtered before this PR. spans.ts:448 already counts `EXCEPT` only after the FROM, so consider the same guard here, e.g. `token.matchKeyword('EXCEPT') && node.from !== undefined`.

src/data/adHocFilter.ts
```diff
@@ -102,10 +102,12 @@ export class AdHocFilter {
...
-    // Strip only a trailing semicolon before appending the settings clause. A
-    // bare replace(';', '') would delete the first semicolon anywhere, e.g.
-    // inside splitByChar(';', col).
-    sql = sql.replace(/;\s*$/, '');
+    // Strip a trailing statement terminator before appending the settings
+    // clause: a `;` followed only by whitespace and comments. Otherwise the
+    // clause lands after `;`, which ClickHouse rejects as multiple statements.
+    // The pattern is anchored to the end so a `;` inside a literal, e.g.
+    // splitByChar(';', col), is left intact.
+    sql = sql.replace(/;(?:\s|--[^\n]*|\/\*[\s\S]*?\*\/)*$/, '');
```
adamyeats, line 110: This alternation is ambiguous, so a `;` followed by a dash-ruler comment line and then more SQL backtracks exponentially, a 40-dash ruler takes 18 s in Node, and `apply()` runs on every query while a filter is active. Consider anchoring the line-comment branch to its line end, `--[^\n]*(?=\n|$)`, or locating the trailing `;` with the ch-parser lexer, which already emits Comment and Semicolon tokens.

adamyeats, line 110: The regex does not know about string literals or `#` comments, so `SELECT '; -- x' AS a FROM t` loses everything after the opening quote, and `FROM t; # note` keeps its `;` and still fails as multi-statements. The lexer route above covers both, since it already tokenises `#` and `#!` comments and quoted strings.

src/components/suggestions.ts
```diff
@@ -104,6 +104,20 @@ function getCursorInSelectQueryNode(root: SelectQueryNode, cursorPosition: numbe
...
   }
 
+  // When the cursor is in a FROM/JOIN position on a `$`-prefixed target, surface
+  // that target (the nearest FROM/JOIN before the cursor) so variable
+  // completions are offered for a JOIN target too, not just the first FROM. The
+  // clause guard keeps this from changing a later clause's column context.
+  if (cursorData.clause === ClauseType.From || cursorData.clause === ClauseType.Join) {
+    for (const node of root.children) {
+      const fromNode = node as FromQueryNode;
+      if (node.type === QueryNodeType.From && fromNode.table?.startsWith('$') && node.token.begin <= cursorPosition) {
```
adamyeats, line 114: `typingVariable` at line 200 treats `data.table || data.database` as the variable, so `FROM ${db}.` offers `${table}`, but this check reads only `fromNode.table`, so `JOIN ${db}.` still offers nothing. Consider mirroring line 200 here with `(fromNode.table || fromNode.database || '').startsWith('$')`.

src/components/suggestions.ts
```diff
@@ -104,6 +104,20 @@ function getCursorInSelectQueryNode(root: SelectQueryNode, cursorPosition: numbe
     }
   }
 
+  // When the cursor is in a FROM/JOIN position on a `$`-prefixed target, surface
+  // that target (the nearest FROM/JOIN before the cursor) so variable
+  // completions are offered for a JOIN target too, not just the first FROM. The
+  // clause guard keeps this from changing a later clause's column context.
+  if (cursorData.clause === ClauseType.From || cursorData.clause === ClauseType.Join) {
```
adamyeats, line 111: nit: the parser never assigns `ClauseType.Join`, since a JOIN token is consumed at parser.ts:215 as a From node with `clause: ClauseType.From` and the branch at parser.ts:286 is unreachable. The `Join` half of this guard is therefore inert, which is harmless, but the comment above reads as if the two clause types were distinct.

src/ch-parser/parser.ts
```diff
@@ -95,26 +95,18 @@ export interface SelectQueryNode extends QueryNode {
...
-  'PREWHERE',
-  'JOIN',
-  'FORMAT',
-  'UNION',
-]);
+// Clause-starting keywords are never a table name in the FROM position. Without
+// this guard `SELECT * FROM  WHERE ...` would take `WHERE` as the table while
+// the user is mid-edit, and no clause node would be created. FORMAT and PREWHERE
+// are intentionally absent: pgsql-ast-parser did not reserve them, so they
+// resolved as table names before, and neither can begin a statement.
```
adamyeats, line 108: nit: every word left in this set also cannot begin a statement, so that clause does not explain why these two are safe to drop. The reason that holds is that the parser has no clause branch for FORMAT or PREWHERE, so the only cost is a mid-edit `FROM  PREWHERE x` resolving `prewhere`, and the line 116 comment about backtick quoting is now false for these two.

src/ch-parser/parser.test.ts
```diff
@@ -111,9 +111,16 @@ describe('parseSelect', () => {
...
   });
 
-  it('flags a top-level UNION', () => {
-    expect(parseSelect('SELECT a FROM t1 UNION ALL SELECT a FROM t2')?.hasUnion).toBe(true);
-    expect(parseSelect('SELECT a FROM t1')?.hasUnion).toBeFalsy();
+  it('flags a top-level set operation (UNION / INTERSECT / EXCEPT)', () => {
+    expect(parseSelect('SELECT a FROM t1 UNION ALL SELECT a FROM t2')?.hasSetOperation).toBe(true);
+    expect(parseSelect('SELECT a FROM t1 INTERSECT SELECT a FROM t2')?.hasSetOperation).toBe(true);
+    expect(parseSelect('SELECT a FROM t1 EXCEPT SELECT a FROM t2')?.hasSetOperation).toBe(true);
+    expect(parseSelect('SELECT a FROM t1')?.hasSetOperation).toBeFalsy();
```
adamyeats, line 118: nit: same shape as on #1936, a `parseSelect` that returns null also satisfies `toBeFalsy()` through the `?.`, so this negative case cannot tell the flag being unset from the parser rejecting the input. Consider asserting `not.toBeNull()` on the node first, or reading it into a variable and asserting on that.

src/components/suggestions.test.ts
```diff
@@ -268,4 +268,36 @@ describe('Suggestions', () => {
     expect(labels).toContain('${table}');
     expect(labels).toContain('${database}');
   });
+
+  it('offers Grafana variables when a variable is typed as a JOIN target', async () => {
```
adamyeats, line 272: nit: consider a companion case such as `SELECT * FROM otel_logs JOIN ${t} ON a = b WHERE ` asserting that column suggestions still come from `otel_logs`. The clause guard in the new block is what keeps a later clause's context intact, and nothing exercises it.

src/data/adHocFilter.test.ts
```diff
@@ -272,6 +272,17 @@ describe('AdHocManager', () => {
     expect(val).toContain('-- note\nsettings additional_table_filters=');
   });
 
+  it('strips a trailing semicolon that is followed only by a comment', () => {
```
adamyeats, line 275: nit: the block-comment branch of the new regex is untested, so a `; /* note */` case and a `;\n-- a\n-- b` case would pin it. If the strip moves to the lexer, the same cases carry over.

adamyeats (changes requested):

PR #2180, opened 2026-09-22
fix(settings): apply connection pool settings from the data source config

### LINKED ISSUE #2179: Connection pool settings (maxOpenConns / maxIdleConns / connMaxLifetime) in jsonData are ignored
**What happened**:

We set `maxOpenConns`, `maxIdleConns` and `connMaxLifetime` in a datasource's `jsonData` and they had no effect. Connection counts on the ClickHouse side rose well past the configured `maxOpenConns`, and idle retention stayed at 2 regardless of what `maxIdleConns` was set to. Same result whether the values came from the datasource UI or from file provisioning.

Looking at the source, `LoadSettings` never reads those three keys out of `jsonData` — nothing parses them, so nothing reaches the `*sql.DB`. Every datasource therefore runs on `database/sql` defaults: `maxOpenConns` unlimited, `maxIdleConns` 2.

`[sql_datasources]` in `grafana.ini` is not a fallback. The SDK exposes those as `SQLConfig.DefaultMaxOpenConns` and friends, and this plugin does not read that either — so there is no path to a bounded pool from any layer.

The three pool keys to be honoured when present in `jsonData`, whether set through the datasource UI or through file provisioning.

The 4.8.0 changelog entry "Add support for connection pool settings" is what set the expectation — from the outside the feature reads as shipped.

1. Provision a ClickHouse datasource with pool settings in `jsonData`:

   ```yaml
   apiVersion: 1
   datasources:
     - name: clickhouse-pool-test
       jsonData:
         host: clickhouse
         port: 9000
         protocol: native
         maxOpenConns: 5
         maxIdleConns: 2
         connMaxLifetime: 60
   ```

2. Drive concurrent queries through that datasource — roughly 50 in flight is enough.
3. Count connections on the ClickHouse side:

   ```sql
   SELECT value FROM system.metrics WHERE metric = 'TCPConnection';
   ```

Observed: connection count rises well past `maxOpenConns`. Expected: it caps at 5.

The fix looks small and local. `LoadSettings` already handles exactly this shape for `dialTimeout` — the string-or-float64 branch at `pkg/plugin/settings.go:178-185` — so the three pool keys can follow the same pattern and then be applied to the `*sql.DB`.

Why it matters for us, in case it helps prioritise: we run one datasource per tenant, each with its own ClickHouse user, and a warm pool retains ~7 connections for five minutes after the last query. With unbounded `maxOpenConns` and no way to lower `maxIdleConns`, per-tenant retention is what sets our ceiling rather than actual concurrency. Being able to set `maxIdleConns: 2` would move the point at which we hit ClickHouse's `max_connections` by roughly 3.5x on our own numbers. We are shipping around it for now — this is a capacity constraint for us, not an outage.

Happy to open a PR if that would be welcome.

**Environment**:

- Grafana version: 13.0.2 (OSS)
- Plugin version: 4.20.0
- OS Grafana is installed on: Linux (official `grafana/grafana-oss:13.0.2` container)
- User OS & Browser: n/a — reproduced with provisioned datasources and server-side connection counts, no UI involved
- Others: ClickHouse 26.1.2.11, native protocol on port 9000

adamyeats:
## Type of Change

- [x] 🐛 Bug Fix

### What is the bug?

The data source config exposes Connection Max Lifetime, Max Idle Connections and Max Open Connections, and the docs list the same keys for provisioning. `LoadSettings` never read them from `jsonData`, so every data source ran with the 4.8.0 defaults (5 minutes, 25, 50) whatever the user set.

### How to reproduce

1. Provision a data source with `maxOpenConns: 5` in `jsonData`.
2. Run about 50 queries at the same time.
3. Count connections on the server: `SELECT value FROM system.metrics WHERE metric = 'TCPConnection'`. The count passes 5 and stops near 50.

### Related Issues

Closes #2179

## User-facing impact

- Pool values already stored in `jsonData` take effect after upgrade. Before this release the defaults (50 open, 25 idle, 5 minute lifetime) always applied, whatever was stored. Review any value you set before upgrading: a small `maxOpenConns` on a busy dashboard queues queries inside the pool until `queryTimeout` fires.
- A decimal or exponent typed into the config editor (for example `1.5` or `1e3`) is truncated to an integer, the same as an unquoted YAML number.
- A non-numeric value in provisioning (for example `maxOpenConns: five`) now fails the data source with `could not parse maxOpenConns value`. Before, it was ignored.

## Please check that:

- [x] Tests for this change have been added/updated. Unit tests cover number, string, empty and non-numeric values. An integration subtest connects with `maxOpenConns: 5` and asserts `db.Stats().MaxOpenConnections == 5` (fails on main with 50).
- [x] Documentation has been added/updated (where applicable). The docs already describe the keys and their defaults, so no change is needed.


pkg/plugin/settings.go
```diff
@@ -400,3 +410,37 @@ func loadHttpHeaders(jsonData map[string]interface{}, secureJsonData map[string]
...
 	return httpHeaders
 }
+
+// intSetting reads an integer jsonData key into the string form Settings
+// stores. Provisioned YAML gives a JSON number; the config editor gives a
+// numeric string, which its number input lets carry a fraction or an
+// exponent. Both paths truncate to an integer so the same value behaves the
+// same from either producer. An empty string means "not set", so the caller
+// applies its default. Anything else is a downstream error.
+func intSetting(jsonData map[string]interface{}, key string) (string, error) {
```
bossinc: Why dont we use `strconv.Atoi()` instead of truncating a float? I think it makes more sense to say its invalid input if we get a float when expecting and int. What do you think?

bossinc (changes requested):

pkg/plugin/settings.go
adamyeats: e069b9c4 parses strings with `strconv.ParseInt`, requires JSON numbers to be whole, and types the `Settings` fields as `int`. Truncation of `0.5` to `0` gave an unlimited pool, so an error is the safer result.

bossinc: sounds good! I see you made more changes. i'll play with it more in the morning

pkg/plugin/settings.go
bossinc: should we add intSetting to queryTimeout

bossinc (dismissed):
if its not worth doing this for dialtimeout and querytimeout then we are good to merge

pkg/plugin/settings.go
adamyeats: 4a1c85df applies `intSetting` to `dialTimeout`, `queryTimeout` and the legacy `timeout` key, so all five are `int` fields under the same whole-number rule. The `Atoi` re-parses in `driver.go` go with it.

